PR #35, opened 2015-07-29
move version to api package

ischneider:


ischneider:
@kapadia This prevents possible confusion with another top-level 'planet' module. Any issues?

kapadia:
I am curious why you suspect such a collision? If there happens to be another module called `planet`, then this is resolved by binding the `planet` module to another value using `import planet as blah`.

Most packages place versions at the top-level, making it a natural place for python users to check a package's version. Not sure why we would stray from that convention.

ischneider:
If the other package named 'planet' also has an `__init__` that declares a version, the first `planet/__init__` encountered (as determined by PYTHONPATH order) runs and the other doesn't. If the version is not 'really' used, this is not a big deal but if the code is doing any version sniffing, it could cause needless confusion.

See 'Version numbers for namespace packages' here for conventions: https://www.python.org/dev/peps/pep-0396/#examples

Another resolution would be to rename the top-level package, e.g. `planet_api`

kapadia:
Sorry, I let this one slip. I'm not concerned about colliding with another package. It's kind of cool that we snagged `planet` on PyPI, and don't see a compelling reason to worry about other packages using the same name. If there's concern, we can rename it `planet-api` or something similar. 

Either way, I do prefer to keep the version at the top level.

ischneider:
@kapadia We have at least one internal project using the planet namespace and my guess is, intentionally or not, there will be more. I have enough concern to have brought it up :smile:

Reminder, there are a few names rolling around here:

1. the name of the the pypi 'package' is `planet` (this won't change but does not need to reflect the internal packaging name)
1. a package providing a namespace, e.g. planet, numpy, etc., that may contain sub-packages or modules

Currently, the top-level namespace package only contains `__version__` so it's actually separated from the 'code' which all lives in the `api` namespace. In a contrived example of a 3rd party wanting to check the version (via code):

Makes more sense
```python
from planet import api
if api.__version__ == 'something':
  # api too old
# do stuff with api
api.do_stuff()
```

Makes less sense
```python
import planet
if planet.__version__ == 'something':
  # api too old
# do stuff with api
planet.api.do_stuff()
```

In addition to referencing the `__version__` field directly, a better approach is to get it from the distribution (pypi package):
```python
import pkg_resources
pkg_resources.get_distribution('planet').version
```

In fact, looking again, the recommendation is (4th bullet point in from https://www.python.org/dev/peps/pep-0396/#specification):

   For modules which live inside a namespace package, the module SHOULD include the `__version__` attribute. The namespace package itself SHOULD NOT include its own `__version__` attribute.

So we either rename the top-level namespace or move the version. I don't see any solid reasons to _not_ move the `__version__` declaration.

kapadia:
@ischneider If we have another project that collides with the package name `planet`, that is a compelling enough reason to change the name.

I'm still not convinced about burying the version. Very often I (and I'm assuming others) check versions at the top level. Examining large libraries in the Python ecosystem, you'll notice that, for instance, `numpy.linalg` does not have a `__version__` property. Rather it exists on the top level (`numpy.__version__`). Same is true for scipy and scikit-learn, and many others.

That said, I'll yield the decision to you.

jacobstr:
From the PEP:

Version numbers for namespace packages:

    >>> import flufl.i18n
    >>> import flufl.enum
    >>> import flufl.lock
    
    >>> print flufl.i18n.__version__
    1.0.4
    >>> print flufl.enum.__version__
    3.1
    >>> print flufl.lock.__version__
    2.1

ischneider:
Closing as this has been fixed in the V1 series. For reference, the version is available at `planet.api.__version__`

PR #36, opened 2015-07-31
initial workspace support + tests

ischneider:
~~not great for creating workspaces from scratch via command-line args yet~~


ischneider:
@kapadia I'd like to merge this ASAP as I have a 2nd PR that will depend on it. Any concerns?

FYI: note a good chunk of additions are tests + fixture and also some code duplication removed

kapadia:
Do it. One item - i'm not a fan of the `@-` convention, mostly because other geo tools (e.g. Rasterio, Fiona) do not use it. I'd like to maintain conventions with these packages, especially since the CLI is most inspired by them.

ischneider:
@kapadia As discussed, the `@` syntax remains as an mechanism for explicitly specifying the value for the option/argument should be read from the file (and it's an error if the file doesn't exist). Providing a value of `@-` or `-` still specifies stdin and by default, an attempt will be made to read the provided argument as if it was a file and if the file doesn't exist, the provided value is used. (I'm on :notebook: for noting this might cause confusion at some point)

kapadia:
Thanks for entertaining me @ischneider :) Just opened https://github.com/planetlabs/planet-client-python/issues/55, hoping we can find a solution to the ambiguous errors that `-` and `@-` open up.

Issue #60, opened 2015-08-10
add intersects support to mosaic queries once supported in API


ischneider:
currently not supported

ischneider:
fixed with @323f78d

PR #70, opened 2015-08-27
[NOT READY] area search for mosaic quads

kapadia:
@ischneider want to flag you on this. I've been in need of AOI search for mosaic quads. This should do it, however, there's an outstanding bug with the API that prevents it's usage.

### DIFF
--- a/planet/api/client.py
+++ b/planet/api/client.py
@@ -213,7 +213,7 @@ class Client(object):
         return self._get('workspaces/').get_body()
 
     def get_workspace(self, id):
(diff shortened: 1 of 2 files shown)


ischneider:
Right, see #60. Thanks @kapadia

kapadia:
Cool. Can't wait for that support.

ischneider:
For fun, I tried with just the geometry from a feature collection and this works.

kapadia:
Awesome.

Issue #53, opened 2015-08-07
signal handling broken on 2.7 with requests_futures


ischneider:
Ctrl-C has no effect and async tasks proceed. See: https://github.com/agronholm/pythonfutures/issues/25

Unless we can work-around with requests_futures or the futures backport is fixed , might need to use a multiprocessing approach.

kapadia:
@ischneider I just ran into a similar problem with a client i'm writing for the Sentinel archive. It uses `requests`, no futures, and SIGINT also doesn't work as expected.

Within `planet/scripts/__init__.py`, it might be worth appending the following to the top of the script.

```
import signal

def signal_handler(signal_code, frame):
    sys.exit(0)

signal.signal(signal.SIGINT, signal_handler)
```

PR #75, opened 2015-09-09
implement cancellation of async processes

ischneider:
see #53 and close when verified


ischneider:
While some of this complexity is related to #53 (and a work-around), mostly relates to the sync tool and would be more complex despite the bug.

PR #82, opened 2015-09-23
Fix problems when creating new workspaces from commandline

bosth:
Workspace now optional parameter to set-workspace command. This is useful when createing a new workspace with an AOI but no other filters.

It also suppresses a stacktrace when you don't provide any parameters to the `set-workspace` command.


ischneider:
outdated api functionality

PR #86, opened 2015-10-16
Updating Readme with new API commands

kmader:
Fixing the commands in the readme file so they all work


ischneider:
Thanks @kmader !

Issue #92, opened 2015-12-15
download and thumbnails cli commands do not handle missing API key well


ischneider:
Expected:
User sees terse error message without a stack trace

Actual:
User sees a big scary stack trace containing said message

PR #93, opened 2015-12-15
handle potential exceptions in download/thumbnails

ischneider:
some exceptions may be triggered synchronously in some
async cases (API key check, etc.) so handle these in the CLI
to avoid stack traces

most of this patch is adjustments to the cli test module to
allow using both a mock and normal Client

see #92


jacobstr:
:+1:

PR #97, opened 2016-04-20
intersection -> intersects

simgislab:
documentation fix


ischneider:
It's been a while, I know, and thank you for your enhancement, but the V1 API makes this obsolete.

PR #99, opened 2016-08-17
trying to make a quick rate limit handler

Chandler:


planet/api/client.py
```diff
@@ -19,7 +19,7 @@
 from .exceptions import (InvalidIdentity, APIException)
 from . import models
 from .utils import check_status
-
+from retrying import retry
```
ischneider: Where does this come from?

Chandler: ohh sorry, I meant to pull this against my own fork. I will open a proper PR here when I get this figured out.

PR #106, opened 2017-06-19
Check whether all images are either active or activating rather than …

gsganden:
…whether either all images are active or all images are activating.

I encountered an exception (`Exception: unexpected state [u'activating', u'active']`) when I tried to download both visual and analytic assets of a given item after previously downloading the visual assets. I believe that the exception arose because the visual asset was "active" while the analytic asset was still "activating." I think the fix is to have the _all_status function in `downloader.py` check whether all requested assets are either "active" or "activating" (allowing for a case in which some are active while others are activating), rather than checking whether they are either all "active" or all "activating."

### DIFF
--- a/planet/api/downloader.py
+++ b/planet/api/downloader.py
@@ -15,8 +15,12 @@ import logging
 import os
 import threading
 import time
+
+from typing import Iterable
+
 from .utils import write_to_file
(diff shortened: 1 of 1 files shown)


gsganden:
I'm getting other errors now, so I don't think I have solved the problem.

PR #107, opened 2017-06-21
Use filenames of the form <item_id>__<asset_type>.tif

gsganden:
The filenames that the CLI creates are not always sufficient for reconstructing item ids, which makes it difficult at best to retrieve metadata later. They also provide asset type information in a manner that is inconsistent at best. I propose adopting the convention that filenames should have the form {item_id}__{asset_type}, using a dunder between the item id and the asset type to make it easy to parse out the item id and asset type later. 

I have implemented this approach in this PR. The implementation is not ideal in that it modifies the write_to_file function rather than the get_filename function, but modifying get_filename would require substantially more work because get_filename is called by body objects that as far as I can tell do not have simple ways to access item id's.

### DIFF
--- a/planet/api/utils.py
+++ b/planet/api/utils.py
@@ -105,8 +105,8 @@ def write_to_file(
     is skipped because the destination exists, the callback will be invoked
     with ``callback(skip=body)``.
 
-    The name of the file written to will be determined from the Body.name
-    property.
+    The name of the file written to will be determined by the
+    get_custom_filename function.
 
     :param directory str: The optional directory to write to.
     :param callback func: An optional callback to receive notification of


ischneider:
Thanks @gsganden - I think the issue you note in the PR comment is valid but as is, this PR does a bit too much...

It seems like an option to support differing output filenames might suffice? Perhaps this includes template evaluation so if someone doesn't like your proposed format, they could provide their own?

I agree that having a standardized means to computing reflectance would be nice but would prefer not requiring rasterio or bs4 (AFAIK the builtin xml libs would suffice) *required* as part of the core.

In general, our goal is to keep the client rather light-weight and free of native dependencies.

gsganden:
That all makes sense. I might return to the filename issue at some point, but right now I'm not planning to work on it in the near future, so I am closing this PR.

PR #113, opened 2017-09-07
Get lists of item & asset types from the server

deleted:
Pursuant to [PE-15836](https://hello.planet.com/jira/browse/PE-15836).

Here's a naive approach, with minimal impact on existing option parsing code... Still doing client-side argument validation, but fetching valid type lists from the server.  It will fall back to the hard coded list if the server errors out or takes too long to respond.

It doesn't seem to add a huge amount of response time (percentagewise) to normal API requests, though each of these list endpoints, tested individually, seems to take about 200ms to respond.

Tests TK.

### DIFF
--- a/planet/scripts/item_asset_types.py
+++ b/planet/scripts/item_asset_types.py
@@ -1,0 +1,56 @@
+import requests
+
+ITEM_TYPE_URL = 'https://api.planet.com/data/v1/item-types/'
+ASSET_TYPE_URL = 'https://api.planet.com/data/v1/asset-types/'
+
+_item_types = None
+_asset_types = None
+
+# Default values here are used as a fallback
+# In case the API fails to respond or takes too long.
+_default_item_types = [
+    "PSScene4Band", "PSScene3Band", "REScene", "SkySatScene",
+    "REOrthoTile", "Sentinel2L1C", "PSOrthoTile", "Landsat8L1G"]
+
+_default_asset_types = [
+    "analytic", "analytic_b1", "analytic_b10", "analytic_b11", "analytic_b12",
+    "analytic_b2", "analytic_b3", "analytic_b4", "analytic_b5", "analytic_b6",
+    "analytic_b7", "analytic_b8", "analytic_b8a", "analytic_b9",
+    "analytic_bqa", "analytic_dn", "analytic_dn_xml", "analytic_ms",
+    "analytic_xml", "basic_analytic", "basic_analytic_b1",
+    "basic_analytic_b1_nitf", "basic_analytic_b2", "basic_analytic_b2_nitf",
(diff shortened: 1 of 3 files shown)


planet/scripts/item_asset_types.py
```diff
@@ -0,0 +1,56 @@
...
+    "basic_analytic_dn_rpc_nitf", "basic_analytic_dn_xml",
+    "basic_analytic_dn_xml_nitf", "basic_analytic_nitf", "basic_analytic_rpc",
+    "basic_analytic_rpc_nitf", "basic_analytic_sci", "basic_analytic_xml",
+    "basic_analytic_xml_nitf", "basic_panchromatic_dn",
+    "basic_panchromatic_dn_rpc", "basic_udm", "browse", "metadata_aux",
+    "metadata_txt", "udm", "visual", "visual_xml"
+]
+
+def _get_json_or_raise(url, timeout=0.7):
+    resp = requests.get(url, timeout)
```
rpaul80: Do we want to retry here ?

cp16net:
The approach looks good to me. Fix the import errors in the tests and i think its good.

planet/scripts/item_asset_types.py
deleted: That's a good question. It's a trade off between accuracy and sluggishness.

My thinking is that if a request fails the first time, it's pretty likely to fail subsequently, and since we have hard-coded lists to fail over to, we're better off giving up on retrieving the server-side list after one try. You could probably convince me otherwise, though.

deleted:
Tests are fixed.
I'd like to get sign-off from @ischneider before merging.

ischneider:
In some testing, I'm seeing that this is invoked way too often. Makes command completion miserable, too. Also takes time of `planet --help` from .34s to .98s.

It feels like we should never reach to the network until needed - I played with a 'lazy' `_LenientChoice` that resolves the list only on access but this also felt unnaturally laggy.

It seems like we should use the builtin list and only attempt the fetch if the choice is not found. `_LenientChoice` is already overriding `click.Choice` - instead of failing if `not matches`, this is likely the best place to attempt fetching from the API.

deleted:
@ischneider This last commit should address your concerns.
Gets `planet --help` back down under 250ms on my dev box.

harithadodda:
With new asset_types of Skysat, able to download using client succesfully. Tried with following.

planet data download --item-type SkySatScene --asset-type ortho_analytic_dn  --limit 1
planet data download --item-type SkySatScene --asset-type ortho_panchromatic_dn --limit 1
planet data download --item-type SkySatScene --asset-type ortho_pansharpened --limit 1
planet data download --item-type SkySatScene --asset-type ortho_visual --limit 1

ischneider:
Looks good to me - thanks for the improvement!

PR #119, opened 2017-09-13
move version and update user agent with client name and version

cp16net:


planet/api/dispatch.py
```diff
@@ -92,6 +93,10 @@ def rebuild_auth(self, prepared_request, response):
                     )
 
 
+def _get_user_agent():
+    return 'planet-python-client/%s' % __version__
```
ischneider: to reduce potential confusion, can we just use the project name?

ischneider:
Looks nice and thanks for the header debug addition. Minor comment re agent name.

planet/api/dispatch.py
cp16net: oh man i must have been dreaming and thought thats what the project name was...

cp16net: fixed

PR #125, opened 2017-10-03
Add new SR & skysat asset types to defaults

sarasafavi:
Just in case API takes too long, new asset types are now in our fallbacks


sarasafavi:
@danabauer :+1: docs & examples to come next!

PR #129, opened 2017-11-09
cli - implement filter combining

### LINKED ISSUE #128: cli download command improperly rejects --filter-json argument
To make the download experience nicer, the download command implicitly creates a permissions filter but does not combine it with the provided filter argument and exits with the message:

`Error: Specify filter options or provide using --filter-json, not both`

It seems like the proper behavior would be:

1. if the provided filter is an AND filter, add the permissions filter as a predicate
2. otherwise create a new top-level AND filter with the permissions and provided filter as predicates

ischneider:
when the filter-json flag is present, AND the filter
with whatever other filters are specified

also test addition + minor doc tweaks

this resolves #128 and adds some nice convenience

### DIFF
--- a/planet/scripts/util.py
+++ b/planet/scripts/util.py
@@ -79,20 +79,23 @@ def check_writable(dirpath):
 
 
 def filter_from_opts(**kw):
-    '''Build a AND filter from the provided filter_in OR kwargs defaulting to an
-    empty 'and' filter (@todo: API workaround).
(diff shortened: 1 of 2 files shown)


ischneider:
Once #130 is merged, the test should pass again.

ischneider:
Updated with the ability to extract the filter from a provided saved search in the JSON from the `--filter-json` flag. CI still expected to fail until #131 is merged

PR #130, opened 2017-11-09
add sentinel1 item-type and fix test breakage

ischneider:
An API addition broke a glob test in the client. This fixes the tests on master


ischneider:
Replacing this with #131 since it fixes CI

PR #137, opened 2017-12-14
Recommend '$ pip install --user' instead of just '$ pip install'

### LINKED ISSUE #134: difficulty installing on mac
Trying to follow the instructions at: https://www.planet.com/docs/api-quickstart-examples/cli/

I have python on my mac. Just installed pip 9.0.1.

When I type pip install planet, I get the following error:

image

Anything I am missing about how to successfully install this?

geowurster:
Kinda closes #134 

I have yet to encounter a good guide for introducing `pip` and communicating simple best practices but linking to something like that would be a big documentation win.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -11,9 +11,14 @@ Python client library and CLI for Planet's public API.
 Installation
 ------------
 
-Via pip::
+Via pip:
 
-    pip install planet
+.. code-block:: console
+
+    $ pip install planet
+
+The `--user <https://pip.pypa.io/en/stable/user_guide/#user-installs>`__
+flag is highly recommended for those new to `pip <https://pip.pypa.io>`__.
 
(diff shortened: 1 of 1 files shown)


geowurster:
@sarasafavi Good point.  My only concern is that this probably requires modifying `$PATH` to pick up the directory where executables are placed, we should maybe also provide a bit of documentation around that like https://github.com/planetlabs/planet-client-python/issues/134#issuecomment-349497952.  It looks like `pip 9.0` issues a warning when installing to a location that is not listed in `$PATH` (https://github.com/pypa/pip/pull/4553).

sarasafavi:
@geowurster fair argument; IMO we just need a solid pip+virtualenv doc to point people at. I'll merge this and we can adjust down the road as needed.

PR #138, opened 2018-01-18
(DRAFT) Added Quota Tool to Client

samapriya:
Created a new quota tool using experimental API endpoint
* Initializes client if not initialized before fetching quota details
For now prints:
* Allocation Name, Allocation active from, Quota Enabled, Total quota in SqKm, Total Quota used, Remaining quota in SqKm

* Changed version number to 1.1.1 assuming the naming convention

**In future api endpoint will change as experimental api url will change.**


planet/scripts/cli.py
```diff
@@ -103,3 +107,37 @@ def init(email, password):
     response = call_and_wrap(clientv1().login, email, password)
     write_planet_json({'key': response['api_key']})
     click.echo('initialized')
+
+@cli.command('quota')
+def quota():
+    '''Print allocation and remaining quota in Sqkm.'''
+    try:
+        with open (os.path.join(os.path.expanduser('~'), '.planet.json')) as f:
+            data=json.load(f)
```
sarasafavi: `f` here is the same as `planet.api.utils._planet_json_file()` -- however you can simplify this file read entirely, and avoid code duplication, by calling `planet.api.utils.read_planet_json()` (see: https://github.com/samapriya/planet-client-python/blob/master/planet/api/utils.py#L31)

planet/scripts/cli.py
```diff
@@ -103,3 +107,37 @@ def init(email, password):
...
+        main=requests.get('https://api.planet.com/auth/v1/experimental/public/my/subscriptions', auth=HTTPBasicAuth(data["key"], ''))
+        if main.status_code==200:
+            content=main.json()
+            n=(str(main.json()).count("quota_sqkm"))
+            i=0
+            while i<n:
```
sarasafavi: No need to set loop counters like this in Python; you can delete L120-L121 & L135; then just change L122 to `for item_id in content`

planet/scripts/cli.py
```diff
@@ -103,3 +107,37 @@ def init(email, password):
...
+                item_id=(content[i])
+                print(" ")
```
sarasafavi: use `click.echo` instead of `print()`, for consistency with the rest of this file (see also: http://click.pocoo.org/5/utils/#printing-to-stdout)

planet/scripts/cli.py
```diff
@@ -103,3 +107,37 @@ def init(email, password):
...
+                print("Total Quota used: "+str(item_id['quota_used']))
+                if (item_id['quota_sqkm'])is not None:
+                    print("Remaining Quota in SqKm: "+str(float(item_id['quota_sqkm'])-float(item_id['quota_used'])))
+                else:
+                    print("No Quota Allocated")
+                print("")
+                i=i+1
+        else:
+            print("Failed with exception code: "+str(main.status_code))      
+    except Exception as e:
```
sarasafavi: Using `except Exception` is a bad idea! What happens if anything in the block above throws an Exception that's _not_ related to an un-initialized client (e.g., what if the response JSON is missing an expected field, and one of the `print` lines throws a `KeyError`?)? We'll never know about it, and the CLI will just try to ask the user to initialize regardless.

Instead, do:
`except IOError:  # note that you don't need "as e"`

planet/scripts/cli.py
```diff
@@ -103,3 +107,37 @@ def init(email, password):
...
+        #print(e)
+        print("Initialize Planet Client First and try again")
+        subprocess.call(["planet", "init"],shell=False)
```
sarasafavi: So at first glance, this is a nice bit of user-friendly assistance. However, what about users that use the documented option of `-k` to pass in their API key? In that case, they may legitimately not have a `.planet.json` file, but since we only need their API key to make the quota call that shouldn't matter.

What we should do instead:

1) Allow the user to pass in their API key for the `quota` command, just as they can do with e.g., `planet -k <some key> data`. This means adding a `@click.option()` decorator to the `quota()` function for `-k` option.

2) If the check for `planet.json` throws an `IOError`, before forcing the user to initialize the client we should first check for a passed-in API key. IFF that is not provided, then you could continue with the forced initialization. Consider however, that it may be more user-friendly to simply fail with a message asking them to _either_ provide an API key _or_ initialize the client (since they may prefer one over the other).

sarasafavi:
@samapriya see the review comments above, also:

Travis CI is failing due to your changes not meeting PEP8 standards (i.e., the `flake8` errors seen here: https://travis-ci.org/planetlabs/planet-client-python/jobs/330317917#L605). If you're not familiar with PEP8 styles or the flake8 tool, have a glance at https://www.python.org/dev/peps/pep-0008/, then check out http://flake8.pycqa.org/en/latest/. I highly recommend installing flake8 in your local dev env and running it against your Python code!

planet/scripts/cli.py
samapriya: Resolved

planet/scripts/cli.py
samapriya: Resolved

planet/scripts/cli.py
samapriya: Resolved

planet/scripts/cli.py
samapriya: Resolved

samapriya:
Flake8 corrections and reviews have been taken care of the tool now functions as 

* planet quota (if you have used email to initialize planet client)
* planet quota --key "a12kh3kh21kh1" (to initialize simply with key)

Look forward to any other corrections

planet/scripts/cli.py
```diff
@@ -103,3 +107,85 @@ def init(email, password):
     response = call_and_wrap(clientv1().login, email, password)
     write_planet_json({'key': response['api_key']})
     click.echo('initialized')
+
+
+@cli.command('quota')
+@click.option('-k', default=None, prompt=False, hide_input=True, help=(
+    'Enter your API Key'))
```
sarasafavi: So this one is my fault: I did suggest you add `-k` as an option to this new command "quota". However, since we are already capturing an API key (if the user passes one in) earlier in cli.py, there's no need for us to reinvent the wheel here.

So instead, let's do:
- remove L113-114
- add `k = client_params.get('api_key')` inside the function.

If you're not sure what that does, or why, let me know!

Also, a bit of trivial nagging: this isn't strictly _wrong_, but it's usually nicer (and more "Pythonic") to be explicit with variable names. No reason to use `k` when we mean `key` or `apikey` here.

planet/scripts/cli.py
```diff
@@ -103,3 +107,85 @@ def init(email, password):
...
+                        "Total Quota used: " + str(
+                            item_id['quota_used']))
+                    if (item_id['quota_sqkm'])is not None:
+                        click.echo(
+                            "Remaining Quota in SqKm: " + str(
+                                float(item_id['quota_sqkm']) - float(
+                                    item_id['quota_used'])))
+                    else:
+                        click.echo("No Quota Allocated")
+                    print("")
```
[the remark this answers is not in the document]
samapriya: Fixed

ischneider:
One thing I'd like to see is a separation of UI + transport - note that you had to add requests to package because all other operations are done elsewhere. Specifically, this means:

* move the HTTP logic into the client: https://github.com/planetlabs/planet-client-python/blob/master/planet/api/client.py

Additionally, note that no other major functionality is in `cli.py` - it's brought in by the `v1.py` module. I'd like to see something similar here.

planet/scripts/cli.py
```diff
@@ -103,3 +107,85 @@ def init(email, password):
...
+                'https://api.planet.com/auth/v1/' +
+                'experimental/public/my/subscriptions',
+                auth=HTTPBasicAuth(
+                    k, ''))
+            if main.status_code == 200:
+                content = main.json()
+                for item_id in content:
+                    click.echo(" ")
+                    click.echo(
+                        "Allocation Name: " + str(
```
ischneider: The more idiomatic way is using format strings. I personally prefer the `%s` variety but would reluctantly accept `{}`.

For example, `click.echo('Allocation Name: %s' % item_id['organization']['name'])`

Also: strings should be single quoted (as in example above)

planet/scripts/cli.py
```diff
@@ -103,3 +107,85 @@ def init(email, password):
...
+                        click.echo(
+                            "Remaining Quota in SqKm: " + str(
+                                float(item_id['quota_sqkm']) - float(
+                                    item_id['quota_used'])))
+                    else:
+                        click.echo("No Quota Allocated")
+                    print("")
+        else:
+            dkey = open(os.path.join(os.path.expanduser('~'), '.planet.json'))
+            data = json.load(dkey)
```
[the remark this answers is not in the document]
samapriya: fixed

planet/scripts/cli.py
samapriya: fixed

planet/scripts/cli.py
```diff
@@ -103,3 +107,85 @@ def init(email, password):
...
+                            "Remaining Quota in SqKm: " + str(
+                                float(item_id['quota_sqkm']) - float(
+                                    item_id['quota_used'])))
+                    else:
+                        click.echo("No Quota Allocated")
+                    print("")
+            else:
+                click.echo("Failed with exception code: " + str(
+                    main.status_code))
+    except IOError:
```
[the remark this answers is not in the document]
samapriya: fixed

planet/scripts/cli.py
samapriya: Made the changes Ian, thank you for the suggestions

planet/scripts/cli.py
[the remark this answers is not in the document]
samapriya: I think I finally got the internal logic correct, going to upload the latest version to run Travis-CI tests

sarasafavi:
@samapriya changes are looking good. Please see @ischneider's message above re: migration of the request logic into `planet.api.client`, and migration of the new `quota` command itself into `planet.scripts.v1`, for next steps.

sarasafavi:
closing; outdated/obsolete

PR #140, opened 2018-03-22
Add mosaic support

joferkington:
Currently, the Python client does not support the v1 mosaics api.  As a result, there's no straightforward way for customers to download quads in bluk, as the Python client provides the main CLI tool for downloading. 

This still needs tests added, but I wanted to at least start discussion and see if this is the right place/way to include mosaic support in the python client.

### DIFF
--- a/planet/api/client.py
+++ b/planet/api/client.py
@@ -270,3 +270,76 @@ class ClientV1(_Base):
         '''
         url = 'data/v1/item-types/%s/items/%s/assets' % (item_type, id)
         return self._get(url).get_body()
+
+    def get_mosaics(self):
+        '''Get information for all mosaics accessible by the current user.
+
+        :returns: :py:Class:`planet.api.models.Mosaics`
+        '''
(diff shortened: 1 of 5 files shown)


planet/api/client.py
```diff
@@ -270,3 +270,77 @@ def get_assets_by_id(self, item_type, id):
...
+        return self._get(url, models.Mosaics).get_body()
+
+    def get_mosaic_by_name(self, name):
+        '''Get the API representation of a mosaic by name.
+
+        :param name str: The name of the mosaic
+        :returns: :py:Class:`planet.api.models.Mosaics`
+        :raises planet.api.exceptions.APIException: On API error.
+        '''
+        params = {'name__is': name}
```
joferkington: This api parameter is still undocumented, but I'm not sure there's another way to do this (other than iterating through every item returned and doing client-side filtering).

joferkington:
@tjmac - The issue was a bug in the api service introduced a few months back.  It's recently been fixed and things should work correctly.

PR #149, opened 2018-10-03
Fix mosaic quad info/contribution

joferkington:
Mosaic quad information retrieval (i.e. `mosaics quad-info` and `mosaics contribution`) is currently broken.  For example:

```
$ planet mosaics quad-info basemap_medres_weekly_permian_20180917_20180923 446-1201
Error: InvalidAPIKey: {"message":"Insufficient Privileges"}
```

`ClientV1.get_quad_by_id` currently constructs the url it requests by naively appending paths to the mosaic's `_links.self`.  However, mosaic url returned by the api has the user's api key included as a parameter.  This results in a url with the quad's path as part of the `api_key` parameter, resulting in the 403 response.

E.g. `planet mosaics quad-info basemap_medres_weekly_permian_20180917_20180923 446-1201`
actually requests `https://api.planet.com/basemaps/v1/mosaics/1e9ae02c-64b7-428d-bbcd-7df3f9a94f14?api_key=FOOBAR/quads/446-1201`

This PR fixes the issue by constructing the quad url based on the mosaic's id and the quad endpoint.

### DIFF
--- a/planet/api/client.py
+++ b/planet/api/client.py
@@ -316,8 +316,8 @@ class ClientV1(_Base):
         :returns: :py:Class:`planet.api.models.JSON`
         :raises planet.api.exceptions.APIException: On API error.
         '''
-        url = '{}/quads/{}'.format(mosaic['_links']['_self'], quad_id)
-        return self._get(url).get_body()
+        path = 'basemaps/v1/mosaics/{}/quads/{}'.format(mosaic['id'], quad_id)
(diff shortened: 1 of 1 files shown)


joferkington:
Note that test failures are unrelated to this PR. They're because additional item types have been added to the api, and the test checks against a fixed list. (Specifically, `SkySatCollect` and the various MODIS item types are not in the fixed list.)

PR #150, opened 2018-10-11
Add LICENSE explicitly to setup.py so that pex can find it

ericrdunham:
With more recent versions of `setuptools` (or `distutils`), `pex` will explode on trying to include `LICENSE` unless it is explicitly declared.


ericrdunham:
@ischneider I don't know who else to ping here. :)

sarasafavi:
:+1: test failures unrelated

Issue #148, opened 2018-09-20
planet init fails


strixcuriosus:
`planet init` seems to be broken:

```
planet init
Email: asdf@example.com
Password:
Error: Unexpected response: invalid parameters, login process has changed
```

planet 1.1.0
Python 2.7.15

sarasafavi:
Thanks for the report @strixcuriosus ! Fix is in the works.

craigdsouza:
Hello.. has there been any resolution here? I'm trying to run pclip commands and I'm getting the same error.. thanks..
Python version is 2.7.15, Windows 10, pclip 0.2.7

sarasafavi:
@craigdsouza @strixcuriosus this issue is now fixed in the latest release, 1.2.1: https://pypi.org/project/planet/1.2.1/

PR #152, opened 2018-11-08
set login headers as content-type:application/json during planet init

sarasafavi:
smallest patch ever, but now `planet init` works again


sarasafavi:
addresses #148

sarasafavi:
test failures unrelated (see #149 for same failures; we need to update that fixed list of item types)

Issue #98, opened 2016-04-25
better Windows experience


ischneider:
- check if recent pex versions improve compatibility (errors running w/ current pex 'binary')
- pywin32 dependency

sarasafavi:
ref #162

PR #162, opened 2019-03-06
add pywin32 dep for windows platforms

### LINKED ISSUE #145: win32api module issue on windows machine
I had this crop up on someone's windows computer when they installed the planet client
ImportError: **No module named 'win32api'**

Recommended fix is **pip install pywin32**

Would it great to check os type
something like
```python 
import os
import subprocess
if os.name=="nt":
    subprocess.call('pip install pywin32',shell=True)
```

Either in setup or during first run

sarasafavi:
closes #145 and #98

see https://www.python.org/dev/peps/pep-0508/#environment-markers for formatting reference

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -77,6 +77,7 @@ setup(name='planet',
           'click',
           'requests',
           'requests_futures',
+          'pywin32 >= 1.0;platform_system=="Windows"'
       ],
       extras_require={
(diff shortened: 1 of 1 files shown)


sarasafavi:
@samroy-pl could you sanity check this on a windows machine?

PR #179, opened 2019-07-10
Analytics release 0.1

quaterneon:


CHANGES.txt
```diff
@@ -1,4 +1,8 @@
-1.3.4 (2019-06-27)
+1.3.0 (2019-07-10)
+-------------------------------------
+- Added support for Analytic Feeds
+
+1.2.4 (2019-06-27)
```
sarasafavi: thanks for the typo fix :)

planet/scripts/cli.py
```diff
@@ -62,8 +75,12 @@ def configure_logging(verbosity):
 @click.option('-u', '--base-url', envvar='PL_API_BASE_URL',
               help='Change the base Planet API URL or ENV PL_API_BASE_URL'
                    ' - Default https://api.planet.com/')
+@click.option('-u', '--analytics-base-url', envvar='PL_ANALYTICS_API_BASE_URL',
```
sarasafavi: this is re-using the same short option name (`-u`) as `base-url` above, is `analytics-base-url` supposed to function as an alias?

quaterneon: Oh, good catch, it shouldn't be an alias as the two may need to be controlled separately.

quaterneon: Alright, I just made it `-au` so that it won't overlap.

PR #194, opened 2019-10-21
Simplify analytics configuration

quaterneon:


planet/scripts/cli.py
```diff
@@ -75,13 +62,8 @@ def configure_logging(verbosity):
...
               help='Change the base Planet API URL or ENV PL_API_BASE_URL'
                    ' - Default https://api.planet.com/')
-@click.option('-au', '--analytics-base-url',
-              envvar='PL_ANALYTICS_API_BASE_URL',
-              help=('Change the base Planet API URL or ENV '
-                    'PL_ANALYTICS_API_BASE_URL'
-                    ' - Default https://api.planet.com/analytics'))
 @click.version_option(version=__version__, message='%(version)s')
-def cli(context, verbose, api_key, base_url, analytics_base_url, workers):
+def cli(context, verbose, api_key, base_url, workers):
```
strixcuriosus: I notice that many of the derived urls assume that the base_url has a trailing slash.
e.g. `client.base_url + 'analytics/feeds'`

This probably only matters to Planeteers for now, but if you feel so inclined you could handle urls like `https://my-other-planet-api.example.com` with grace or update the help string for the `--base-url` flag.

strixcuriosus:
LGTM, thanks for simplifying the url configs!

planet/scripts/cli.py
quaterneon: I added a case to the client init to make sure that the trailing slash exists.

PR #196, opened 2019-10-22
WIP: Get data as pandas dataframes

quaterneon:


samapriya:
**Fiona, geopandas, shapely and gdal installations** will mostly fail in windows. If added to requirements and the planet tool itself will not install. **Windows** does not play well with these libraries and they all have their own set of installation instructions for windows.  :)

PR #200, opened 2019-11-07
removing monkey patch

carl-adams-planet:
I don't believe this monkey patch works with Python 3.7.  I'm also not so sure that suppressing warnings about bad SSL practices was a good idea in the first place. 

On python 3.7, if you set the env variable DISABLE_STRICT_SSL to true to deactivate SSL cert checks in dispatch.py, you will encounter this error: 

> [  . . . Stack trace trimmed for brevity . . .  ]
>   File "/Library/Frameworks/Python.framework/Versions/3.7/lib/python3.7/warnings.py", line 110, in _showwarnmsg
>     msg.file, msg.line)
> TypeError: hack() takes 4 positional arguments but 6 were given

This appears to be an issue with how the monkey patch interacts with the Python 3.7 warnings.py. 

(I also am not sure that the existing DISABLE_STRICT_SSL feature in dispatch.py works in all cases - I've seen it not cover everything the QGIS plugin does. But that's a question for another merge request.)

### DIFF
--- a/planet/scripts/util.py
+++ b/planet/scripts/util.py
@@ -1,4 +1,4 @@
-# Copyright 2017 Planet Labs, Inc.
+# Copyright 2017-2019 Planet Labs, Inc.
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
--- a/planet/scripts/util.py
+++ b/planet/scripts/util.py
@@ -23,11 +23,9 @@ import tempfile
 import textwrap
 import threading
 import time
-import warnings
 
 import click
 from click import termui
-from requests.packages.urllib3 import exceptions as urllib3exc
 
 from planet import api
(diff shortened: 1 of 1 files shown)


sarasafavi:
that monkeypatch predates my involvement on the project and I agree that suppressing that warning doesn't seem like a good idea -- lgtm :+1:

carl-adams-planet:
As a related matter, I wonder if we should either remove the existing DISABLE_STRICT_SSL behavior, or promote it to a first class feature.  Hooks like that to tweak a library's behavior are always a mixed bag.  On the one hand, the utility in developing/debugging is obvious. On the other, when the library is instantiated in a larger application that may include overlapping functionality from a multitude of libraries, the inconsistent behavior and configuration is maddening for the end-user.  The shouldn't have to debug the code to figure out how to configure it.

PR #206, opened 2020-04-24
create orders additions ids_from_search, --clip

jar-o:


jar-o:
hrm... `make check html-docs` runs fine locally

docs/source/cli/examples.rst
```diff
@@ -227,19 +227,61 @@ The minimal command to create a simple order looks something like::
 If no toolchain or delivery details are specified, a basic order with download
 delivery will be placed for the requested bundle including the item id(s) specified.
 
-Additionally, optional toolchain & delivery details can be provided on the
-command line, e.g.:::
+In the place of `--id`, you can insert a Data search string. This will populate
+the list of IDs from the search. For example::
+
+    planet orders create --name "my order" \
+      --ids_from_search "--item-type PSScene3Band --date acquired gt 2017-02-14 --date acquired lt 2017-03-14" \
```
strixcuriosus, line 234: It seems likely that someone creating an order for a date range might also want to apply a geometry filter. An example of `--ids_from_search` with both spatial (geometry filter) and temporal (date rnage) filtering would be helpful!

planet/scripts/v1.py
```diff
@@ -711,8 +715,14 @@ def cancel_order(order_id, pretty):
...
 @click.option('--name', required=True)
-@click.option('--id', required=True,
-              help='One or more comma-separated item IDs')
+@click.option('--id', help='One or more comma-separated item IDs',
+                cls=RequiredUnless, this_opt_exists='ids_from_search')
+# Note: This is passed as a string, because --item-type is a required field for
+# both 'data search' and 'orders create'.
+@click.option('--ids_from_search',
+              help='Embedded data search')
+@click.option('--clip',
```
strixcuriosus, line 724: it would be nice to be able to pass a geojson file path here.

planet/scripts/v1.py
```diff
@@ -735,6 +745,18 @@ def cancel_order(order_id, pretty):
...
+    if ids_from_search is not None:
+        data_search = kwargs.get('ids_from_search').split()
+        if len(data_search) > 0:
+            runner = CliRunner()
+            resp = runner.invoke(quick_search, data_search).output
+            try:
+                id_list=ids_from_search_response(resp)
+            except ValueError:
+                raise click.ClickException('ids_from_search, {}'.format(resp))
+            kwargs['id'] = id_list
```
strixcuriosus, line 758: It might be worth putting a note in this client about how `clip` and `id` interact on the backend.
If ids are provided for items that do not intersect the clip AOI, the [tools reference](https://developers.planet.com/docs/orders/tools-reference/#clip) indicates that those items will be discarded. If the `ids_from_search` don't intersect the `clip` you could create an order that would have no results. Does the backend allow such an order to be created? If not, perhaps adding some spatial validation in this client would be worthwhile for now.

planet/scripts/v1.py
```diff
@@ -711,8 +715,14 @@ def cancel_order(order_id, pretty):
...
-@click.option('--id', required=True,
-              help='One or more comma-separated item IDs')
+@click.option('--id', help='One or more comma-separated item IDs',
+                cls=RequiredUnless, this_opt_exists='ids_from_search')
+# Note: This is passed as a string, because --item-type is a required field for
+# both 'data search' and 'orders create'.
+@click.option('--ids_from_search',
+              help='Embedded data search')
+@click.option('--clip',
+              help='Provide a GeoJSON AOI for clipping')
```
strixcuriosus, line 725: Clarify what you mean by AOI here. Is this a geojson Feature Collection, Feature, or Geometry? (Should we support all of those possibilities?)

strixcuriosus: Since this client is already `click` based, [cligj](https://github.com/mapbox/cligj) could be helpful for parsing geojson args.

planet/scripts/v1.py
jar-o: so there is currently a `--tools` option that requires a path to a file. the clip operation can be specified there already. the motivation for a string-based param is specifically to eliminate the need   to create extraneous files to run a command.

planet/scripts/v1.py
jar-o: it's geometry only, which is what orders supports https://developers.planet.com/docs/orders/tools-reference/#clip ... I will update the `help` message to be more clear.

also the geometry will be parsed/validated when submitted to the backend. this happens almost immediately, so I don't see too much value in also parsing it here in the client.

docs/source/cli/examples.rst
jar-o: makes sense, I'll update

planet/scripts/v1.py
jar-o: yeah this is a good callout and kind of one of those UX things that have bounced around orders for some time. the outcome can be 1) all items intersect clip and are delivered, 2) some intersect and only those are delivered, and 3) none intersect and order has no results.

I'm kind of inclined to just update docs here. I sort of feel like the `planet` tool, being a thin front-end wrapper, should always defer to whatever backend for behaviors, or we have to make sure behaviors in the client are always updated to stay in sync with the backend. thoughts?

planet/scripts/v1.py
jar-o: ... thinking about it more, I could see using the `curl` inspired `@file/path` string notation here to allow both, if you feel strongly it's beneficial

jar-o: oh I see, this is already the case for `--geom` under `data search`. I will add to be consistent

```
  GEOM

  Specify a geometry in GeoJSON format either as an inline value, stdin, or
  a file. '@-' specifies stdin and '@filename' specifies reading from a file
  named 'filename'.
```

thanks for bring this up 👍

docs/source/cli/examples.rst
jar-o: added spatial + temporal filter example to PR, and a couple new ways to reference geo things to my vocabulary :)

planet/scripts/v1.py
jar-o: can `--clip @file` now, also updated docs

planet/scripts/v1.py
jar-o: clarified docs to `Geometry`

planet/scripts/v1.py
jar-o: I have updated the docs to clarify the behavior here. marking as resolved (feel free to unresolve, if you think more discussion is needed).

PR #207, opened 2020-05-08
WIP: Experimental basemaps client

strixcuriosus:
While the existing cli tooling in this package is quite useful for searching, ordering, and downloading from the command line, I have found the `ClientV1` class to be cumbersome and nonintuitive to use directly within python scripts. I almost always write my own client or use `requests` directly when working with our public apis.

The goal of this pull request is to demonstrate a lightweight importable client for one specific service (basemaps).

TODO:
- [ ] additional tests
- [ ] docstrings
- [ ] usage examples

### DIFF
--- a/planet/api/client.py
+++ b/planet/api/client.py
@@ -20,12 +20,13 @@ from .exceptions import (InvalidIdentity, APIException)
 from . import models
 from . import filters
 
+BASE_URL = 'https://api.planet.com/'
+
 
 class _Base(object):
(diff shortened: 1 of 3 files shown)


sgillies:
@strixcuriosus we're winding version 1.x of this package down and so I'm going to close this PR. I recommend getting involved in v2 development. We're very interested in making the next version more Pythonic and ergonomic.

PR #211, opened 2020-05-12
Fix make html-docs

strixcuriosus:
This fixes the `make check html-docs` part of the build, which is now failing on master:
https://travis-ci.org/github/planetlabs/planet-client-python/builds/686344081

**Changes:**
- fix sphinx parameter to get rid of a warning
- update the checked in generated reference.rst file
- fix a flake8 lint error

**Demo**:
- `make check`
- `make html-docs`

**More Details**
Sphinx warning:
```
/path/to/planet-client-python/planet/api/client.py:docstring of planet.api.ClientV1.quick_search:4: WARNING: Inline strong start-string without end-string.
```

Repro steps for the sphinx warning fix (try on master and on this branch):
- `pip install -e ".[dev]"` in a new virtualenv for this repo
- `make html-docs`

### DIFF
--- a/docs/source/cli/reference.rst
+++ b/docs/source/cli/reference.rst
@@ -10,30 +10,40 @@ Option Types Formatting
 -----------------------
 
 
-.. _cli-metavar-FIELD-COMP-VALUE:
+.. _cli-metavar-ASSET-TYPE:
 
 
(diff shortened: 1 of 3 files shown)


quaterneon:
LGTM!!

PR #227, opened 2020-09-15
fixed orders model object with next key, Fix #226

### LINKED ISSUE #226: The Orders object is not using the correct NEXT_KEY parameter name.
When I tried to use the `items_iter()` or `iter()` methods over the `get_orders()` client method, they were only displaying the first page. It is because the response is using the "next" key instead of "_next".

https://github.com/planetlabs/planet-client-python/blob/883ff64437dd0edc07f82a101048ada7abc60440/planet/api/models.py#L323

proposed fix:

```python
class Orders(Paged):
    ITEM_KEY = 'orders'
    NEXT_KEY = 'next'
```

ingdanielguerrero:
fixed orders model object with next key, Fix #226

### DIFF
--- a/planet/api/models.py
+++ b/planet/api/models.py
@@ -322,6 +322,7 @@ class WFS3Features(AnalyticsPaged):
 
 class Orders(Paged):
     ITEM_KEY = 'orders'
+    NEXT_KEY = 'next'
 
 
 class Order(JSON):


sarasafavi:
thank you @ingdanielguerrero !

PR #229, opened 2020-10-07
WIP: Orders cli file download 222

### LINKED ISSUE #222: orders CLI sometimes doesn't download all files
Sometimes downloading an order via the CLI does not download all files.

Example JSON:
```
{
    "name": "20200505_193841_ssc4d1_0018",
    "products": [
        {
          "item_ids": [
            "20200505_193841_ssc4d1_0018"
          ],
          "item_type": "SkySatScene",
          "product_bundle": "analytic"
        }
      ],
    "state": "success",
    "tools": [
        {
          "reproject": {
            "kernel": "cubic",
            "projection": "EPSG:4326"
          }
        }
      ]
}
```

Expected files:
```
'20200505_193841_ssc4d1_0018_analytic_reproject.tif',
'20200505_193841_ssc4d1_0018_analytic_udm_reproject.tif',
'20200505_193841_ssc4d1_0018_metadata.json'
```

jreiberkyle:
Find and fix the bug that is causing the CLI to intermittently miss downloading some files in an order.

Closes #222


jreiberkyle:
So far the issue can be reproduced by running. the integration test `integration_tests/test_download.py`.

The unit test that attempted to reproduce this bug (`tests/test_v1_cli.py`) succeeds without issue.

The issue is likely somewhere in the threading. There are two places where threading is used. 1) in handle_interrupt, which is cli specific and 2) within download stages.

Now the goal is to add some testing around these two functionalities to see if we can find the bug.

jreiberkyle:
This issue is made less acute in #232 and will be completely non-existant in V2.

PR #232, opened 2020-10-23
force zip of entire order via cli and create test of download

### LINKED ISSUE #231: Zip all CLI-based Orders to avoid incomplete downloads
see also: https://github.com/planetlabs/planet-client-python/issues/222

a non-deterministic bug occasionally results in incomplete downloads when Orders placed via CLI are not zipped: we should update the interface to force zipping for all orders in order to avoid this.

jreiberkyle:
changes cli to force zip of entire order. Also adds a script to test download. Sometimes the test finds the run fails to download both files (need to check if its always the zip or sometimes the manifest). The frequency of failure is about 1/8 whereas without zip the frequency of failure was 3/5 - so, improvement!

Closes #231

### DIFF
--- a/planet/scripts/util.py
+++ b/planet/scripts/util.py
@@ -98,7 +98,6 @@ def create_order_request(**kwargs):
     bundle = kwargs.get('bundle')[0]
     ids = kwargs.get('id').split(',')
     email = kwargs.get('email')
-    archive = kwargs.get('zip')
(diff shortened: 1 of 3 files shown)


jreiberkyle:
While the failure always includes the zip not being downloaded, sometimes (1 out of 3 failures) the manifest is not downloaded either.

sarasafavi (approved):

PR #234, opened 2020-10-29
create orders api client

### LINKED ISSUE #217: add support for orders via the api
Add support for orders via the client api.

I have been trying to reproduce what the cli does for orders download ([download_order](https://github.com/planetlabs/planet-client-python/blob/883ff64437dd0edc07f82a101048ada7abc60440/planet/scripts/v1.py#L774), but so far unsuccessful.

Try 1: no polling

immediately following submitting an order:

```python
from planet.api import downloader
dl = downloader.create(client, order=True)
items = client.get_individual_order(order_id).items_iter(limit=None)
dl.download(items, [], 'data')
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-83-d86cc6078d77> in <module>
      1 dl = downloader.create(client, order=True)
----> 2 items = client.get_individual_order(order_id).items_iter(limit=None)
      3 dl.download(items, [], 'data')

/opt/conda/lib/python3.7/site-packages/planet/api/models.py in items_iter(self, limit)
    348         '''
    349 
--> 350         locations = iter(self.get_locations())
    351 
    352         # if limit is not None:

/opt/conda/lib/python3.7/site-packages/planet/api/models.py in get_locations(self)
    337     def get_locations(self):
    338         results = self.get_results()
--> 339         locations = [r[self.LOCATION_KEY] for r in results]
    340         return locations
    341 

TypeError: 'NoneType' object is not iterable
```

Try 2: with outside polling

after submitting an order:
```python
def poll_for_success(order_id, client, num_loops=50):
    count = 0
    while(count < num_loops):
        count += 1
        order_info = client.get_individual_order(order_id).get()
        state = order_info['state']
        print(state)
        success_states = ['success', 'partial']
        if state == 'failed':
            raise Exception(response)
        elif state in success_states:
            break
        
        time.sleep(10)
        
poll_for_success(order_id, client)
```

now that the order is successful:
```
from planet.api import downloader
dl = downloader.create(client, order=True)
items = client.get_individual_order(order_id).items_iter(limit=None)
dl.download(items, [], 'data')
```
this actually works (though the stats messaging is inaccurate) but only the first time we try to download that `order_id`. If we try to download it to another location, even using another downloader instance, it says it is already complete and doesn't download the items.

jreiberkyle:
Closes #217

Create orders api client for v2


jreiberkyle:
@sarasafavi I would love some feedback on the interplay between OrdersClient and Orders in list_orders(). They are tightly interwoven and I have a feeling there may be a cleaner way to break up paging. Let me know if you spot anything!

jreiberkyle:
We decided to add concurrency at this point. This means only supporting python 3.7+ and then using asyncio, which joins the standard library in 3.7. This also means moving from requests for http communication to another library. Two major contenders are [httpx](https://github.com/encode/httpx) and [aiohttp](https://docs.aiohttp.org/en/latest/index.html). The first major steps will be 1. determining the http communication library and 2. determining concurrency ux design.

jreiberkyle:
@sarasafavi what would you think about merging this PR and having a new ticket/PR related to adding concurrency?

jreiberkyle:
According to @sarasafavi , this looks good to merge and #238 will address concurrency.

PR #239, opened 2021-01-14
Removed plural in bundle

cholmes:
I just had my hope raised that I could put in two bundles in one order because the 'help' said `bundle(s)`. It then failed, since required is false. If the order system can actually take two bundles that'd be awesome, but in the meantime this just corrects the language.


jreiberkyle:
Thanks for catching this @cholmes. Also, if you have a use case for ordering multiple bundles, we can look into supporting that in v2 (it's going to be **nice**). Feel free to make a ticket.

PR #240, opened 2021-02-05
Move v2 orders client over to async

### LINKED ISSUE #238: add concurrency to the core of the client
Many Planet API calls will benefit from concurrency, which will allow the client to do other things while e.g. waiting for an order to be ready or an asset to be activated. Therefore, add concurrency to the lowest level of http interaction.

jreiberkyle:
Switch v2 orders client to async. Also adds sophisticated order details handling (definition of products, tools, notifications, and delivery into a single order detail).

Note, this is just the python library. CLI will be implemented in #244.

Closes #238

TODOs that will be addressed in the future: make all docs markdown docs #242, and convert testing/docbuilds to nox #243


jreiberkyle:
Functionality is implemented, fully tested. Todo: update documentation

jreiberkyle:
@sarasafavi @eblossom @strixcuriosus - A great place to start, and the place with the highest impact if your time is limited, is with the python library code sample in the README. Two usability questions I am considering:
1. Whether APlanetSession, AOrdersClient and even the orders_details pieces (OrdersDetails, Product, Tool) should be elevated to import straight from `planet` instead of `planet.api` in the case of `APlanetSession` and `AOrdersClient` and `planet.api.orders_details` for `OrdersDetails`, `Product`, and `Tool`. Currently leaning towards elevating though I'm not sure if this will conflict with other api clients. Maybe even rearranging so the import is `planet.orders` for all orders_details-related classes.
2. Whether APlanetSession and AOrdersClient should keep the A prefix, which indicates they are async and keeps the door open for implementing sync versions, or drop the A because it is a new naming convention. Currently leaning towards keeping.

README.md
```diff
@@ -28,31 +28,134 @@ flag is highly recommended for those new to [pip](https://pip.pypa.io).
 A PEX executable (Windows not supported) and source releases are
 [here](https://github.com/planetlabs/planet-client-python/releases/latest).
 
-## Documentation
+## API Key
 
-Online documentation:
-https://planetlabs.github.io/planet-client-python/index.html
+The planet API requires an account for use. [Signup here](https://www.planet.com/explorer/?signup).
```
ericrdunham, line 33: The phrase "planet API" does not seem consistent with the previously-mentioned "Planet's public APIs". What about "Planet's public APIs require an account for use."?

ericrdunham, line 33: "Signup here" appears to want the phrasal verb, i.e. "Sign up here".

README.md
```diff
@@ -28,31 +28,134 @@ flag is highly recommended for those new to [pip](https://pip.pypa.io).
...
 
-Documentation is also provided for download
-[here](https://github.com/planetlabs/planet-client-python/releases/latest).
+This can be provided via the environment variable `PL_API_KEY` or the flag `-k` or `--api-key`.
```
ericrdunham, line 35: Consider linking to the documentation for the CLI options. https://planetlabs.github.io/planet-client-python/cli/reference.html#general-options

README.md
```diff
@@ -28,31 +28,134 @@ flag is highly recommended for those new to [pip](https://pip.pypa.io).
 A PEX executable (Windows not supported) and source releases are
 [here](https://github.com/planetlabs/planet-client-python/releases/latest).
 
-## Documentation
+## API Key
```
ericrdunham, line 31: This 'API Key' section seems sparse from a usability perspective. Consider adding information on how to get at their API key after a `planet init` as well as information on how to set that as an environment variable for use in the following library example, e.g. `export PL_API_KEY=$(jq --raw-output '.key' $HOME/.planet.json)` and maybe the equivalent for `setx` or PowerShell.

README.md
```diff
@@ -28,31 +28,134 @@ flag is highly recommended for those new to [pip](https://pip.pypa.io).
...
+Using `planet init` your account credentials (login/password) can be used to obtain the api key.
 
-## Development
+## Quick Start
 
-To contribute or develop with this library, see
-[CONTRIBUTING](https://github.com/planetlabs/planet-client-python/CONTRIBUTING.md)
+### Python Library
 
+The python library is asynchronous, which greatly speeds up many interactions
```
ericrdunham, line 43: I believe the legal trademark has this stylized as 'Python'.

ericrdunham, line 43: The phrase 'Python Library' is used in the title of the section as well as immediately in the first sentence, which provides a visual stutter. Consider replacing the second occurrence with a pronoun or other phrase e.g. "The client modules are asynchronous...".

README.md
```diff
@@ -28,31 +28,134 @@ flag is highly recommended for those new to [pip](https://pip.pypa.io).
...
+with the planet API.
```
ericrdunham, line 44:
```suggestion
with Planet's public APIs.
```

README.md
```diff
@@ -28,31 +28,134 @@ flag is highly recommended for those new to [pip](https://pip.pypa.io).
...
 
+The python library is asynchronous, which greatly speeds up many interactions
+with the planet API. 
 
-## API Key
+```python
+import asyncio
+import os
+from planet.api import APlanetSession, AOrdersClient
+from planet.api.order_details import OrderDetails, Product
```
ericrdunham, line 50: It seems likely that the name `Product` would be a good candidate for name collisions. I'd suggest namespacing this instead of importing the class directly.

README.md
```diff
@@ -28,31 +28,134 @@ flag is highly recommended for those new to [pip](https://pip.pypa.io).
...
+### Python Library
 
+The python library is asynchronous, which greatly speeds up many interactions
+with the planet API. 
 
-## API Key
+```python
+import asyncio
+import os
+from planet.api import APlanetSession, AOrdersClient
```
ericrdunham, line 49: Does something exist at the level of `import planet` for a user to use instead of nested modules?

README.md
```diff
@@ -28,31 +28,134 @@ flag is highly recommended for those new to [pip](https://pip.pypa.io).
...
 
-This can be provided via the environment variable `PL_API_KEY` or the flag `-k` or `--api-key`.
+image_ids = ['3949357_1454705_2020-12-01_241c']
+order_detail = OrderDetails(
+    'test_order',
+    [Product(image_ids, 'analytic', 'psorthotile')]
+)
 
-Using `planet init` your account credentials (login/password) can be used to obtain the api key.
+async def create_order(order_detail):
```
ericrdunham, line 60: This seems like the sort of thing that might be nice for a batteries-included experience, e.g.

```python
import asyncio

import planet

order = {
    "name":
    "test_order",
    "products": [{
        "image_ids": "3949357_1454705_2020",
        "item_type": "psorthotile",
        "product_bundle": "analytic"
    }]
}

oid = asyncio.run(planet.create_order(planet.get_api_key(), order))
print(oid)
```

README.md
```diff
@@ -28,31 +28,134 @@ flag is highly recommended for those new to [pip](https://pip.pypa.io).
...
+    async with APlanetSession(auth=(API_KEY, '')) as ps:
+        client = AOrdersClient(ps)
+        return await client.create_order(order_detail)
 
+oid = asyncio.run(create_order())
```
ericrdunham, line 65: `TypeError: create_order() missing 1 required positional argument: 'order_detail'`

README.md
```diff
@@ -28,31 +28,134 @@ flag is highly recommended for those new to [pip](https://pip.pypa.io).
...
+print(oid)
+```
+
+Not into async? No problem. Just wrap the library and async operations together
+and call from your synchronous code.
 
-# Example CLI Usage
+```python
+def sync_create_order():
```
ericrdunham, line 73:
```suggestion
def sync_create_order(order_details):
```

README.md
```diff
@@ -28,31 +28,134 @@ flag is highly recommended for those new to [pip](https://pip.pypa.io).
...
+    return asyncio.run(create_order())
+
+print(sync_create_order)
```
ericrdunham, line 76:
```suggestion
print(sync_create_order(order_details))
```

README.md
```diff
@@ -28,31 +28,134 @@ flag is highly recommended for those new to [pip](https://pip.pypa.io).
...
+```
+
+Why async? Because things get *really cool* when you want to work with multiple
+orders. Here's an example of submitting two orders, waiting for them to
+complete, and downloading them. The orders each clip a set of images to a
+specific area of interest (AOI), so they cannot be combined into one order.
+(hint: [Planet Explorer](https://www.planet.com/explorer/)) was used to define
+the AOIs and get the image ids.
```
ericrdunham, line 84:
```suggestion
(hint: [Planet Explorer](https://www.planet.com/explorer/) was used to define
the AOIs and get the image ids.)
```

README.md
```diff
@@ -28,31 +28,134 @@ flag is highly recommended for those new to [pip](https://pip.pypa.io).
...
+
+async def main():
+    async with APlanetSession(auth=(API_KEY, '')) as ps:
+        client = AOrdersClient(ps)
+        await asyncio.gather(
+            create_and_download(iowa_order, client),
+            create_and_download(oregon_order, client)
+        )
+
+asyncio.run(main())
```
ericrdunham, line 153: This specific example felt opaque on the progress. Consider adding a timestamp to the status messages or even a fancy "running for <n> <units-of-time>" that gets updated in-place. Also consider changing the order details for this example so it doesn't take several minutes to complete. :)

README.md
```diff
@@ -28,31 +28,134 @@ flag is highly recommended for those new to [pip](https://pip.pypa.io).
...
+```
+[Example output](example_output.md)
+
+
+## Command-Line Interface (CLI)
```
ericrdunham, line 158:
```suggestion
## CLI Usage
```
The term CLI is defined above. :)

ericrdunham:
Attempting to use this locally (i.e. `planet --help`) immediately explodes with a traceback. Is that expected?

README.md
jreiberkyle: Yes! You are right. This was implemented only in the CLI in v1, I agree it makes total sense to include this in the python lib.

README.md
jreiberkyle: Great idea re: timestamp. And I agree it would be great to have the code example complete faster here. I do not have any insight into ways to decrease order completion time via the order details. I could try by removing the clip tool, though that would make the example a bit less interesting because there is no reason to use two separate orders if they do not have different toolchains.

README.md
jreiberkyle: Ahh shoot. I tested the code example but something must have shifted. Good catch, thanks! (and as a part of the doc switch to markdown, I am going to implement code tests in markdown docs - let me know if you have any good leads on packages for this)

README.md
jreiberkyle: In v1, there was nothing at the level of `import planet`. I haven't changed that yet in v2, but I'd like to. My only blocker is that I haven't tackled the CLI yet, so I don't know if that would cause any conflicts with the CLI. But honestly, it is unlikely to cause issues and the usability advantage would make any extra work with the CLI, if it does come up, worth it.

README.md
jreiberkyle: https://github.com/planetlabs/planet-client-python/issues/248

jreiberkyle:
yep. cli isn't implemented yet (see #244). I apologize, that wasn't clear from the description of this PR.

README.md
jreiberkyle: good point re: conflicts. A nice usage would be to import this directly from `planet` so maybe a name change to `OrderProduct` (and `OrderTool` etc) would make sense.

README.md
strixcuriosus: +1 for enabling top level imports

README.md
jreiberkyle: or maybe keep this namespaced. I just don't really like the namespace `order_details`

planet/api/http.py
```diff
@@ -13,135 +13,170 @@
...
-#     # if debug level set then its nice to see the headers of the request
-#     if log_level == logging.DEBUG:
-#         http_client.HTTPConnection.set_debuglevel(1)
-#     else:
-#         http_client.HTTPConnection.set_debuglevel(0)
-#
 
-def _log_request(req):
-    LOGGER.info('%s %s %s %s', req.method, req.url, req.params, req.data)
+class APlanetSession():
```
strixcuriosus, line 30: What are your thoughts on renaming this to something like `AsyncPlanetSession` to be more explicit and descriptive?

planet/api/order_details.py
```diff
@@ -0,0 +1,539 @@
...
+
+from .. import specs
+
+
+class OrderDetailsException(Exception):
+    """Exceptions thrown by OrderDetails"""
+    pass
+
+
+class OrderDetails():
```
strixcuriosus, line 28: How do you plan to keep classes like these up to date with changing API specs?

README.md
ericrdunham: Unfortunately, I don't know of a tool that allows you to run/test Python code inside Markdown documents. :-( The solutions that I can think of right now include:
- extract the code to a Python file and run it as part of the test suite and then link to that file in the README (or copy-paste it for inline usability and put a note at the top of the Python file to include it in the README)
- use reStructuredText for the README and hopefully the `include` directive is supported (which goes along with the previous suggestion
- use the output-linking style for both, such that there would be something like: `Here is an [example script to create an order](examples/create_order.py) along with [an example of what the output might look like](examples/create_order_output)`

Unfortunately I don't know of a cleaner way to solve this that doesn't involve extracting this code to a separate Python file.

README.md
jreiberkyle: Those look really nice from a user perspective. I also like removing the redundant `Planet` from `planet.ASession`. I'll go ahead and implement this change.

README.md
jreiberkyle: Yeah, timestamps sound good! I have a fancy in-place visual cue implemented as a part of the orders client for downloading orders. We could consider adding similar functionality for polling orders as well. If that sounds like it would make the user experience better, I can write up an issue to implement it.

Side question: for the fancy in-place terminal updates for downloads (and potentially for polling), would you expect displaying the updates to be the default behavior, disabled by a `quiet` flag or to have it be disabled by default and enabled with a `verbose` flag?

planet/api/http.py
jreiberkyle: Yeah, this is a good question. One I have as well. All the async stuff I have prepended with `A`. A lot of libraries either use `A` or `Async`. I am ambivalent - one is shorter, the other is more explicit. If it would help you or anyone else orient to what's going on, I am fully on-board with moving to `Async`.

planet/api/order_details.py
jreiberkyle: Manual updates that require a whole release cycle. Adding or changing the name of Items/Product Bundles will be easy to update via a list at the top. Anything else will require us to get our hands dirty and change the class associated with whatever change happens in the API. It's a pain, but honestly the API shouldn't change much. If it does, it either brakes our client or breaks the massive JSON blocks all the users of the client would have to build if we didn't have these classes. Do you have any thoughts on making this easier on us and users?

planet/api/http.py
jreiberkyle: @ericrdunham do you have any thoughts on this from your experience orienting to this library?

README.md
jreiberkyle: The option of moving the long example to a separate file that can be run as a part of the test suite makes a lot of sense. We are moving to markdown for all of the documentation and there appears to be a python library (or more) that can help with doctesting python code in markdown. This work will be covered in #249.

README.md
jreiberkyle: Awesome! We are all on the same page! I'm on it.

README.md
```diff
@@ -28,31 +28,134 @@ flag is highly recommended for those new to [pip](https://pip.pypa.io).
...
+        [-117.558734, 45.229745]]]
+}
+
+oregon_images = [
+    '20200909_182525_1014',
+    '20200909_182524_1014'
+]
+oregon_order = OrderDetails(
+    'oregon_order',
+    [Product(oregon_images, 'analytic', 'PSScene4Band')],
```
strixcuriosus, line 131: I find building up these bespoke request objects to be a bit cumbersome. It would be great if the python client could provide some simplified abstractions (not necessarily in this already quite large PR).

As one example, we could have an order utility that accepts an AOI and TOI, does a search, and places an order.

README.md
jreiberkyle: This is a great point. We have functionality for storing the API key via the CLI but no functionality for getting it. The primary use case for pulling the api key from the environmental variable is for use in a Docker container (as done in [planet notebooks](https://github.com/planetlabs/notebooks)). The api key is passed as a command-line argument when the Docker container is built. What about adding a CLI command for retrieving the api key so users and docs don't have to hard-code the path to the planet auth information?

So CLI commands for interacting with auth would be e.g. 
storing api key
`planet auth --init`

retrieving api key
 `planet auth --key`

usage example:
`export PL_API_KEY=$(planet auth --key)`

planet/api/order_details.py
strixcuriosus: One option is to generate classes from the swagger specs with a tool such as https://github.com/swagger-api/swagger-codegen

You have more flexibility and room for simplified abstractions by defining your own classes, though.

strixcuriosus: My main concern can be expressed as wanting a one liner to place an order if I have an AOI and know which assets/item_types I want.

strixcuriosus:
Great use of asyncio to parallelize ordering 🎉

planet/api/http.py
```diff
@@ -13,135 +13,170 @@
...
-        self._session.close()
+    async def aclose(self):
+        await self._client.aclose()
 
-    def request(self, request, retry_count=5):
-        '''Submit a request with retry.
+    async def retry(self, func, *a, **kw):
+        '''Run an asynchronous request function with retry.'''
+        retry_count = self.retry_count
+        wait_time = self.retry_wait_time
```
strixcuriosus, line 62: instead of a static wait time, this could use exponential backoff

README.md
jreiberkyle: Would love additional discussion in #248. For now, for this PR, I'll mark this as resolved.

README.md
jreiberkyle: Agreed, and saved in #250. Given that we just have the orders client implemented now and don't have that higher level functionality, how is the feel of building up the request object this way? The old, JSON blob way is also supported. Is this an improvement over the JSON blob?

planet/api/http.py
jreiberkyle: ohhhhh yes, I agree fully! Actually, when the next version of httpx is released, retries with exponential backoff will be supported. I had that documented close-by but should also document that here.

README.md
ericrdunham: That sounds great to me! My vote would be for `planet auth --key` to explode with a non-zero return if it's unset, similar to `set +u` or `set +o nounset`.

README.md
```diff
@@ -28,31 +28,134 @@ flag is highly recommended for those new to [pip](https://pip.pypa.io).
...
+oid = asyncio.run(create_order())
+print(oid)
+```
+
+Not into async? No problem. Just wrap the library and async operations together
+and call from your synchronous code.
 
-# Example CLI Usage
+```python
+def sync_create_order(order_details):
```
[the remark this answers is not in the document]
jreiberkyle: I hear you. I agree, this would be nice. Here's the deal with async and sync - really, if we want to support sync in our library, we should reimplement all the async commands as sync (and they shouldn't be each running an event loop - why? right now all I can say is it's frowned upon, maybe it is costly to spin up an event loop and will slow things down?). Implementing sync versions of every public function (and the lower level functionality as well) is a lot of redundancy to take on. I now remember that I was going to add an issue to track requests for synchronous support within the library and then link to it in the README so users can weigh in and we can assess when/if it makes sense to add synchronous support.

Added #251 and updated README to include more information about usage of `asyncio.run` and a link to the ticket.

Does that feel adequate for now?

README.md
jreiberkyle: I just realized I had assumed you were just talking about the polling portion of this code. Was I correct in that or is there a way the download portion could be improved as well?

planet/api/http.py
ericrdunham: **TL;DR: let the namespace do the thing, and not the name of the class/function/whatever.**

My thoughts would be to remove it entirely and have the documentation push the fact that it's async over and over! The named classes in Python's `asyncio` module don't have any sort of prefix, and I can imagine myself as a user going "okay, I get it" after the umpteenth time of typing it out. :)

I was thinking yesterday about having something like `import planet.sync as planet` for people that didn't want to use the async stuff, and basically everything in there would be a generated/decorated version of everything in the base `planet` module, but it would all be decorated/returned with `asyncio.run`, which would allow a user to do stuff like:

```python
import asyncio

import planet

order = ...
oid = asyncio.run(planet.create_order(planet.get_auth(), order))
print(oid)
```

... as well as ...

```python
import planet.sync as planet

order = ...
oid = planet.create_order(planet.get_auth(), order)
print(oid)
```

I hope that helps!

planet/api/order_details.py
jreiberkyle: **Thank you** for making me aware of this amazing tool! I am checking it out further. I'm also with you on wanting a one-liner. Is there any chance you could provide an ideal usage example?

README.md
jreiberkyle: Awesome! This work will be done in #252

planet/api/http.py
jreiberkyle: @ericrdunham this pattern looks really good to me vs APlanetSession or AsyncPlanetSession (or ASession/AsyncSession). @strixcuriosus how does it look to you?

One note, we don't actually have a function `planet.create_order()` and I don't see us having one, at least for async operations. This is because communication via the api must be done within the context of a session (I realize it may be beneficial to document why this is the case in our own documentation. For now, I'll reference [httpx](https://www.python-httpx.org/advanced/)). For sync it is simply inefficient to not group all communication within a session (due to having to build/destroy the session each time). For async all concurrent communication must be grouped within a session because there can only be one active session. I'm going to add this information to the docs.

I am also leaning toward keeping specific client functions namespaced and not elevating them. So, a usage I'd propose is:

```python
import planet

order = ...
with planet.Session as session:
    cl = planet.OrdersClient(session)  # there will be other clients, DataClient, TaskingClient, etc..
    oid = cl.create_order(planet.get_auth(), order)
```

jreiberkyle: adding additional information regarding usage of the session will be covered as a part of #242

README.md
jreiberkyle: https://github.com/planetlabs/planet-client-python/pull/240/commits/a33ac22592eb00829cf188379e1f88c434a17928

README.md
jreiberkyle: I am going to just break this off as a different effort, in-place updates with status and duration: #253

planet/api/order_details.py
jreiberkyle: looking into codegen: #254, one-liner: #250

jreiberkyle: looking into codegen: #254, one-liner: #250

README.md
jreiberkyle: Ok, in the interest of keeping this PR rolling I am going to mark this as resolved with the new issue. I am still interested in hearing if you see other opportunities for improvement. If you have thoughts, new issue with suggestions is *very* welcome!

README.md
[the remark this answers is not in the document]
jreiberkyle: Marking this as resolved. Feel free to follow up either inside that ticket or with a new ticket for continued discussion.

README.md
ericrdunham: **TL;DR: I prefer dictionaries over classes for this sort of thing. :)**

I personally prefer the JSON blobs and dictionaries, to avoid the cost of JSON <--> `dict` <--> class as well as the slightly slower dot attribute lookup compared to the O(1) of a `dict` lookup. What I'm saying might be a premature optimization, but I like the parallel-ish UX of something like:

```shell
curl -X POST --u ${PL_API_KEY}: -H '<json>' <create-order-endpoint> -d @order.json
```

```shell
planet orders create --json=@order.json
```

```python
import planet

def read_json(path: str) -> str:
    with open(path, "r") as f:
        return f.read()

oid = planet.create_order(read_json("order.json"))
print(oid)
```

I feel like there might be two different use cases here, and I'm not sure how to address all the things, but there's one case where using classes for examples and stuff is super nice, but then there's the case where folks are spinning up millions of these things and pushing them at the API and walking away, and so the class-based ergonomics might not be as beneficial or valuable.

Here's an example of what I'm trying to communicate :)

```python
from __future__ import annotations
import dataclasses
import dis
import functools
import json
import timeit
import typing


@dataclasses.dataclass
class SerializableDataclass:

    @classmethod
    def fromdict(cls, data: dict) -> SerializableDataclass:
        return cls(**data)

    @classmethod
    def fromjson(cls, data: str) -> SerializableDataclass:
        return cls.fromdict(json.loads(data))

    def asjson(self) -> str:
        return json.dumps(dataclasses.asdict(self))


@dataclasses.dataclass
class Product(SerializableDataclass):
    image_ids: typing.List[str]
    item_type: str
    product_bundle: str


@dataclasses.dataclass
class Order(SerializableDataclass):
    name: str
    products: typing.List[Product]


order_json = """
    {
        "name": "test_order",
        "products": [{
            "image_ids": "3949357_1454705_2020",
            "item_type": "psorthotile",
            "product_bundle": "analytic"
        }]
    }
"""


def test_order_dict_name(order_json: str) -> str:
    order = json.loads(order_json)
    return order["name"]


def test_order_class_name(order_json: str) -> str:
    order = Order.fromjson(order_json)
    return order.name


def main():
    print(dis.dis(test_order_dict_name))
    print(dis.dis(test_order_class_name))
    timeme = functools.partial(timeit.timeit, number=10_000_000)
    dict_time = timeme(lambda: test_order_dict_name(order_json))
    cls_time = timeme(lambda: test_order_class_name(order_json))
    print(f"dict = {dict_time}")
    print(f"cls = {cls_time}")


if __name__ == "__main__":
    main()
```

And this is the output I got from one run of that:
```shell
 51           0 LOAD_GLOBAL              0 (json)
              2 LOAD_METHOD              1 (loads)
              4 LOAD_FAST                0 (order_json)
              6 CALL_METHOD              1
              8 STORE_FAST               1 (order)

 52          10 LOAD_FAST                1 (order)
             12 LOAD_CONST               1 ('name')
             14 BINARY_SUBSCR
             16 RETURN_VALUE
None
 56           0 LOAD_GLOBAL              0 (Order)
              2 LOAD_METHOD              1 (fromjson)
              4 LOAD_FAST                0 (order_json)
              6 CALL_METHOD              1
              8 STORE_FAST               1 (order)

 57          10 LOAD_FAST                1 (order)
             12 LOAD_ATTR                2 (name)
             14 RETURN_VALUE
None
dict = 31.057720659009647
cls = 44.18623552902136
```

And there's a bug in this because the `Product` class as a type of the `products` member of `Order` is still a list of `dict`s because only top-level deserialization is here. Thar be dragons! :)

planet/api/http.py
ericrdunham: @jreiberkyle I apologize for any confusion! I was mentioning `planet.create_order` as a example convenience thing to maybe make things easier for some people in the future, but not something that at all was expected soon-ish :). I was noodling on the two separate use cases of someone who wants to play around with the library and see what's what as opposed to someone who already knows that they want to use it and use it a lot. :)

I was thinking of the different between something like `requests.get` and the [more advanced `requests.Session` usage](https://requests.readthedocs.io/en/master/user/advanced/#session-objects). The thought for me was that `planet.create_order` could delegate to a new session and a new client for the easy use case, but if someone wants to make ten million orders, then the persistent session and persistent client (i.e. not using `with`) could be valuable and the existing request/session pattern seems like it's already pretty well-established to me, so piggybacking on that would hopefully be less of a learning curve for folks.

With some code:
```python
import planet

# an example for tutorials, notebooks, etc. maybe?
order = ...
oid = planet.sync.create_order(planet.get_auth(), order)  # or the equivalent with e.g. asynio.run(...)
print(oid)

# an example for batch stuff maybe?
orders = read_ten_million_orders_from_somewhere()
session = planet.Session(planet.get_auth())
client = planet.OrdersClient(session)
oids = await asyncio.gather(map(client.create_order, orders))
print(oids)
```

jreiberkyle: Yeah these convenience functions are nice, I agree. And would be a nice enhancement and is captured in #255. Using a Session as a context manager is the recommended approach with httpx (the http library we build upon) and I feel comfortable having this be the recommended approach with our library. The context manager takes care of cleanup. Without it, the user has to remember to do cleanup on their own. That being said, using the session as a persistent session is actually possible as things are implemented. I will make a note to add documentation on this usage (captured in #242)

README.md
jreiberkyle: @ericrdunham this is some great feedback I am processing here. 

First, in mentioning the bug, are you talking about the code you've written here or in the client code?

Currently, the creation of an order can be done with a JSON blob as well as classes. That would address the user experience (note to self: document this better #242). But internally the JSON blob is converted into classes. So that will slow things down. If I am reading this right, the performance hit is 13 seconds in 10 million runs. That doesn't sound so bad but also that performance hit is between 1/3 and 1/2 of the optimized runtime. So that sounds a little worse. Seems worthwhile to at least put up a ticket to consider that optimization.

The ultimate goal with all of the `order_details` module is to help the user construct an order without having to parse all of the API docs (and cross-reference compatible product bundles and item types). I am very open to any ideas about different ways to achieve this!

jreiberkyle: ok, I would love to continue discussion on how we can improve order specification in #254. It is valuable and also big enough to warrant it's own ticket and effort. For this PR I'll consider the work captured and thus resolved here.

scripts/aorders_api.py
```diff
@@ -0,0 +1,98 @@
...
+
+async def main():
+    start = time.time()
+    auth = (API_KEY, "")
+    async with APlanetSession(auth=auth) as ps:
+        client = AOrdersClient(ps)
+        await asyncio.gather(
+            list_orders(client),
+            list_orders(client, limit=5)
+            # create_and_download(client),
```
strixcuriosus, line 85: leftovers?

planet/api/orders.py
```diff
@@ -42,24 +41,16 @@ class OrdersClientException(Exception):
...
+    :param session: Open session connected to server
+    :type session: planet.api.http.Session
     :param base_url: The base URL to use. Defaults to production orders API
         base url.
     :type base_url: int, optional
     """
-    def __init__(self, api_key=None, base_url=BASE_URL):
-        api_key = api_key or auth.find_api_key()
-        self.auth = api_key and auth.APIKey(api_key)
+    def __init__(self, session, base_url=BASE_URL):
```
strixcuriosus, line 52: For easier usability, it would be nice to make the session arg optional and to instantiate a new session if one isn't provided:
e.g.
```
    def __init__(self, session=None, base_url=BASE_URL, api_key=None):
        if session is None:
            api_key = api_key or somehow_find_api_key()
            session = planet.Session(auth=(api_key, ''))
        self._session = session
```
to allow the user to then do this
```
client = planet.OrdersClient(api_key=some_key)
asyncio.run(client.create_order(order_details))
```

README.md
ericrdunham: @jreiberkyle the bug is in the example code I wrote because it doesn't deserialize child `Product` class instances that are supposed to be in the `Order` instance. Instead of `typing.List[Product]`, they are effectively `typing.List[typing.Mapping[str, str]]` at the moment. :-/ 

The performance hit there was only for single key/attribute lookup, for what that's worth, and I don't know if that's representative of what a customer would want to do with it. :)

The question I suppose I'd have here is, would users want this? My first thought was to advocate for JSON and `dict`s over classes and say something about the Zen of Python like:
> There should be one-- and preferably only one --obvious way to do it.
... but I feel as though that's because of my workflows and performance concerns and how I've preferred to do things.

I'd like to chat about this asynchronously over e-mail or direct messages, and as context for that conversation I'd really like to see what user feedback we have about what folks would like to see in a library for this sort of thing and use more of that type of feedback to push for changes. Thoughts @jreiberkyle @strixcuriosus @SaraPlanet?

planet/api/orders.py
ericrdunham: @strixcuriosus it seems to me like this approach crosses boundaries between the notions of sessions and clients, which looks like it could be confusing and error-prone. My understanding is that a session is the thing responsible for holding the authentication information and doing stuff, and the client delegates some pre-determined logical flows to the session, assuming that the session knows things about API keys and what-not.

For example:
```python
with planet.Session(auth=planet.get_auth()) as session:
    with planet.OrdersClient(base_url=a_shiny_url) as client:
        oid = asyncio.run(client.create_order(order))
```
There's a bug there, and it wouldn't show up immediately. :-/

I personally _would_ like to see the batteries-included approach with top-level things, like Requests and other libraries do, e.g. `planet.create_order` would create a default `Session` _and_ a default `Client` for you, but not doing the mixing and matching of sessions and clients.

README.md
```diff
@@ -28,31 +28,159 @@ flag is highly recommended for those new to [pip](https://pip.pypa.io).
...
 
-Online documentation:
-https://planetlabs.github.io/planet-client-python/index.html
+Planet's APIs require an account for use.
+[Sign up here](https://www.planet.com/explorer/?signup).
 
-Documentation is also provided for download
-[here](https://github.com/planetlabs/planet-client-python/releases/latest).
+Authentication information can be obtained and stored locally with the CLI
+`planet init` command. Both the Python library and CLI will default to using
```
ericrdunham, line 37: This documented behavior of `planet init` is not synchronized with the actual behavior in this branch. What are thoughts about getting the behavior and the documentation to play nice?

README.md
```diff
@@ -28,31 +28,159 @@ flag is highly recommended for those new to [pip](https://pip.pypa.io).
...
+the stored authentication information, if available (not yet implemented, see
+issues #244 and #248).
 
+Additionally, authentication information can be provided to the CLI by
+storing the api key in the environmental variable `PL_API_KEY` or by specifying
```
ericrdunham, line 42: Consider standardizing the stylizing of 'API' to uppercase (code, CLI flags, etc. notwithstanding).

ericrdunham, line 42: I think "environmental variable" could be confusing as a non-standard term compared to "environment variable".

README.md
```diff
@@ -28,31 +28,159 @@ flag is highly recommended for those new to [pip](https://pip.pypa.io).
...
 
-To contribute or develop with this library, see
-[CONTRIBUTING](https://github.com/planetlabs/planet-client-python/CONTRIBUTING.md)
+### Python Library Usage
 
+The client modules within the Python library are asynchronous, which greatly
+speeds up many interactions with Planet's APIs. Support for asynchronous
+development is native to Python 3.6+ via the
+[`asyncio` module](https://docs.python.org/3/library/asyncio.html) and is
+fairly easy to implement.
```
ericrdunham, line 54: "fairly easy to implement" could come across as a little harsh, especially for our users that are new to the space or who may be struggling to grasp these concepts initially. I think we want them to feel encouraged and empowered instead. What do you think about saying something like "Support for asynchronous development is native to Python 3.6+ via the [`asyncio` module](...). For the sake of brevity, here is a starting point if you feel unfamiliar or would like some more information about asynchronous programming in Python to help get you started: https://project-awesome.org/timofurrer/awesome-asyncio."?

README.md
```diff
@@ -28,31 +28,159 @@ flag is highly recommended for those new to [pip](https://pip.pypa.io).
...
+API_KEY = os.getenv('PL_API_KEY')
+
+image_ids = ['3949357_1454705_2020-12-01_241c']
+order_details = planet.OrderDetails(
+    'test_order',
+    [planet.Product(image_ids, 'analytic', 'psorthotile')]
+)
+
+async def create_order(order_details):
+    async with planet.Session(auth=(API_KEY, '')) as ps:
```
ericrdunham, line 71: This appears to be out-of-date based on things like the `planet.get_auth()` conversations we've had recently, and similarly above with the `API_KEY` assignment and `os` import.

README.md
```diff
@@ -28,31 +28,159 @@ flag is highly recommended for those new to [pip](https://pip.pypa.io).
...
+print(oid)
+```
 
-Using `planet init` your account credentials (login/password) can be used to obtain the api key.
+When using `asyncio.run` to develop synchronous code with the async library,
+keep in mind this excerpt from the
+[asyncio.run](https://docs.python.org/3/library/asyncio-task.html#asyncio.run)
+documentation:
+
+"*This function always creates a new event loop and closes it at the end. It
```
ericrdunham, line 95: Consider using a Markdown blockquote here instead of the custom italics and double quotes.

README.md
ericrdunham: Argh, I think the internet swallowed my reply. :) For the inline polling, I think the timestamp could be helpful for an example, because that first order can take quite a while, especially for a user that isn't expected it. With regard to CLI functionality only, I personally would expect the default to show indicators and then get silenced with a `--quiet` or `--silent` or something tantamount. :)

README.md
```diff
@@ -28,31 +28,159 @@ flag is highly recommended for those new to [pip](https://pip.pypa.io).
...
+    oid = await client.create_order(order_detail)
+    print(oid)
+    state = await client.poll(oid, verbose=True)
+    print(state)
+    filenames = await client.download_order(oid, progress_bar=True)
+    print(f'downloaded {oid}, {len(filenames)} files downloaded.')
+
+
+async def main():
+    async with planet.Session(auth=(API_KEY, '')) as ps:
```
ericrdunham, line 171: I have the same concern here as above with regard to synchronization and using the abstract function call instead of the authentication tuple.

planet/api/orders.py
strixcuriosus: +1 to something like a top-level easy to use `planet.create_order` 
My main concern is that this seems harder to use than I think it needs to be.

strixcuriosus: I think a `planet.bulk_create_orders` that leveraged the async client calls under the hood would be great too

planet/api/__init__.py
```diff
@@ -12,9 +12,11 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+from .http import Session
 from .orders import OrdersClient
 from .__version__ import __version__  # NOQA
```
ericrdunham, line 17: This looks like it clobbers the module and makes `planet.api.__version__` a `str`, which can be confusing and/or explosive depending upon which order things are imported. For example:
```
>>> from planet.api.__version__ import __version__ as v
>>> type(v)
<class 'str'>
>>> v
'2.0.0'
>>> import planet.api.__version__ as v2
>>> v2.__version__
  File "<stdin>", line 1, in <module>
AttributeError: 'str' object has no attribute '__version__'
>>> 
```

ericrdunham, line 17: This import is also unused and can be removed.

planet/__init__.py
```diff
@@ -11,3 +11,26 @@
...
 # See the License for the specific language governing permissions and
 # limitations under the License.
+from .api.http import Session
+from .api.models import Order
+from .api.orders import OrdersClient
+from .api.order_details import (
+    OrderDetails, Product, Notifications, Delivery, AmazonS3Delivery,
+    AzureBlobStorageDelivery, GoogleCloudStorageDelivery,
+    GoogleEarthEngineDelivery, Tool)
+from .api.__version__ import __version__  # NOQA
```
ericrdunham, line 21: This import appears unused and can be removed.

planet/api/http.py
```diff
@@ -13,135 +13,176 @@
...
-#     if log_level == logging.DEBUG:
-#         http_client.HTTPConnection.set_debuglevel(1)
-#     else:
-#         http_client.HTTPConnection.set_debuglevel(0)
-#
 
-def _log_request(req):
-    LOGGER.info('%s %s %s %s', req.method, req.url, req.params, req.data)
+class Session():
+    '''Context manager for asynchronous communication with the Planet server.
```
ericrdunham, line 31: What are thoughts around something like `s/Planet servers?/Planet APIs/gi`?

planet/api/http.py
```diff
@@ -13,135 +13,176 @@
...
 
+    Authentication for Planet servers is given as ('<api key>', '').
```
ericrdunham, line 33: Did we want this to reference something a little more abstract like `planet.get_auth()`?

planet/api/http.py
```diff
@@ -13,135 +13,176 @@
...
+    :param auth: Planet server authentication.
+    :type auth: httpx.Auth or tuple.
+    '''
 
-    def __init__(self):
-        # general session for sync api calls
-        self._session = RedirectSession()
-        self._session.headers.update({'User-Agent': self._get_user_agent()})
-        self._session.verify = USE_STRICT_SSL
+    def __init__(self, auth=None):
```
ericrdunham, line 39: Consider adding typing for `auth`.

planet/api/http.py
```diff
@@ -13,135 +13,176 @@
...
+        await self.aclose()
 
-    def close(self):
-        self._session.close()
+    async def aclose(self):
+        await self._client.aclose()
 
-    def request(self, request, retry_count=5):
-        '''Submit a request with retry.
+    async def retry(self, func, *a, **kw):
```
ericrdunham, line 59: Consider adding typing information here for this function.

planet/api/http.py
```diff
@@ -13,135 +13,176 @@
...
         for i in range(max_retry):
             try:
-                resp = self._do_request(request)
-                return resp
+                return await func(*a, **kw)
             except exceptions.TooManyRequests:
                 if i < max_retry:
                     LOGGER.debug(f'Try {i}')
-                    LOGGER.info('Too Many Requests: sleeping and retrying')
+                    LOGGER.info(f'Too Many Requests: sleeping {wait_time}s')
```
ericrdunham, line 77: This doesn't seem to provide enough information for a user to being able to dig in to which requests are being throttled. Consider adding more identifying information like verb, URI, and/or a temporary (shortened) SHA to provide more filtering criteria on these log entries.

planet/api/http.py
```diff
@@ -13,135 +13,176 @@
...
                     # TODO: consider exponential backoff
                     # https://developers.planet.com/docs/data/api-mechanics/
-                    time.sleep(self.retry_wait_time)
+                    await asyncio.sleep(wait_time)
         raise Exception('too many throttles, giving up')
```
ericrdunham, line 81: Consider using a more narrowly-scoped error, e.g. `HTTPError`.

planet/api/http.py
```diff
@@ -13,135 +13,176 @@
...
 
-    def _do_request(self, request, **kwargs):
+    async def request(self, request, stream=False):
```
ericrdunham, line 83: Consider adding typing to this function.

planet/api/http.py
```diff
@@ -13,135 +13,176 @@
...
-                    prepared_request.prepare_url(
-                        prepared_request.url, {
-                            'api_key': key.group(1)
-                        }
-                    )
+        http_resp = await self._client.send(request.http_request,
+                                            stream=stream)
+        return models.Response(request, http_resp)
+
+    def stream(self, request):
```
ericrdunham, line 107: Do we think that this happens often enough for this function? The reason I ask is because the pattern appears inconsistent and can be confusing. That is to say, encouraging users to `with planet.Session(...) as session: with planet.OrdersClient(session) as oc: ...` feels more consistent to me than `with ... as session: with session.stream(req): ...`. I'm curious about what benefit this provides over something like `with ... as session: with planet.Stream(session, req) as stream: ...`?

ericrdunham:
For the sake of not making more noise than I already have, are there thoughts in general around typing functions and such? I'm going to stop commenting on every function for now. :)

planet/api/http.py
```diff
@@ -13,135 +13,176 @@
...
+            session=self,
+            request=request
+        )
 
     @staticmethod
-    def _is_subdomain_of_tld(url1, url2):
-        orig_host = urlparse(url1).hostname
-        re_host = urlparse(url2).hostname
-        return orig_host.split('.')[-2:] == re_host.split('.')[-2:]
+    def _get_user_agent():
```
ericrdunham, line 121: Consider raising this to a class or module-level constant instead of a non-public static method for UX as well as performance.

planet/api/http.py
```diff
@@ -13,135 +13,176 @@
...
+    @staticmethod
+    async def _raise_for_status(response):
+        # TODO: consider using http_response.reason_phrase
+        status = response.status_code
+
+        if status < 300:
+            return
+
+        exception = {
+            400: exceptions.BadQuery,
```
ericrdunham, line 144: Consider using the `HTTPStatus` constants here since they're more explicit than numbers.

planet/api/http.py
```diff
@@ -13,135 +13,176 @@
...
+        }.get(status, None)
+
+        try:
+            msg = response.text
+        except httpx.ResponseNotRead:
+            await response.aread()
+            msg = response.text
+
+        # differentiate between over quota and rate-limiting
+        if status == 429 and 'quota' in msg.lower():
```
ericrdunham, line 159: I believe this branch can be eliminated with something like `lambda m: exceptions.OverQuota(m) if 'quota' in m.lower() else exceptions.TooManyRequests(m)` in the key lookup for `exception` in the anonymous `dict`.

planet/api/http.py
```diff
@@ -13,135 +13,176 @@
...
+            return
+
+        exception = {
+            400: exceptions.BadQuery,
+            401: exceptions.InvalidAPIKey,
+            403: exceptions.NoPermission,
+            404: exceptions.MissingResource,
+            429: exceptions.TooManyRequests,
+            500: exceptions.ServerError
+        }.get(status, None)
```
ericrdunham, line 150: I think if the `None` default here were changed to something like `lambda m: exceptions.APIException(f'{status}: {m}')`, L#162-165 could be replaced with `raise exception(msg)` by itself.

planet/api/http.py
```diff
@@ -13,135 +13,176 @@
...
+        if status == 429 and 'quota' in msg.lower():
+            exception = exceptions.OverQuota
+
+        if exception:
+            raise exception(msg)
+
+        raise exceptions.APIException(f'{status}: {msg}')
+
+
+class APlanetStream():
```
ericrdunham, line 168: Consider renaming this to `Stream` for consistency with the other conversations. :)

planet/api/http.py
```diff
@@ -13,135 +13,176 @@
...
+    :type session: APlanetSession
+    :param request: Request to submit
+    :type request: planet.api.models.Request
+    '''
+    def __init__(self, session, request):
+        self.session = session
+        self.request = request
+
+    async def __aenter__(self):
+        self.response = await self.session.request(
```
ericrdunham, line 181: I feel like I'm misunderstanding something here, but can all instances of this class be replaced with this request instead? It looks like it could save a bit of overhead if that's the case.

planet/api/models.py
```diff
@@ -36,52 +37,43 @@ class Request():
...
-            headers.update({
-                'Authorization': 'api-key %s' % self.auth.value
-            })
-        else:
-            raise exceptions.InvalidAPIKey('No API key provided')
-        return headers
+        self.http_request.url = httpx.URL(url)
 
 
 class Response():
```
ericrdunham, line 79: This class seems like it could be gotten rid of now that `_body` is gone. :)

planet/api/models.py
```diff
@@ -180,105 +134,277 @@ def name(self):
         :returns: name of this resource
         :rtype: str
         '''
-        return utils.get_filename(self.response)
+        name = (_get_filename_from_headers(self.response.headers) or
```
ericrdunham, line 137: These `*filename*` functions don't look like they belong in this module, and I'm curious why they seem to have been moved here?

planet/api/models.py
```diff
@@ -180,105 +134,277 @@ def name(self):
...
+                                 f'wrote {perc}% of {self.total}')
+                    self.previous = new
+
+        unit = 1024*1024
+
+        mode = 'wb' if overwrite else 'xb'
+        try:
+            with open(filename, mode) as fp:
+                _log = _LOG(self.size, 16*unit, filename, disable=progress_bar)
+                with tqdm(total=self.size, unit_scale=True,
```
ericrdunham, line 211: This is a red flag to me along with `_LOG` and I'm curious why TTY-emitting functionality is in the response model's body's write method?

ericrdunham (changes requested):
Please remove all console functionality from models.

planet/api/models.py
```diff
@@ -180,105 +134,277 @@ def name(self):
...
+    :rtype: str
+    """
+    extension = mimetypes.guess_extension(content_type or '') or ''
+    characters = string.ascii_letters + '0123456789'
+    letters = ''.join(random.sample(characters, 8))
+    name = 'planet-{}{}'.format(letters, extension)
+    return name
+
+
+class Paged():
```
ericrdunham, line 261: Are there some concrete examples available for usage of this class? I'm trying to visualize a user workflow leveraging `Paged`, `Order`, and `Orders` from this module. :)

scripts/aorders_api.py
jreiberkyle: yep. Made a note to add this usage information to docs (#242) and removed the script.

planet/api/orders.py
jreiberkyle: Thank you @ericrdunham for providing an example that shows how having the client manage the session could be error-prone. If it wasn't for needing a session that served multiple clients (data, orders, etc), the usage proposed by @strixcuriosus would be awesome. I **hear** the need for top-level, usage. I am on board with implementing this. I could use input from both of you on how to make this top-level usage focused and useful. aka, let's come up with some use cases. I have created #256 to capture these use cases. That will be the starting point for implementing top-level helper functions.

README.md
jreiberkyle: Yeah, good catch. We can be more understanding and supportive here. Updated the wording here to include this awesome (hah!) resource.

README.md
jreiberkyle: the authentication work is captured in #248 - this will be changed there.

README.md
jreiberkyle: the authentication work is captured in #248 - this will be changed there.

planet/api/http.py
jreiberkyle: the authentication work is captured in #248 - this will be changed there.

planet/api/http.py
jreiberkyle: I would like to add typing to all functions (#257). Is there something about this function that raises it's priority?

planet/api/http.py
jreiberkyle: This pattern follows the pattern used in [httpx](https://www.python-httpx.org/async/#streaming-responses), the library we are using for all API calls. So the primary benefit is not having to do a lot of implementation and add confusion to those familiar with httpx by changing the usage pattern. In reality, this is not a function the public should be using, it is used internally to manage receiving and writing information from the server (aka downloading an image). Your confusion here is valuable information. Is there a way we can use documentation to let users know more about this usage pattern?

jreiberkyle:
@ericrdunham 'Please remove all console functionality from models.' can you be more explicit?

planet/api/http.py
jreiberkyle: Noted. good suggestion. #258

planet/api/http.py
jreiberkyle: Ahh, I see below that we are on the same page. All the functions. Agreed. Considered out of scope for this PR and covered in the above-mentioned ticket.

planet/api/models.py
jreiberkyle: A user will work with `Order`. All clients will receive paged responses. They will mostly be the same except they keys for e.g. next link will vary due to variations in the paged responses by the Data, Orders, etc APIs. So `Paged` captures the joint functionality, `Orders` specifies the special Orders API - specific functionality. These two classes will make more sense when the Data API client is implemented. `Orders` is used by `OrdersClient`. I don't expect `Orders` or `Paged` to be a part of a user workflow.

planet/api/models.py
jreiberkyle: I moved it because its only used in this module and I dislike using `utils` as a dumpster for random functionality.

planet/api/http.py
jreiberkyle: good catch. done

planet/api/http.py
jreiberkyle: Hmm... I consider API to be a definition of an interface, whereas a server is the thing hosting the API. So we don't communicate with an API, we communicate with a server.

planet/api/http.py
ericrdunham: I'm curious, but it seems as though the call chain specifically looks like it goes from `Session.stream` to `APlanetStream` to `Session.request`. I feel like I'm missing something because calling `Session.request(..., stream=True)` appears, at first glance, to be able to replace those extra jumps/layers. I want to make sure I'm understanding enough of this to continue commenting on it. :)

ericrdunham:
> @ericrdunham 'Please remove all console functionality from models.' can you be more explicit?

Sure thing! The "console functionality" that I'm thinking of is anything that assumes a console or GUI in the models and the first example (and maybe the only one?) is in `planet.api.models.StreamingBody.write` with the `tqdm` and progress measuring/rendering.

planet/api/http.py
ericrdunham: Okay. I was curious because the language in the README and examples mentions APIs and not servers and so I wanted to ask because I found it a little jarring to read documentation and go "wait, is there a `planet.Server` I missed?!". :)

planet/api/http.py
jreiberkyle: This class manages an async response, both initiating it as a stream and closing it when the context is left. One job, and an important job.

planet/api/models.py
jreiberkyle: I prefer to have an interface between our codebase and the one we rely on. aka, I like to have a `Response` class that wraps the `httpx.Response` class. This is good practice in general, although it is particularly important since `httpx` is still beta, very advanced beta but beta, so it is not guaranteed that the API will not change.

planet/api/models.py
jreiberkyle: I agree, I wasn't happy with the reporting logic being within the write method either. We want to achieve real-time reporting without having to use threading. Real-time reporting means it has to be done within the context of the open file. In sync code (e.g. v1) this was achieved by defining a reporting function elsewhere, and passing into this function as a callback. I'm not certain how callbacks and async behave together. So this has the benefit of being simple. It has the drawback of being somewhat 'polluted'. Given that we can implement an async callback, is that any better than providing reporting within the write function? What implementation would you propose?

Here's the inspiration for the current tqdm usage (from httpx, which we depend upon here): https://www.python-httpx.org/advanced/#monitoring-download-progress

@sarasafavi it would be great to get your take on this as well

jreiberkyle:
@ericrdunham gotcha. Let's hammer this out within the context of that one comment thread.

planet/api/http.py
ericrdunham: I don't feel like my previous question was answered. What value does this provide over directly calling `session.request(request, stream=True)`?

planet/api/models.py
ericrdunham: In that case, how about moving them to somewhere more relevant to their functionality? To me, `planet.api.models._get_filename_from_headers` is more astonishing/surprising than `planet.api.utils._get_filename_from_headers`, for example.

planet/api/models.py
ericrdunham: As far as I'm aware, putting view code in a model is an anti-pattern and is widely and strongly discouraged. Additionally, I'm concerned about the violation of the Principle of Least Astonishment I see immediately as a developer (i.e. I think it would be widely-accepted as surprising to have a `session.request(...)` or `session.stream(...)` try and render progress bars). I haven't run tests/timeit but I'm also concerned about the performance of this at scale as well as the potential for unintentional dark patterns creeping up like Hidden Costs. For an implementation I would suggest to have all the console-ish responsibilities in the CLI.

planet/api/models.py
ericrdunham: In that case, what about having it be a subclass of `httpx.Response`?

planet/api/http.py
jreiberkyle: Good to know! I will keep this in mind when I revisit the documentation in #242. Confusion is always good to note, thanks for the heads up.

planet/api/http.py
jreiberkyle: Yeah, good point. Throwing new `SessionException` in latest commit.

planet/api/http.py
jreiberkyle: `APlanetStream` is a context manager, that's the value it adds. So in addition to calling `Session.request`, it is calling `response.aclose()` when the stream is no longer needed.

ericrdunham: Okay, so in that case it seems like:
```python
response = await session.request(request, stream=True)
body = StreamingBody(response)
return await body.write(...)
```
is the same as...
```python
async with session.stream(request) as response:
    body = StreamingBody(response)
    return await body.write(...)
```
right?

planet/api/http.py
jreiberkyle: Great catch. Changed in http.py and unit tests.

planet/api/http.py
jreiberkyle: calling `request.aclose()` when the context is left

planet/api/models.py
jreiberkyle: I do not see how subclassing will provide for a clean interface that is easy to adapt to changes in `httpx.Response`

planet/api/http.py
ericrdunham: I'm not understanding your most recent response. It looks like if I were to call the request directly instead of using this class, that it would obviate the need for this class, i.e. the response is closed when you call it directly.

planet/api/models.py
ericrdunham: Okay, so then what about using that interface in our tests, so instead of using `httpx.Response` we can use this class?

README.md
sarasafavi: agree, this PR doesn't yet implement CLI behavior. @jreiberkyle let's remove all CLI-related documentation from the PR and we can put it back in down the road.

README.md
sarasafavi: This can be resolved in a future PR.

sarasafavi: yes, let's use `environment variable`

README.md
```diff
@@ -28,31 +28,159 @@ flag is highly recommended for those new to [pip](https://pip.pypa.io).
...
-Documentation is also provided for download
-[here](https://github.com/planetlabs/planet-client-python/releases/latest).
+Authentication information can be obtained and stored locally with the CLI
+`planet init` command. Both the Python library and CLI will default to using
+the stored authentication information, if available (not yet implemented, see
+issues #244 and #248).
 
+Additionally, authentication information can be provided to the CLI by
+storing the api key in the environmental variable `PL_API_KEY` or by specifying
+the api key with the flag `-k` or `--api-key` (TODO: link to cli documentation
```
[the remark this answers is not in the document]
sarasafavi: this should also go away with removal of CLI docs

README.md
sarasafavi: This is stylistic and can be resolved later.

README.md
```diff
@@ -28,31 +28,159 @@ flag is highly recommended for those new to [pip](https://pip.pypa.io).
...
+[asyncio.run](https://docs.python.org/3/library/asyncio-task.html#asyncio.run)
+documentation:
+
+"*This function always creates a new event loop and closes it at the end. It
+should be used as a main entry point for asyncio programs, and should ideally
+only be called once.*"
+
+Do you have a use case where native synchronous support is essential? If so,
+please contribute to
+[Determine need for synchronous support](https://github.com/planetlabs/planet-client-python/issues/251)
```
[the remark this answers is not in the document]
sarasafavi: This is stylistic and can be resolved later.

planet/api/http.py
sarasafavi: the colloquial usage of "Planet server" here is fine and matches the voice of our documentation elsewhere.

planet/api/http.py
sarasafavi: no, I'm fine with `TooManyRequests` just saying "yo dudes too many requests". If a dev wants more info they can wrap their own call with logging.

planet/api/http.py
sarasafavi: agree with @jreiberkyle ... this is an async context manager, and its implementation and usage suggested here is in alignment with python+async norms. This is fine.

planet/api/http.py
sarasafavi: We can revisit this in future development

planet/api/http.py
sarasafavi: We can revisit this in future development

planet/api/http.py
sarasafavi: @ericrdunham I don't follow your question here, circumventing the async context manager would negate the utility of async execution. This is good as-is.

For reference: https://www.python.org/dev/peps/pep-0492/#asynchronous-context-managers-and-async-with

planet/api/models.py
sarasafavi: Agree with @jreiberkyle on this, wrapping the may-change-in-the-future Response is the best for future users; subclassing doesn't address future adaptability. This is fine as-is.

planet/api/models.py
sarasafavi: My feedback might be to raise these methods up a level, but since they're internal-use-only it doesn't really matter. I am strong +1 on _not_ putting them back in a catch-all utils. This is fine as-is.

planet/api/models.py
sarasafavi: I agree that it is a bit muddy (polluted, as @jreiberkyle said) but it's certainly not a red flag or anti-pattern. If future development discovers significant timing issues we can revisit it. At this point in development though this is fine as-is.

planet/api/order_details.py
```diff
@@ -0,0 +1,539 @@
...
+
+    @classmethod
+    def from_dict(cls, details: dict) -> Delivery:
+        '''Create Delivery instance from Orders API spec representation.'''
+        try:
+            details = copy.deepcopy(details)
+            cloud_details = details.pop(cls.cloud_key)
+            cloud_details.update(details)
+            details = cloud_details
+        except AttributeError:
```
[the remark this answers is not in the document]
sarasafavi: it really isn't. https://docs.python.org/3/faq/design.html#how-fast-are-exceptions

Regardless, I am absolutely not concerned about the relative cost of a try/except; this is fine.

sarasafavi (commented):
@jreiberkyle I added a comment elsewhere (lost somewhere in the depths of github, maybe) but my only request is we go ahead and pull out CLI docs pending #244 - otherwise LGTM

jreiberkyle:
great! thanks @sarasafavi , @ericrdunham and @strixcuriosus - your thoughtful reviews and understanding around such a large PR are much appreciated.

PR #245, opened 2021-03-01
Add 'PSScene' item type

ericrdunham:
This pull request adds the 'PSScene' item type. For this functionality to work, the undocumented prefix-matching functionality in the item type matcher was removed. The previously-documented glob-matching style, i.e. `--item-type=psscene*` versus `--item-type=psscene` would be the exclusive way to match multiple item types, save for the 'all` magic string.


sarasafavi:
thank you @ericrdunham !

Issue #224, opened 2020-08-28
Update Travis CI to test newer Python


sarasafavi:
Currently Travis only runs Python 2.7 & Python 3.4 testing; this, at very least, is at odds with our tox config: we should update Travis to also test against Python 3.6 at a minimum.

sgillies:
We've moved to GitHub Actions.

PR #260, opened 2021-03-24
v2: move to nox/github actions from tox/travis-ci

### LINKED ISSUE #243: convert to using nox for testing and github actions for CI
nox is basically tox with a python interface. using nox would allow us to stay in the python language while specifying our tests. nox also allows easy management of environments and supports linting and flake8 and mypy checking. Also can build docs! respx has a nice example [noxfile.py](https://github.com/lundberg/respx/blob/master/noxfile.py)

Nox and travis-ci are somehow tricky to get working together. [Travis-ci is slowing down open-source builds](https://github.com/theacodes/nox/issues/368), making the ci test loop a bit painful. nox and github actions work together [nicely](https://github.com/lundberg/respx/blob/master/.github/workflows/test.yml)! So include a move to github actions for CI as a part of this work.

This solve / nullify #224

jreiberkyle:
Simplify testing and linting by moving functionality from tox and the Makefile to nox. Also move CI from travis to github actions.

Closes #243 and #224


.github/workflows/test.yml
```diff
@@ -0,0 +1,52 @@
...
+on: [pull_request]
+
+jobs:
+  test:
+    name: Test Python ${{ matrix.python-version }}
+    runs-on: ubuntu-latest
+
+    strategy:
+      matrix:
+        python-version: [3.7, 3.8, 3.9]
```
sarasafavi, line 12: you love to see it 😍

setup.py
```diff
@@ -35,25 +24,27 @@
             continue
 
 
+install_requires = [
+    'httpx>=0.16',
+    'tqdm>=4.56',
+    'pywin32 >= 1.0;platform_system=="Windows"'
```
sarasafavi, line 30: can we confirm that this works? I have no way of testing Windows stuff unfortunately. If we want to constrain to Python >3.5 support it might be better to do `pywin >= 300` (ref: https://github.com/mhammond/pywin32/releases/tag/b300)

Makefile
```diff
@@ -1,52 +0,0 @@
-GH_PAGES_SOURCES = planet docs Makefile
```
sarasafavi, line 1: At first glance I was concerned by this file removal, however on further review I'm okay with it. There's potential regression in developer-environment functionality **but** most of that is related to the build-and-release process, which shouldn't be occurring for now from this V2 branch anyway. The real reason I'm okay with this is some of those make targets were already fragile/broken so it's a good time to push for a re-do. It's something we will certainly need to resolve before V2 can be released but it doesn't need to happen _here_.

@jreiberkyle before we wrap this PR can you raise an issue to re-implement Makefile functionality for V2 with nox in mind?

sarasafavi (commented):
@jreiberkyle couple small questions but no big change requests. Following question resolution this LGTM!

ericrdunham (commented):
It's fun to see the new GitHub actions! I didn't have anything else beyond what Sara mentioned. :)

Makefile
jreiberkyle: Great point. Done #262

setup.py
```diff
@@ -35,25 +24,27 @@
...
-    'pytest-cov',
-    'sphinx',
-    'wheel',
 ]
 
 setup(name='planet',
       version=version,
       description=u"Planet API Client",
-      long_description=long_description,
+      long_description=Path("README.md").read_text("utf-8"),
```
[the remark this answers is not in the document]
jreiberkyle: yep, [README.md](https://github.com/planetlabs/planet-client-python/blob/v2-nox-243/README.md) exists in v2 and in this v2 subbranch. That does bring up the point that it would be nice to have the packaging process tested as a part of CI. Created #263

setup.py
jreiberkyle: Good catch, this is a weak point. First, I'll admit this is left over from v1 and hasn't really gotten any attention in v2. The whole packaging and distribution process is being overhauled. As per our offline discussion, we will remove this dependency for now and address supporting windows in the build/distribution process in #264.

PR #269, opened 2021-03-31
Skip mosaic quads that a user does not have access to during downloading

joferkington:
Permission for mosaic quad download often varies spatially for a given user.  It's common for someone to have access to some quads within an AOI but not all.  Currently, when a user tries to download quads they don't have access to the client raises an unexpected error:

```
ERROR - unexpected error in <planet.api.downloader._MosaicDownloadStage object at 0x10d87b240> Traceback (most recent call last):   File
"~/pl/planet-client-python/planet/api/downloader.py", line 159, in _process_task     self._do(self._doing)   File "/~/pl/planet-client-
python/planet/api/downloader.py", line 471, in _do     self._client.download_quad(task, writer)))   File "~/pl/planet-client-python/planet/api/client.py", line
363, in download_quad     download_url = quad['_links']['download'] KeyError: 'download'
```

This PR changes `client.download_quad` to raise a `NoPermission` error when that occurs and changes the `planet mosaics download` command (and associated downloader infrastructure) to skip mosaic quads that a user does not have permission for during downloading.

### DIFF
--- a/planet/api/client.py
+++ b/planet/api/client.py
@@ -16,7 +16,7 @@ import base64
 import json
 from .dispatch import RequestsDispatcher
 from . import auth
-from .exceptions import (InvalidIdentity, APIException)
+from .exceptions import (InvalidIdentity, APIException, NoPermission)
 from . import models
 from . import filters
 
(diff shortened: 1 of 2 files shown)


sarasafavi:
thanks @joferkington !

PR #278, opened 2021-04-29
add config for readthedocs

sarasafavi:
Config file that tells readthedocs how to build our docs: based on https://docs.readthedocs.io/en/stable/config-file/v2.html

Docs build will be available at https://planet-sdk-for-python.readthedocs.io/

For now (while V2 is in development) a manual trigger is required to (re)build the RTD docs; after release we can change that to automatic builds out of `master`.


jreiberkyle (approved):
Gorgeous. Simple and yet does **so much**.

Issue #276, opened 2021-04-28
upgrade to httpx 0.18


jreiberkyle:
httpx 0.18 introduces some really great functionality (https://github.com/encode/httpx/blob/master/CHANGELOG.md) and also breaks stuff. To fix the breakage, we pinned to httpx 0.16. Check to see if exponential backoff retry is added and, if so, upgrade.

jreiberkyle:
Yes, httpx 0.18 adds retries (via [httpcore](https://github.com/encode/httpcore/blob/master/CHANGELOG.md#0121-november-7th-2020) so it makes sense to upgrade.

jreiberkyle:
Looking into the [retry code](https://github.com/encode/httpcore/blob/4a86afdc8cb616b4a477d26c6ddcbee7209670bf/httpcore/_async/connection.py#L129) in httpcore, retry is only triggered on connection error or timeout. So exponential backoff for TOO_MANY_REQUESTS status code still needs to be implemented in our codebase.

PR #280, opened 2021-04-30
V2: upgrade httpx to version 0.18.1

jreiberkyle:
Upgrade httpx to version 0.18.1

Closes # 276

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -25,7 +25,7 @@ with open('planet/api/__version__.py') as f:
 
 
 install_requires = [
-    'httpx==0.16',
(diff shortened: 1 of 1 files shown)


jreiberkyle:
Ahhhhh great find on #312! Can you actually have this fix moved to directly address #312? It seems like it will resolve that ticket but not #276.

Curious what issues you found with 0.19. Would you be willing to document them in #276? That could be really helpful when we do tackle the upgrade.

jreiberkyle:
closing this since  #276 is on the back burner for now

Issue #275, opened 2021-04-28
remove prefixes from code in rendered markdown


jreiberkyle:
Removing the `>>>` and `...` line prefixes from code snippets would make it easier copy from the rendered markdown. However, these prefixes are currently included to enable doctests. Consider other options for testing code snippets e.g.
- https://pypi.org/project/exdown/ (first-glance favorite)
- https://github.com/tmarktaylor/phmdoctest

Ref:  https://github.com/planetlabs/planet-client-python/pull/271#discussion_r617731380_

Also, rework doctests so that tests don't fail on auth.

cholmes:
closing in favor of #338

PR #282, opened 2021-05-12
V2: Add authentication handling

### LINKED ISSUE #248: abstract and improve authentication handling
Abstract authentication in v2 so that it does not rely on api keys and tuples. Provide functionality for obtaining and storing the api key from username/password with the python library. Also provide functionality for using a stored api key, api key provided as an argument, and username/password provided as an argument to perform authentication

current usage:
```python
API_KEY = os.getenv('PL_API_KEY')

async def create_order(order_detail):
    async with APlanetSession(auth=(API_KEY, '')) as ps:
        client = AOrdersClient(ps)
```

This requires obtaining the api key from elsewhere and is subject to mistakes in the ordering of the authentication tuple.

Suggested usage-

Stored api key - local file or environmental variable (for docker):
```python
import planet

auth = planet.get_auth()

async with APlanetSession(auth=auth) as ps:
    client = AOrdersClient(ps)
```
Provided api key:
```python
import planet

auth = planet.get_auth(api_key=<api_key>)

async with APlanetSession(auth=auth) as ps:
    client = AOrdersClient(ps)
```

Provided username/password:
```python
import planet

auth = planet.get_auth(api_key=<api_key>)

async with APlanetSession(auth=api_auth) as ps:
    client = AOrdersClient(ps)
```

Implementation in v1
https://github.com/planetlabs/planet-client-python/blob/5dc42d4f93b3c8826a1078c2dc063ad1ccc88d75/planet/scripts/cli.py#L101

jreiberkyle:
This PR introduces an `auth` module that handles obtaining, managing, and using authentication with the Planet services. `auth.Auth` is elevated to a first-level import. Additionally, `Session` provides default behavior for obtaining authentication information which can be overridden with an `Auth` instance. To support obtaining authentication information from the Planet Auth API, a new, synchronous HTTP session is introduced, `http.AuthSession`, and both `http.Session` and `http.AuthSession` inherit from a new base class `http.BaseSession`.

Places to check on changes in usability are the docs, `README.md` and `docs/guide.md` being the primary two.

(Resolution: documented that doctests fail and updated #275 to fix doctests) Right now, there is a code example for obtaining authentication using login in the "User Guide" documentation page that fails doctests. This is because the example username and password are, obviously, not going to lead to a successful authentication. I'm not sure how to skip this test at this time and may add this to or make it a cousin of #275, which is about how to enable doctests without prefixes and may lead to more refined control over which doctests are run. I would love any other suggestions. 

(Resolution: remove reading from environment variable by default, mock the read of the secret file and override locally for auth tests) Another issue I've found is that the local tests may pass but the remote tests may fail because the remote tests are in an environment where the default secret file and environment variable do not exist. I am interested in finding solutions to this.

Closes #248

### DIFF
--- a/planet/auth.py
+++ b/planet/auth.py
@@ -13,6 +13,7 @@
 # limitations under the License.
 '''Manage authentication with Planet APIs'''
 from __future__ import annotations  # https://stackoverflow.com/a/33533514
+import abc
 import json
 import logging
 import os
@@ -31,15 +32,13 @@ class AuthException(Exception):
     pass
 
 
-class Auth():
+class Auth(metaclass=abc.ABCMeta):
     '''Handle authentication information for use with Planet APIs.'''
     @staticmethod
     def read(
-        key: str = None,
-        environment_variable: str = None,
-        secret_file_path: str = None
+        key: str = None
     ) -> APIKeyAuth:
-        '''Read authentication information.
+        '''Reads authentication information.
 
         If key is provided, uses the key. Otherwise, tries to find key from
         environment variable `PL_API_KEY`. Finally, tries to find key from the
@@ -48,44 +47,26 @@ class Auth():
 
         Parameters:
             key: Planet API key
-            environment_variable: Alternate environment variable.
-            secret_file_path: Alternate path for the planet secret file.
         '''
         if key:
             auth = Auth.from_key(key)
         else:
             try:
-                environment_variable = environment_variable or ENV_API_KEY
-                auth = Auth.from_env(environment_variable)
-            except APIKeyAuthException:
+                auth = Auth.from_env(ENV_API_KEY)
+            except AuthException:
                 try:
-                    filename = secret_file_path or SECRET_FILE_PATH
-                    auth = Auth.from_file(filename)
+                    auth = Auth.from_file(SECRET_FILE_PATH)
                 except FileNotFoundError:
                     raise AuthException(
                         'Could not find authentication information. Set '
                         f'environment variable {ENV_API_KEY} or store '
-                        'information in secret file with `APIKeyAuth.write()`')
+                        'information in secret file with `Auth.write()`')
         return auth
-
-    @staticmethod
-    def write(
-        auth: Auth,
-        filename: str = None
-    ):
-        '''Write authentication information.
-
-        Parameters:
-            filename: Alternate path for the planet secret file.
-        '''
-        filename = filename or SECRET_FILE_PATH
-        secret_file = _SecretFile(filename)
-        secret_file.write(auth.to_dict())
 
     @staticmethod
     def from_key(
         key: str
-    ) -> APIKeyAuth:
+    ) -> Auth:
         '''Obtain authentication from api key.
 
         Parameters:
@@ -98,7 +79,7 @@ class Auth():
     @staticmethod
     def from_file(
         filename: str = None
-    ) -> APIKeyAuth:
+    ) -> Auth:
         '''Create authentication from secret file.
 
         The secret file is named `.planet.json` and is stored in the user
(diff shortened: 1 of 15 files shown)


planet/auth.py
```diff
@@ -0,0 +1,285 @@
...
+                }
+
+        sess = http.AuthSession()
+        req = models.Request(url, method='POST', json=data)
+        resp = sess.request(req)
+        auth_data = self.decode_response(resp)
+        return auth_data
+
+    @staticmethod
+    def decode_response(response):
```
strixcuriosus, line 200: A library like pyjwt could help with decoding: 
https://pyjwt.readthedocs.io/en/latest/

(It's also compatible with some of the upcoming admin system changes for token verification)

planet/auth.py
```diff
@@ -0,0 +1,285 @@
...
+        '''This is magic I don't understand'''
+        jwt = response.json()['token']
+
+        # stuff before the first '.' and after the second '.' doesn't matter
```
strixcuriosus, line 204: Those parts are useful for knowing the jwt algorithm and verifying the token signature

planet/auth.py
```diff
@@ -0,0 +1,285 @@
...
+        payload = jwt.split('.')[1]
+
+        # the '===' addition ensures adequate padding
```
strixcuriosus, line 207: I'm curious if this is valid for possible JWT changes in the new admin system.

planet/auth.py
```diff
@@ -0,0 +1,285 @@
...
+    DICT_KEY = 'key'
+
+    def __init__(
+        self,
+        key: str
+    ):
+        '''Initialize APIKeyAuth.
+
+        Parameters:
+            key: API key.
```
strixcuriosus, line 229: why not call this `api_key` rather than `key` to be more explicit?

planet/auth.py
```diff
@@ -0,0 +1,285 @@
...
+        url = self._base_url + 'login'
+        data = {'email': email,
+                'password': password
+                }
+
+        sess = http.AuthSession()
+        req = models.Request(url, method='POST', json=data)
+        resp = sess.request(req)
+        auth_data = self.decode_response(resp)
+        return auth_data
```
strixcuriosus, line 197: Eventually you may want to consider keeping the full JWT around since various services are moving toward using JWT auth headers over API key headers. Realistically it will likely be a while before everything supports JWT auth headers, though.

planet/auth.py
```diff
@@ -0,0 +1,285 @@
...
+import httpx
+
+from . import constants
+from .api import http, models
+
+AUTH_URL = constants.PLANET_BASE_URL + 'v0/auth/'
+
+ENV_API_KEY = 'PL_API_KEY'
+
+SECRET_FILE_PATH = os.path.join(os.path.expanduser('~'), '.planet.json')
```
sarasafavi, line 31: Just commenting to say this made me check the docs for `expanduser()` and TIL it works for Windows envs too. Nice!

sarasafavi (approved):

planet/auth.py
jreiberkyle: Yes! Thank you for this suggestion. I just removed a ton of confusing 'magic' code and replaced it with one simple function call.

planet/auth.py
jreiberkyle: right now we are returning what `jwt.decode()` returns, which is the body of the JWT. That body contains more than the api key tho that's all we are using atm.

planet/auth.py
jreiberkyle: thanks! That's right outta V1, so was also a learning exercise for me!

planet/auth.py
jreiberkyle: not sure but we don't need to worry about it since we are using `jwt` now. yay!

planet/auth.py
jreiberkyle: Mostly for succinctness. Since the class is called `APIKeyAuth`, it felt redundant to use `api_key`.

PR #284, opened 2021-06-09
v2: add cli for orders api

### LINKED ISSUE #244: implement cli for orders client
once the orders client python library is complete, implement the cli

### LINKED ISSUE #252: implement cli auth key retrieval
We have functionality for storing the API key via the CLI but no functionality for getting it. The primary use case for pulling the api key from the environmental variable is for use in a Docker container (as done in [planet notebooks](https://github.com/planetlabs/notebooks)). The api key is passed as a command-line argument when the Docker container is built. Add a CLI command for retrieving the api key so users and docs don't have to hard-code the path to the planet auth information.

So CLI commands for interacting with auth would be e.g. 
storing api key
`planet auth --init`

retrieving api key
 `planet auth --key` (explodes with a non-zero return if the planet key is not set, similar to `set +u` or `set +o unset`)

usage example:
`export PL_API_KEY=$(planet auth --key)`

_Originally posted by @jreiberkyle in https://github.com/planetlabs/planet-client-python/pull/240#discussion_r591825737_

jreiberkyle:
This PR adds a CLI as a thin wrapper around the python api orders client. Functionality was added to the orders client and other modules to support back-compatible CLI functionality while keeping the CLI thin.

One addition to the API is implemented:
Simple management of geojson, via the new `planet/geojson.py` module

Two changes to the python API are implemented:
1. `OrdersClient.create_order()` now returns the entire order description (as returned by the API) instead of just the id.
       This aligns with the rule of "make the client as similar to the REST API as possible" and allows the cli to easily report the details as a json blob that can be processed or viewed via the command line.
2. `Delivery.from_dict()` tries to create a subclass if possible.
        For example, if the dict describes an `AmazonS3Delivery`, it makes an instance of that class. This is necessary because a `Delivery` only holds the minimum delivery information while a subclass holds more information about the specific cloud. This change allows the cli to just call `Delivery.from_dict()` while parsing a file and have the proper information passed into `OrdersClient.create_order()`.

Other than that, the API changes are improved error handling.

Wish list: The cli.py module got huge quick. I would like to break it up into different modules based on api (e.g. `auth` and `orders`). Ticket #288 

Closes #244 
Closes #252

### DIFF
--- a/planet/geojson.py
+++ b/planet/geojson.py
@@ -31,7 +31,7 @@ class WrongTypeException(GeoJSONException):
     pass
 
 
-class Geometry():
+class Geometry(dict):
     '''Manage and validate GeoJSON geometry description.'''
     def __init__(
         self,
@@ -46,8 +46,9 @@ class Geometry():
         Parameters:
             data: GeoJSON geometry, Feature, or FeatureClass.
         """
-        self._data = self._geom_from_dict(data)
-        self._type = self._geom_type(self._data)
+        geom = self._geom_from_dict(data)
+        super().__init__(geom)
+        self._validate()
 
     @classmethod
     def _geom_from_dict(
@@ -83,17 +84,21 @@ class Geometry():
                 ret = cls._geom_from_dict(features[0])
         return ret
 
-    @staticmethod
-    def _geom_type(data: dict):
-        '''
-        Parameters:
-            data: GeoJSON geometry
+    def __eq__(self, other):
+        def _tuple_to_list(obj):
+            return json.loads(
+                    json.dumps(obj).replace(")", "]").replace("(", "["))
 
+        return _tuple_to_list(self) == _tuple_to_list(other)
+
+    def _validate(self):
+        '''
         Raises:
             GeoJSONException: If data is not valid GeoJSON geometry, Feature,
             or FeatureClass.
-            WrongTypeException: If geometry coordinates to not fit type.
+            WrongTypeException: If geometry coordinates do not fit type.
         '''
+        data = self
         if 'type' not in data:
             raise GeoJSONException(
                 'Missing \'type\' key.')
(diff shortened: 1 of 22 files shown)


planet/geojson.py
```diff
@@ -0,0 +1,123 @@
...
+            ret = as_geom(data["geometry"])
+        except KeyError:
+            try:
+                # featureclass
+                features = data['features']
+            except KeyError:
+                raise GeoJSONException('Invalid GeoJSON: {data}')
+
+            if len(features) > 1:
+                LOGGER.warning(
```
sarasafavi, line 92: Do you mean to raise DataLossWarning here?

planet/geojson.py
```diff
@@ -0,0 +1,123 @@
...
+
+
+class WrongTypeException(GeoJSONException):
+    '''wrong GeoJSON type'''
+    pass
+
+
+class DataLossWarning(UserWarning):
+    """Warn that data will be lost."""
+
```
sarasafavi, line 35: for consistency add the `pass` no-op here too

tests/unit/test_geojson.py
```diff
@@ -0,0 +1,104 @@
...
+    geom_geojson['coordinates'] = 'invalid'
+    with pytest.raises(geojson.GeoJSONException):
+        _ = geojson.validate_geom(geom_geojson)
+
+
+def test_validate_geom_empty_coordinates(geom_geojson):
+    geom_geojson['coordinates'] = []
+    _ = geojson.validate_geom(geom_geojson)
+
+
```
sarasafavi, line 93: let's test_validate_geom_multiple_features (or whatever) here too :)

sarasafavi (commented):
very minor fix request re: planet.geojson -- otherwise this LGTM! And it works on my box, too :-D

Really great work @jreiberkyle I am SO EXCITED to see this happening!

planet/geojson.py
jreiberkyle: thank you for bringing this up and for our offline discussion. I'll raise an exception when multiple features are given.

jreiberkyle:
thank you so much @sarasafavi for the in-depth review. changes made, merging now. YASSS!!

PR #290, opened 2021-06-22
V2: change order_details objects from classes to functions, add support for specific tools

jreiberkyle:
## Summary
1. Changes `order_details` objects from classes to functions that return dicts. This greatly decreases the amount of code required to achieve building up a request for creating an order.

2. Further, no longer imports the objects at the top level. This is to avoid naming conflicts with the new functions (this was less of a concern when the objects were classes)

e.g. naming conflict: avoid 
```python
from planet import product

product = product(['20170614_113217_3163208_RapidEye-5'], 'analytic', 'REOrthoTile') # uh oh!
```

less chance of a naming conflict: better?
```python
from planet import order_details

product = order_details.product(['20170614_113217_3163208_RapidEye-5'], 'analytic', 'REOrthoTile') # we are good!
```

3. Also adds support for all of the tools supported in the Orders API (e.g. `toar_tool()`, which creates the description for a TOAR tool from arguments)

## Usability impact:

Context - use of `order_details` in `OrdersClient`:
```python
>>> async def main():
...     async with Session() as sess:
...         cl = OrdersClient(sess)
...         order_id = await cl.create_order(order_detail)
...
>>> asyncio.run(main())

```

before - 
```python
>>> from planet import OrderDetails, Product, Tool
>>>
>>> image_ids = ['20170614_113217_3163208_RapidEye-5']
>>> order_detail = OrderDetails(
...     'test_order',
...     [Product(image_ids, 'analytic', 'REOrthoTile')],
...     [Tool('toar', {'scale_factor': 10000}),
...      Tool('reproject', {'projection': 'WGS84', kernel: 'cubic'}),
...      Tool('tile', {'origin_x: -180, 'origin_y: -90, 'pixel_size': 0.000027056277056,
...                         'name_template': 'C1232_30_30_{tilex:04d}_{tiley:04d}'})
... )
...
```

after - 
```python
>>> from planet.api.order_details import (
...     build_request, product, toar_tool, reproject_tool, tile_tool)
...
>>> products = [
...     product(['20170614_113217_3163208_RapidEye-5'], 'analytic', 'REOrthoTile')])
... ]
...
>>> tools = [
...     toar_tool(scale_factor=10000),
...     reproject_tool(projection='WSG84', kernel='cubic'),
...     tile_tool(1232, origin_x=-180, origin_y=-90,
...               pixel_size=0.000027056277056,
...               name_template='C1232_30_30_{tilex:04d}_{tiley:04d}')
... ]
...
>>> order_request = build_request(
[...]

PR #298, opened 2021-07-08
V2: reorganize file structure to accommodate multiple clients and python api as 1st class

### LINKED ISSUE #295: Reorganize structure, specifically remove `api` module and add `client` module
To prepare for the introduction of a new client, the data client, introduce a new module called `clients` that will contain all of the API clients as well as their common `_BaseClient` parent.

Additionally, remove the `api` module, as it is an artifact from v1 and doesn't make sense now that the api is 1st class and the CLI (which has its own module) is a thin wrapper.

Usability impact:
As all of these modules are exposed at the top level, the user experience will not change.

jreiberkyle:
NOTE: This assumes the successful merge of #290.

This closes #295 and all of the changes and impacts are described in that ticket.


sarasafavi (approved):
:+1:

Issue #289, opened 2021-06-16
implement python api support for the data api 


jreiberkyle:
Support the data api in the SDK python api.

jreiberkyle:
Ensure data download supports COGs

jreiberkyle:
Add a function that allows the user to run any calculation on the imagery via a callback and return the result, no download. 

Usage example would be calculating stats on a COG.

jreiberkyle:
This ticket has been broken into smaller tickets.

PR #302, opened 2021-07-22
V2 api error parsing

jreiberkyle:
Move error message parsing into orders client to reflect the fact that different APIs have different error messages.

Part of #289


sarasafavi (approved):
:+1:

PR #313, opened 2021-09-01
Planet orders download default to not overwrite files

kevinlacaille:
- Added option on whether or not to overwrite data in CLI
- Added overwriting logic in orders module
- Made the files download under a directory named after the order ID, consistent with our docs (https://developers.planet.com/docs/orders/delivery/)
- Updated downloading info echoed back to the user
- Updated an example to work with updated orders module

### DIFF
--- a/examples/orders_create_and_download_multiple_orders.py
+++ b/examples/orders_create_and_download_multiple_orders.py
@@ -82,9 +82,9 @@ async def create_and_download(order_detail, directory, client):
 
     # download
     print(f'Downloading {oid} to {directory}.')
-    filenames, NUM_DOWNLOADS, NUM_DOWNLOADS_SKIPPED = await client.download_order(oid, directory, progress_bar=True)
-    print(f'Downloaded {oid}: '
-          f'{NUM_DOWNLOADS} files downloaded to {directory}. {NUM_DOWNLOADS_SKIPPED} skipped downloads (not overwritten).')
+    filenames = await client.download_order(oid, directory, progress_bar=True)
+    for i in range(len(filenames)):
+        print(filenames[i])
 
 
 async def main():
--- a/examples/orders_create_and_download_multiple_orders.py
+++ b/examples/orders_create_and_download_multiple_orders.py
@@ -82,9 +82,9 @@ async def create_and_download(order_detail, directory, client):
 
     # download
     print(f'Downloading {oid} to {directory}.')
-    filenames = await client.download_order(oid, directory, progress_bar=True)
+    filenames, NUM_DOWNLOADS, NUM_DOWNLOADS_SKIPPED = await client.download_order(oid, directory, progress_bar=True)
     print(f'Downloaded {oid}: '
-          f'{len(filenames)} files downloaded to {directory}.')
+          f'{NUM_DOWNLOADS} files downloaded to {directory}. {NUM_DOWNLOADS_SKIPPED} skipped downloads (not overwritten).')
 
 
 async def main():
--- a/examples/orders_create_and_download_multiple_orders.py
+++ b/examples/orders_create_and_download_multiple_orders.py
@@ -83,8 +83,8 @@ async def create_and_download(order_detail, directory, client):
     # download
     print(f'Downloading {oid} to {directory}.')
     filenames = await client.download_order(oid, directory, progress_bar=True)
-    for i in range(len(filenames)):
-        print(filenames[i])
+    for f in filenames:
+        print(f)
 
 
 async def main():
--- a/planet/clients/orders.py
+++ b/planet/clients/orders.py
@@ -75,7 +75,9 @@ class OrdersClient():
                 base url.
         """
         self._session = session
-
+        self.NUM_DOWNLOADS = 0
+        self.NUM_DOWNLOADS_SKIPPED = 0
+        
         self._base_url = base_url or BASE_URL
         if not self._base_url.endswith('/'):
             self._base_url += '/'
@@ -277,6 +279,7 @@ class OrdersClient():
     async def download_asset(
         self,
         location: str,
+        order_id: str,
         filename: str = None,
         directory: str = None,
         overwrite: bool = False,
@@ -286,6 +289,7 @@ class OrdersClient():
 
         Parameters:
             location: Download location url including download token.
+            order_id: The ID of the order            
             filename: Custom name to assign to downloaded file.
             directory: Write to given directory instead of current directory.
             overwrite: Overwrite any existing files.
@@ -301,20 +305,26 @@ class OrdersClient():
 
         async with self._session.stream(req) as resp:
             body = StreamingBody(resp)
-            dl_path = os.path.join(directory or '.', filename or body.name)
-            is_dl_path = os.path.isfile(dl_path)
-            # File doesn't exist or overwrite requested OR file doesn't exist and do not overwrite
-            if (overwrite or not is_dl_path) or (not overwrite and not is_dl_path):
-                await body.write(dl_path,
-                                overwrite=overwrite,
-                                progress_bar=progress_bar)
-            # File exists and do not overwrite it
-            else:
-                # Log something -> didn't download image vecause already existed
-                # Maybe even write something on the terminal?
+            dl_path = os.path.join(directory or '.', order_id)
+            # If path to download data doesn't exist, create directory
+            # Consistent with https://developers.planet.com/docs/orders/delivery/
+            if not os.path.isdir(dl_path):
+                os.mkdir(dl_path)
+            file_path = os.path.join(dl_path, filename or body.name)
+            is_file_path = os.path.isfile(file_path)
+
+            # Overwrite data or file doesn't exist in requested path
+            if overwrite or not is_file_path:
+                await body.write(file_path,
+                                 overwrite=overwrite,
+                                 progress_bar=progress_bar)
+                self.NUM_DOWNLOADS += 1
+            # Do not overwrite and file exists
+            elif is_file_path and not overwrite:
+                self.NUM_DOWNLOADS_SKIPPED += 1
                 pass
             
-        return dl_path
+        return file_path
 
     async def download_order(
         self,
@@ -333,6 +343,8 @@ class OrdersClient():
 
         Returns:
             Paths to downloaded files.
+            Number of downloaded files.
+            Number of skipped downloads.
 
         Raises:
             planet.exceptions.APIException: On API error.
@@ -343,11 +355,12 @@ class OrdersClient():
             f'downloading {len(locations)} assets from order {order_id}'
         )
         filenames = [await self.download_asset(location,
+                                               order_id,
                                                directory=directory,
                                                overwrite=overwrite,
                                                progress_bar=progress_bar)
                      for location in locations]
-        return filenames
+        return filenames, self.NUM_DOWNLOADS, self.NUM_DOWNLOADS_SKIPPED
 
     async def poll(
         self,
--- a/planet/clients/orders.py
+++ b/planet/clients/orders.py
@@ -75,8 +75,6 @@ class OrdersClient():
                 base url.
         """
         self._session = session
-        self.NUM_DOWNLOADS = 0
-        self.NUM_DOWNLOADS_SKIPPED = 0
         
         self._base_url = base_url or BASE_URL
         if not self._base_url.endswith('/'):
@@ -279,7 +277,6 @@ class OrdersClient():
     async def download_asset(
         self,
         location: str,
-        order_id: str,
         filename: str = None,
         directory: str = None,
         overwrite: bool = False,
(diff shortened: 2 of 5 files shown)


planet/api/orders.py
```diff
@@ -30,6 +30,7 @@
 ORDERS_PATH = 'orders/v2/'
 BULK_PATH = 'bulk/orders/v2/'
 
+# Order states: https://developers.planet.com/docs/orders/ordering/#order-states
```
jreiberkyle, line 33: so helpful, great add!

jreiberkyle (changes requested):
High level comments before I dive into review:

"-Added overwriting logic in orders module" - overwriting was happening by default already. What additional logic was needed?

"-Made the files download under a directory named after the order ID, consistent with our docs" - I really like the approach of looking at ways we can sync how the SDK works and how the API works. Forcing the user to download to a subdirectory does not seem like what we want, as that may not be what they want. Making it easy to download to a subdirectory named after the order_id does sound good. I suggest we break this additional feature off from this ticket, which is focused on overwriting, and have a discussion on how we want to implement it.

"-Updated an example to work with updated orders module" - thank you so much for this, very thorough.

planet/scripts/cli.py
```diff
@@ -200,23 +200,49 @@ async def cancel(ctx, order_id):
...
-        filenames = await cl.download_order(
-                str(order_id),
-                directory=dest,
-                overwrite=overwrite,
-                progress_bar=not quiet)
-    click.echo(f'Downloaded {len(filenames)} files.')
 
+    if overwrite:
+        # Explicity ask the user if they want to overwrite their data
+        overwrite_decision = click.prompt(f'You have selected to overwrite your data. Proceed? [y/n]', type=str).lower()
```
jreiberkyle, line 215: Adding the overwrite flag seems like assurance enough that the user wants to overwrite the files. Prompts like this are more work for anyone wanting to use scripting so I tend to avoid them unless absolutely necessary.

planet/api/orders.py
```diff
@@ -74,7 +75,9 @@ def __init__(
                 base url.
         """
         self._session = session
-
+        self.NUM_DOWNLOADS = 0
```
jreiberkyle, line 78: two thoughts here:
1. I suggest using lowercase for variables like this since they are not global vars
2. I'm curious about tracking downloads/skips at this level. What happens if we use the client to download multiple orders? What if the client is downloading multiple orders async? Multiple sources would be changing and accessing these values. Seems like something we want just returned by the functions themselves to avoid confusion.

examples/orders_create_and_download_multiple_orders.py
```diff
@@ -82,9 +82,9 @@ async def create_and_download(order_detail, directory, client):
 
     # download
     print(f'Downloading {oid} to {directory}.')
-    filenames = await client.download_order(oid, directory, progress_bar=True)
+    filenames, NUM_DOWNLOADS, NUM_DOWNLOADS_SKIPPED = await client.download_order(oid, directory, progress_bar=True)
```
jreiberkyle, line 85: I suggest using lower case as these are local variables and upper case suggests global vars

planet/api/orders.py
```diff
@@ -247,15 +250,17 @@ async def aggregated_order_stats(self) -> dict:
     async def download_asset(
         self,
         location: str,
+        order_id: str,
```
jreiberkyle, line 253: I would prefer the user not need to know the order id if they already know the location of the asset they want to download. That just makes more work for them.

jreiberkyle (changes requested):
Sorry about the dual reviews here. I'm still learning LOL!

kevinlacaille:
> "-Added overwriting logic in orders module" - overwriting was happening by default already. What additional logic was needed?

This is mostly poor wording on my part, but I was referring to the "logic" added in lines 279-298 in `orders.py`, where it:
1) chooses whether or not to create a data directory; 
2) then either does or doesn't write the data.

and 213-245 in `cli.py`, where it:
1) explicitly asks the user if they want to overwrite their data;
2) based on this decision, it either does or doesn't download the data;
3) then echos resulting proper information.

> "-Made the files download under a directory named after the order ID, consistent with our docs" - I really like the approach of looking at ways we can sync how the SDK works and how the API works. Forcing the user to download to a subdirectory does not seem like what we want, as that may not be what they want. Making it easy to download to a subdirectory named after the order_id does sound good. I suggest we break this additional feature off from this ticket, which is focused on overwriting, and have a discussion on how we want to implement it.

This is something @sarasafavi and I were discussing last week. We were talking about a scenario where a user already has an order in this directory or if they wanted to download multiple orders worth of data stored side-by-side. With how the Orders API was setup, it would have an issue with multiple `manifest.json` files in the same directory. 

**For example:** If a user were to request to download 3 orders in the same directory, the first order would download without any issue, however on the second order the Orders API would see that a `manifest.json` file that already exists in the chosen directory, and would then stop the download for its `manifest.json` and ask the user to select `-o / --overwrite` to overwrite this order. It would then go on to download the 3rd order and do the same thing. So, the user would be left with 3 orders worth of imagery, but only the first order's `manifest.json` file. 

After speaking with the D&D team, we decided to be consistent with [our docs]((https://developers.planet.com/docs/orders/delivery/)), we would proceed with the following convention :
`{OrderID}/manifest.json`
`{OrderID}/imagery.zip` 

> "-Updated an example to work with updated orders module" - thank you so much for this, very thorough.

☺️

planet/scripts/cli.py
kevinlacaille: I'm not against removing this prompt, I mostly added it as a backup plan, if the user didn't mean to select to overwrite their data. My only concern was that the user may have hit the up arrow a few times in their terminal and accidentally selected to overwrite all the data they may have spent an hour downloading. We could keep it and add a `-f` / `--force` flag, which would skip this prompt.

Thoughts?

jreiberkyle: We already have a flag they have to add to overwrite. Adding a force flag is just requiring two flags for one action (and, honestly, is subject to the same user error). I have trust in the user that they want what they say they want. I suggest getting rid of the prompt all together.

jreiberkyle:
It seems the manifest issue is adequately interesting and novel as to require a different ticket for work. This ticket is focused on overwriting by default, which seems to have agreement and can be closed pretty easily. Can we continue the download directory / rename discussion in a new ticket?

planet/scripts/cli.py
kevinlacaille: Cool - agreed!

examples/orders_create_and_download_multiple_orders.py
kevinlacaille: Makes sense, done!

planet/api/orders.py
kevinlacaille: I'm not sure I fully understand how users interact with this module on the Python side yet, but from the looks of it, I thought users only queried `self.download_order()` to download their order, for which they need to provide an Order ID. I suppose the Order ID could get stored in the class object, `self`, and get updated every time they request to download a new order. I don't think my current changes affect the user on the CLI side of things.

Thoughts?

planet/api/orders.py
kevinlacaille: 1. Done!
2. Good catch - thank you! You’re right, I never reset `NUM_DOWNLOADS` and `NUM_DOWNLOADS_SKIPPED` both back to 0. I think you’re right, it would make more sense to have `download_asset()` return these two values, instead of storing them in the class object, `self`.

kevinlacaille:
Ok, so:
- Removed all of the directory creation work, as it'll be addressed in the next PR, associated with issue #322 
- Fixed all of the testing and linting issues we were having
- Adjusted the docs, accordingly. 

Let me know if there's anything else to be done! :)

jreiberkyle (changes requested):
Works as expected when run from the CLI and passes tests.

One thing: I’m questioning whether it makes sense to report all of the files in the order when none of them are downloaded because the overwrite flag is false. I don’t know, kind of seems counter intuitive. Like, I’m downloading files, I expect the reported files to be the ones that are downloaded.

planet/clients/orders.py
```diff
@@ -278,7 +279,7 @@ async def download_asset(
         location: str,
         filename: str = None,
         directory: str = None,
-        overwrite: bool = True,
+        overwrite: bool = False,
```
jreiberkyle, line 282: It seems there should be a test that would change with the change of the default value of overwrite to False. For example: a. test that runs this function with the default value in two cases: 1. the file already exists 2. the file doesn't already exist. b. same as a but the overwrite parameter is set to True.

planet/clients/orders.py
```diff
@@ -332,11 +334,13 @@ async def download_order(
         LOGGER.info(
             f'downloading {len(locations)} assets from order {order_id}'
         )
+
```
jreiberkyle, line 337: I'm still confused by all these extra lines 😆  This line, line 308 above and line 343 below.

jreiberkyle (commented):
This looks great and CLI runs like a dream. It is **almost** there!!

Two items I'd love to discuss before merging:
1. I would expect any change in the code to somehow trigger a change to the tests, which I'm not seeing here.
2. This is more of a bonus, but is there any change to the documentation that can accompany this change?

tests/integration/test_orders_api.py
```diff
@@ -583,3 +582,123 @@ async def test_download_order(tmpdir, order_description, oid, session):
...
+        tmpdir, order_description, oid, session):
+    '''
+    Test if download_order() overwrites pre-existing data with
+    overwrite flag set to True.
+    '''
+
+    # Create a JSON
+    original_content = {'key': 'original_file'}
+    # Save JSON to tmpdir
+    f = open(Path(tmpdir, 'file.json'), "a")
```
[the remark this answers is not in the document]
kevinlacaille: Oh yeah, I like this! Thanks :)

tests/integration/test_orders_api.py
```diff
@@ -583,3 +582,123 @@ async def test_download_order(tmpdir, order_description, oid, session):
...
+    f.write(json.dumps(original_content))
```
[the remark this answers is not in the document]
kevinlacaille: I didn't realize that `json.dump()` was faster than `json.dumps()`. Changed all `json.dump()`s to `json.dumps()`s.

tests/integration/test_orders_api.py
```diff
@@ -583,3 +582,123 @@ async def test_download_order(tmpdir, order_description, oid, session):
...
+    f.close()
+
+    downloaded_content = {'key': 'downloaded_file'}
+    cl = await get_orders_client(downloaded_content, order_description, oid,
+                                 session)
+    # Download order and overwrite data
+    _ = await cl.download_order(oid, directory=str(tmpdir), overwrite=True)
+
+    assert json.loads(open(Path(tmpdir, 'file.json')).read()) != ...
```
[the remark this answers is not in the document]
kevinlacaille: I didn't realize that json.dump() was faster than json.dumps(). Changed all json.dump()s to json.dumps()s.

tests/integration/test_orders_api.py
```diff
@@ -583,3 +582,123 @@ async def test_download_order(tmpdir, order_description, oid, session):
...
+    cl = await get_orders_client(downloaded_content, order_description, oid,
+                                 session)
+    # Download order and overwrite data
+    _ = await cl.download_order(oid, directory=str(tmpdir), overwrite=False)
+
+    assert json.loads(open(Path(tmpdir, 'file.json')).read()) != ...
+    downloaded_content
+
+
+async def get_orders_client(
```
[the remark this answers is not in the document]
kevinlacaille: This is a great idea! After our call I moved this into a pytest fixture, but removed the Orders client from the function. Now this function only creates the mock response for downloading the JSON file and is now called `create_download_mock()`. I've also put the downloaded content (`{'key': 'downloaded_file'}`) and the original content (`{'key': 'original_file'}`) as pytest fixtures, as all 4 of my tests call for them and the pytest fixture `create_download_mock()` calls for the downloaded content.

tests/integration/test_orders_api.py
```diff
@@ -583,3 +582,123 @@ async def test_download_order(tmpdir, order_description, oid, session):
...
+    f.close()
+
+    downloaded_content = {'key': 'downloaded_file'}
+    cl = await get_orders_client(downloaded_content, order_description, oid,
+                                 session)
+    # Download order and overwrite data
+    _ = await cl.download_order(oid, directory=str(tmpdir), overwrite=True)
+
+    assert json.loads(open(Path(tmpdir, 'file.json')).read()) != ...
+    downloaded_content
```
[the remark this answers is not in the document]
kevinlacaille: Whoops, this was backwards, not sure why it passed, but I've now fixed this - thanks :)

tests/integration/test_orders_api.py
```diff
@@ -583,3 +582,123 @@ async def test_download_order(tmpdir, order_description, oid, session):
...
+    f.write(json.dumps(original_content))
+    f.close()
+
+    downloaded_content = {'key': 'downloaded_file'}
+    cl = await get_orders_client(downloaded_content, order_description, oid,
+                                 session)
+    # Download order and overwrite data
+    _ = await cl.download_order(oid, directory=str(tmpdir), overwrite=False)
+
+    assert json.loads(open(Path(tmpdir, 'file.json')).read()) != ...
```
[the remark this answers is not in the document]
kevinlacaille: You're totally right, checking for `contents != downloaded_content` does not seem intuitive. I fixed this. And like the previous comment, I fixed this error.

tests/integration/test_orders_api.py
```diff
@@ -583,3 +582,123 @@ async def test_download_order(tmpdir, order_description, oid, session):
...
+    set to false without pre-existing data.
+    '''
+
+    downloaded_content = {'key': 'downloaded_file'}
+    cl = await get_orders_client(downloaded_content, order_description, oid,
+                                 session)
+    # Download order and overwrite data
+    _ = await cl.download_order(oid, directory=str(tmpdir), overwrite=False)
+
+    assert json.loads(open(Path(tmpdir, 'file.json')).read()) != ...
```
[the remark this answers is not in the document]
kevinlacaille: Like above, fixed!

tests/integration/test_orders_api.py
```diff
@@ -583,3 +582,123 @@ async def test_download_order(tmpdir, order_description, oid, session):
...
+    set to true without pre-existing data.
+    '''
+
+    downloaded_content = {'key': 'downloaded_file'}
+    cl = await get_orders_client(downloaded_content, order_description, oid,
+                                 session)
+    # Download order and overwrite data
+    _ = await cl.download_order(oid, directory=str(tmpdir), overwrite=True)
+
+    assert json.loads(open(Path(tmpdir, 'file.json')).read()) != ...
```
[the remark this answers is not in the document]
kevinlacaille: Like above, fixed!

kevinlacaille:
- Abstracted away a repetitive HTTP mocked response for 4 tests into a pytest fixture, called `create_download_mock()`
- Put 2 dictionaries, `downloaded_contents` and `original_contents`, later used as JSON contents, in pytest fixtures for 2 reasons:
    1. `downloaded_contents` is used in the pytest fixture `create_download_mock()`
    2. Either both or one of `downloaded_contents` and/or `original_contents` are used in my 4 tests. I wanted to ensure that they are all being called from one place instead of being rewritten up to 4 times, which would give a higher chance of breaking the code in the future.
- Added plenty of useful comments to the code.

jreiberkyle (approved):
LGTM!

PR #318, opened 2021-09-09
Change polling output to a progress bar

### LINKED ISSUE #310: change polling output to a progress bar
Polling can fill up the stdout with all its status messages. Change the status to an updating progress bar. Move status messages to logging.

jreiberkyle:
This PR makes the following changes:
- Introduces a new module, `reporting`, which implements the status progress bar
- Changes spec of `OrdersClient.poll`: removes `verbose` (bool) and adds `report` (callback fcn)
- Fixes a bug in `create_order` where it fails on parsing an error message when there is no field
- Docs:
  - Adds `reporting` to API reference
  - Adds section in the user guide on downloading and polling an order with an example
  - Removes leading `>>> ` and `... ` symbols from user guide code examples to allow copy/paste

Closes #310

Usage:
```python
import asyncio
from planet import reporting, Session, OrdersClient

async def create_poll_and_download():
    async with Session() as sess:
        cl = OrdersClient(sess)
        with reporting.StateBar(state='creating') as bar:
            # create order
            order = await cl.create_order(request)
            bar.update(state='created', order_id=order.id)

            # poll
            await cl.poll(order.id, report=bar.update)

        # download
        await cl.download_order(order.id, progress_bar=True)

asyncio.run(create_poll_and_download())
```

REF: Example `request` for testing:
```python
request = {
  "name": "psorthotile_analytic",
  "products": [
    {
      "item_ids": [
        "4500474_2133707_2021-05-20_2419"
      ],
      "item_type": "PSOrthoTile",
      "product_bundle": "ANALYTIC"
    }
  ],
  "delivery": {
    "single_archive": true,
    "archive_type": "zip",
    "archive_filename": "string"
  },
  "tools": [
    {"file_format": {"format": "COG"}},
    {"toar": {"scale_factor": 10000}}
  ]
}
```

### DIFF
--- a/docs/guide.md
+++ b/docs/guide.md
@@ -8,29 +8,27 @@ The recommended way to use a `Session` is as a context manager. This will
 provide for automatic clean up of connections when the context is left.
 
 ```python
->>> import asyncio
->>> import os
->>> from planet import Session
->>>
->>> async def main():
-...     async with Session() as sess:
-...         # perform operations here
-...         pass
-...
->>> asyncio.run(main())
+import asyncio
+import os
+from planet import Session
 
+async def main():
+    async with Session() as sess:
+        # perform operations here
+        pass
+
+asyncio.run(main())
 ```
 
 Alternatively, use `await Session.aclose()` to close a `Session` explicitly:
 
 ```python
->>> async def main():
-...     sess = Session()
-...     # perform operations here
-...     await sess.aclose()
-...
->>> asyncio.run(main())
+async def main():
+    sess = Session()
+    # perform operations here
+    await sess.aclose()
 
+asyncio.run(main())
 ```
 
 ### Authentication
@@ -53,13 +51,12 @@ It can also be accessed, e.g. to store in an environment variable, as
 For example, to obtain and store authentication information:
 
 ```python
->>> import getpass
->>> from planet import Auth
->>>
->>> pw = getpass.getpass()
->>> auth = Auth.from_login('user', 'pw')
->>> auth.write()
+import getpass
+from planet import Auth
 
+pw = getpass.getpass()
+auth = Auth.from_login('user', 'pw')
+auth.write()
 ```
 
 When a `Session` is created, by default, authentication is read from the secret
@@ -74,18 +71,17 @@ respective functions. For example, authentication can be read from a custom
 environment variable:
 
 ```python
->>> import asyncio
->>> import os
->>> from planet import Auth, Session
->>>
->>> auth = Auth.from_env('ALTERNATE_VAR')
->>> async def main():
-...     async with Session(auth=auth) as sess:
-...         # perform operations here
-...         pass
-...
->>> asyncio.run(main())
+import asyncio
+import os
+from planet import Auth, Session
 
+auth = Auth.from_env('ALTERNATE_VAR')
+async def main():
+    async with Session(auth=auth) as sess:
+        # perform operations here
+        pass
+
+asyncio.run(main())
 ```
 
 ### Orders Client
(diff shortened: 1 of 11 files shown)


kevinlacaille (commented):
On `line 212` in `scripts/cli.py`, the keyword argument "`verbose`" is expected. I see that this was purposefully removed, as the callback function `report` was added, however `verbose` was not removed in the CLI script. When this keyword is removed as a keyword argument, the CLI seems to act as expected.

planet/clients/orders.py
```diff
@@ -344,27 +347,37 @@ async def poll(
         order_id: str,
         state: str = None,
         wait: int = 10,
-        verbose: bool = False
```
kevinlacaille, line 347: Did you mean to get rid of this variable? The CLI current expects `verbose` as an input.

docs/guide.md
```diff
@@ -53,13 +51,12 @@ It can also be accessed, e.g. to store in an environment variable, as
...
->>> auth.write()
+import getpass
+from planet import Auth
 
+pw = getpass.getpass()
+auth = Auth.from_login('user', 'pw')
+auth.write()
 ```
 
 When a `Session` is created, by default, authentication is read from the secret
```
kevinlacaille, line 62: On line 64, "explicitely" should be "explicitly".

planet/clients/orders.py
```diff
@@ -344,27 +347,37 @@ async def poll(
         order_id: str,
         state: str = None,
         wait: int = 10,
```
kevinlacaille, line 349: The user-facing information for the polling updates every 10 seconds. This feels somewhat clunky/slow from a UX point of view. By changing this to 1 second it feels much more smooth and somewhat more reliable, as I feel less anticipation waiting to see what's going to happen next. Is there a reason as to why the user must wait so long between updates? Perhaps not much would happen on such a short timescale (i.e., `running` > `success`), but from a UX point of view, seeing the clock tick every second feels better, or at least it does for me :)

planet/clients/orders.py
kevinlacaille, line 347: On line 212 in `scripts/cli.py` in the function `download()`, the keyword argument "`verbose`" is given as: `cl.poll(str(order_id), verbose=True)`.

I see that `verbose` was purposefully removed and replaced with `report`, however the keyword `verbose` was not removed in the CLI script. When this keyword is removed as argument (i.e., `cl.poll(str(order_id))`), the CLI seems to act as expected.

jreiberkyle: What a great catch! Thank you for being this thorough. I wish we had something in place to do integration tests involving the CLI. Ideally, they would have caught that. I have made a ticket for considering CLI integration tests as well (#327).

jreiberkyle: This was actually fairly in-depth. I had to add in the ability to disable the StatusBar reporter, to honor the `quiet` flag. Fixed now!

jreiberkyle:
Major updates:
1. The models.py issue is ticketed at #330 
2. Poll wait time is now 1s
3. `examples/orders_create_and_download_multiple_orders.py` was updated to use the new poll reporting
4. Added the ability to disable (aka silence) `StatusBar`, which we use in the CLI when the `--quiet` flag is true

kevinlacaille (approved):
This is a very comprehensive PR, great job, Jen! 🎉

I think we can go ahead and merge these commits to `v2`!

docs/guide.md
```diff
@@ -8,27 +8,29 @@ The recommended way to use a `Session` is as a context manager. This will
 provide for automatic clean up of connections when the context is left.
 
 ```python
-import asyncio
-import os
-from planet import Session
+>>> import asyncio
+>>> import os
+>>> from planet import Session
+>>>
```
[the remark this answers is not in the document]
jreiberkyle: great capture, thanks!

PR #334, opened 2021-11-02
added ortho_analytic_3b to default asset list in item_asset_types.py

mkshah605:
Added ortho_analytic_3b to default asset list


sarasafavi:
Thank you @mkshah605 ! Once this new asset is available in the API we can merge & release this.

PR #335, opened 2021-11-04
deleted last_modified function & datetime package in models.py 

mkshah605:
Deleted both the last_modified function and the datetime package in the models.py file, as neither of them are being used. This change successfully passed nox testing.


planet/models.py
```diff
@@ -157,16 +156,6 @@ def num_bytes_downloaded(self):
         '''
         return self.response.num_bytes_downloaded
 
-    def last_modified(self):
```
kevinlacaille, line 160: I believe I was wrong. I think we should keep this function. Although this planet module doesn't call for this function, `last_modified()`, a user might actually want to use this function. So, I suggest putting it back in along with the `datetime` module.

Now, my first comment still stands - we need to either change `datetime.strptime()` to `datetime.datetime.strptime()`. Or you can do `from datetime import datetime`, then `datetime.strptime()` will work :)

planet/models.py
```diff
@@ -13,7 +13,6 @@
 # limitations under the License.
 """Manage data for requests and responses."""
 import copy
-import datetime
```
kevinlacaille, line 16: From my comment bellow, this should go back in either as is or as `from datetime import datetime` - chef's choice!

kevinlacaille (commented):
I was wrong, I think we should keep this function in! See my comments for my reasoning 👍 

Now, I think we should add a test for this function in `tests/unit/test_models.py`!

planet/models.py
```diff
@@ -151,15 +142,13 @@ def size(self):
...
         '''The number of bytes downloaded.
-
         :returns: number of bytes downloaded
         :rtype: int
         '''
         return self.response.num_bytes_downloaded
 
     def last_modified(self):
         '''Read the last-modified header as a datetime, if present.
-
```
jreiberkyle, line 162: Is there a reason these empty lines were removed? Their presence increases readability of the docstrings. I suggest undoing all the 'remove empty line' changes

jreiberkyle (changes requested):
The new tests are focused and clear and cover the expected variations. They are great. I suggest removing all of the unrelated changes, ie the removal of empty lines.

kevinlacaille (approved):
Looks good to me! 🎉

jreiberkyle (approved):
Looks great. Nicely done!

Issue #321, opened 2021-09-17
Add zsh command for pip install in CONTRIBUTING.md


kevinlacaille:
Under the [Development section in the CONTRIBUTING.md](https://github.com/planetlabs/planet-client-python/blob/v2/CONTRIBUTING.md#development), we should denote that this `pip` command is specifically for `bash` users. For those who use `zsh`, the command is as follows:

```zsh
pip install -e .'[dev]'
```

kevinlacaille:
Closed by #340

kevinlacaille:
This issue now occurs in the section titled Alternative to nox in the CONTRIBUTING.md file (https://github.com/planetlabs/planet-client-python/blob/main/CONTRIBUTING.md#alternative-to-nox)

Solution is still the same.

For zsh, the command needs to be ```pip install -e .'[dev]'```

kevinlacaille:
Thank you for getting this merged, @JuLeeAtPlanet !

PR #340, opened 2021-11-19
Added compatible pip install -e command for Zsh.

kevinlacaille:
As title implies. I've added the compatible pip command for Zsh in the CONTRIBUTING doc.

Closing #321


sarasafavi:
woo, zsh!

PR #344, opened 2021-12-02
Added to .gitignore

kevinlacaille:
**Added:**
- `.coverage.*` for coverage reports, which are created in addition to `coverage.xml` on OS X
- `*.tif*` for all .tif and .tiff images.
- `.vscode/` for those who use the editor [VS Code](https://code.visualstudio.com/)


mkshah605 (approved):
looks good!

PR #350, opened 2021-12-16
Handle mix of single target items and source lists when printing the listing of mosaics in feeds

sgillies:
The "target" type appears to have a single dict, but "source" type seems to give a list of dicts. API change or something else? I'm new to the project :smile:

### DIFF
--- a/planet/scripts/v1.py
+++ b/planet/scripts/v1.py
@@ -405,15 +405,9 @@ def list_feeds(pretty, limit, stats):
     echo_json_response(response, pretty, limit)
 
 
-@feeds.command('list-mosaics')
-@click.argument('feed_id')
-def get_mosaic_list_for_feed(feed_id):
-    """List mosaics linked to feed."""
-    cl = clientv1()
-    feed_info = cl.get_feed_info(feed_id).get()
-
+def print_feed_mosaic_list(feed_info, client):
+    """Print the listing of mosaics for a feed."""
     for type_ in ["target", "source"]:
-
         feed_image_conf = feed_info.get(type_, None)
         if not feed_image_conf:
             continue
@@ -429,9 +423,18 @@ def get_mosaic_list_for_feed(feed_id):
 
             click.echo("{} mosaics:".format(type_))
             mosaic_series = conf_item["config"]["series_id"]
-            mosaics = cl.get_mosaics_for_series(mosaic_series)
+            mosaics = client.get_mosaics_for_series(mosaic_series)
             for mosaic in mosaics.get()["mosaics"]:
                 click.echo("\t{}".format(mosaic["name"]))
+
+
+@feeds.command('list-mosaics')
+@click.argument('feed_id')
+def get_mosaic_list_for_feed(feed_id):
+    """List mosaics linked to feed."""
+    cl = clientv1()
+    feed_info = cl.get_feed_info(feed_id).get()
+    print_feed_mosaic_list(feed_info, cl)
 
 
 @feeds.command('describe')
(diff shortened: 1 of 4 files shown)


planet/scripts/v1.py
```diff
@@ -415,10 +415,21 @@ def get_mosaic_list_for_feed(feed_id):
...
+        try:
+            if feed_image_conf['type'] != 'mosaic':
+                # Unsure about the following lines. Why create an
+                # exception if we don't raise it?
+                msg_format = 'The {} for this feed is not a mosaic type.'
+                click.ClickException(msg_format.format(type_))
+                continue
+        except:
+            import pdb; pdb.set_trace()
+            raise
```
sgillies, line 432: Debugging, will remove.

planet/scripts/v1.py
```diff
@@ -415,10 +415,21 @@ def get_mosaic_list_for_feed(feed_id):
     for type_ in ['target', 'source']:
         feed_image_conf = feed_info.get(type_)
 
-        if feed_image_conf['type'] != 'mosaic':
-            msg_format = 'The {} for this feed is not a mosaic type.'
-            click.ClickException(msg_format.format(type_))
-            continue
```
sgillies, line 421: This code has the looks of a accidentally-committed partial solution?

sgillies: Here's the commit https://github.com/planetlabs/planet-client-python/commit/ef726a53e32c12a4c8ad8885063d764e0c3044c5.

planet/scripts/v1.py
```diff
@@ -408,25 +408,28 @@ def list_feeds(pretty, limit, stats):
...
 def get_mosaic_list_for_feed(feed_id):
-    '''List mosaics linked to feed'''
+    """List mosaics linked to feed."""
     cl = clientv1()
     feed_info = cl.get_feed_info(feed_id).get()
 
-    for type_ in ['target', 'source']:
-        feed_image_conf = feed_info.get(type_)
+    for type_ in ["target", "source"]:
+        click.echo("{} mosaics:".format(type_))
```
sgillies, line 416: I want to point out that injecting headers into a program's output makes the output less consumable by other programs. For future versions we might want to print lines like

```json
{"type": "source", "name": "yadayada"}
{"type": "target", "name": "badabing"}
```

so that we can filter with jq or whatever.

cc @jreiberkyle

planet/scripts/v1.py
```diff
@@ -470,6 +475,8 @@ def get_mosaic_list_for_subscription(subscription_id):
         feed_image_conf = feed_info.get(type_)
 
         if feed_image_conf['type'] != 'mosaic':
+            # Unsure about the following lines. The exception is not
+            # raised.
             msg_format = 'The {} for this feed is not a mosaic type.'
             click.ClickException(msg_format.format(type_))
             continue
```
sgillies, line 482: I guess I should take care of this, too. Nothing reported, but there is clearly a bug in here.

planet/scripts/v1.py
```diff
@@ -405,28 +405,37 @@ def list_feeds(pretty, limit, stats):
...
 
 
-@feeds.command('list-mosaics')
-@click.argument('feed_id')
-def get_mosaic_list_for_feed(feed_id):
-    '''List mosaics linked to feed'''
-    cl = clientv1()
-    feed_info = cl.get_feed_info(feed_id).get()
+def print_feed_mosaic_list(feed_info, client):
+    """Print the listing of mosaics for a feed."""
```
sgillies, line 409: I factored this function out of three functions in the module. "Rule of 3" argues for this factor, plus I'm also fixing 3 * 2 bugs.

tests/fixtures/feeds.json
```diff
@@ -24,12 +24,12 @@
...
         "namespace": "{{ .Namespace }}"
       },
-      "source": {
+      "source": [{
         "config": {
           "series_id": "mosaic_series_id"
         },
         "type": "mosaic"
-      },
+      }],
```
sgillies, line 32: A feed.source is a list, see https://developers.planet.com/docs/analytics/reference/#tag/Feeds/paths/~1feeds~1{id}/get. By itself, the change to this file is enough to reproduce the error reported (offline).

planet/scripts/v1.py
jreiberkyle: whats the rule of 3?

sgillies: https://wiki.c2.com/?ThreeStrikesAndYouRefactor

sgillies: I like this take, too https://andrewbrookins.com/technology/the-rule-of-three/

> The principle behind the rule is that it’s easier to make a good abstraction from duplicated code than it is to refactor the wrong abstraction. That’s my take, anyway.

PR #351, opened 2022-01-04
Add missing asset type, clean up tests

sgillies:
This fixes the last failing test and improves the test that was failing. I'm going to make some comments inline that point out some defects in the original tests and suggest better patterns that we should use in future work.

cc @jreiberkyle @kevinlacaille @JuLeeAtPlanet @karrabatcheller @mkshah605 @sarasafavi

### DIFF
--- a/planet/scripts/item_asset_types.py
+++ b/planet/scripts/item_asset_types.py
@@ -78,7 +78,7 @@ DEFAULT_ASSET_TYPES = [
     'ps4b_basic_analytic_nitf', 'ps4b_basic_analytic_rpc_nitf',
     'ps4b_basic_analytic_xml_nitf', 'ps4b_udm', 'udm', 'ortho_udm2',
     'video_file', 'video_frames', 'video_metadata', 'ortho_visual',
-    'visual', 'visual_xml'
+    'analytic', 'visual', 'visual_xml'
 ]
 
 
--- a/tests/test_types.py
+++ b/tests/test_types.py
@@ -46,18 +46,34 @@ def test_item_type():
     assert 'invalid choice: x' in str(e.value)
 
 
-@mock.patch('planet.scripts.types.get_asset_types',
-            new=mock.Mock(return_value=DEFAULT_ASSET_TYPES))
-def test_asset_type():
-    check = convert_asserter(AssetType())
-    check('visual*', ['visual', 'visual_xml'])
-    check('*data_*', ['metadata_aux', 'metadata_txt'])
-    check('analytic', ['analytic'])
-    check('analytic,visual', ['analytic', 'visual'])
+@mock.patch(
+    "planet.scripts.types.get_asset_types",
+    new=mock.Mock(return_value=DEFAULT_ASSET_TYPES),
+)
+@pytest.mark.parametrize(
+    "pattern,type_list",
+    [
+        ("visual*", ["visual", "visual_xml"]),
+        ("*data_*", ["metadata_aux", "metadata_txt"]),
+        ("analytic", ["analytic"]),
+        ("analytic,visual", ["analytic", "visual"]),
+    ],
+)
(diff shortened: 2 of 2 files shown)


planet/scripts/item_asset_types.py
```diff
@@ -78,7 +78,7 @@
     'ps4b_basic_analytic_nitf', 'ps4b_basic_analytic_rpc_nitf',
     'ps4b_basic_analytic_xml_nitf', 'ps4b_udm', 'udm', 'ortho_udm2',
     'video_file', 'video_frames', 'video_metadata', 'ortho_visual',
-    'visual', 'visual_xml'
+    'analytic', 'visual', 'visual_xml'
```
sgillies, line 81: Like "visual", "analytic" was missing from the list of default asset types. That was the core problem.

tests/test_types.py
```diff
@@ -46,18 +46,34 @@ def test_item_type():
     assert 'invalid choice: x' in str(e.value)
 
 
-@mock.patch('planet.scripts.types.get_asset_types',
-            new=mock.Mock(return_value=DEFAULT_ASSET_TYPES))
-def test_asset_type():
-    check = convert_asserter(AssetType())
```
sgillies, line 52: The tests for v1 of this package very often use assertion helpers like this convert_asserter. It is unnecessary and the tests are easier to read and comprehend without it. Let's not do this in future work.

tests/test_types.py
```diff
@@ -46,18 +46,34 @@ def test_item_type():
...
-    check('visual*', ['visual', 'visual_xml'])
-    check('*data_*', ['metadata_aux', 'metadata_txt'])
-    check('analytic', ['analytic'])
-    check('analytic,visual', ['analytic', 'visual'])
```
sgillies, line 56: Each of these checks should be its own test function. When serialized an early failure can hide a later from being seen until the tests are run again.

tests/test_types.py
```diff
@@ -46,18 +46,34 @@ def test_item_type():
...
+    new=mock.Mock(return_value=DEFAULT_ASSET_TYPES),
+)
+@pytest.mark.parametrize(
+    "pattern,type_list",
+    [
+        ("visual*", ["visual", "visual_xml"]),
+        ("*data_*", ["metadata_aux", "metadata_txt"]),
+        ("analytic", ["analytic"]),
+        ("analytic,visual", ["analytic", "visual"]),
+    ],
```
sgillies, line 60: Pytest's parametrization is how we turn that one test with N steps into N tests of 1 step.

tests/test_types.py
```diff
@@ -46,18 +46,34 @@ def test_item_type():
...
+)
+def test_asset_type_convert(pattern, type_list):
+    """Patterns are correctly converted to a list of assert types."""
+    assert sorted(AssetType()(pattern)) == sorted(type_list)
 
-    with pytest.raises(Exception) as e:
-        AssetType().convert('x', None, None)
-    assert 'invalid choice: x' in str(e.value)
```
sgillies, line 60: For the same reasons, this check should be its own separate test.

tests/test_types.py
```diff
@@ -46,18 +46,34 @@ def test_item_type():
...
+    [
+        ("visual*", ["visual", "visual_xml"]),
+        ("*data_*", ["metadata_aux", "metadata_txt"]),
+        ("analytic", ["analytic"]),
+        ("analytic,visual", ["analytic", "visual"]),
+    ],
+)
+def test_asset_type_convert(pattern, type_list):
+    """Patterns are correctly converted to a list of assert types."""
+    assert sorted(AssetType()(pattern)) == sorted(type_list)
```
sgillies, line 64: The code in convert_asserter is replaced by a single line. The assertion is in the body of the test, no need to check elsewhere to see what's going on.

PR #359, opened 2022-01-13
got rid of coverage file that was mistakenly added to the merge

mkshah605:
Also added asterisk in git ignore to coverage files to prevent this in the future


kevinlacaille:
Nice catch! Thank you for this.

jreiberkyle (approved):
LGTM!

Issue #361, opened 2022-01-18
revisit how we handle error messages from the server


jreiberkyle:
Of all of the SDK code, the code for handling error messages from the server is my least favorite.

[this](https://github.com/planetlabs/planet-client-python/blob/82bd7e4a80db2634c08c0c4512c58ee13f3191bb/planet/clients/orders.py#L160) is an example - it is a lot of custom code for converting the error response json into readable text. there are many more examples: [here](https://github.com/planetlabs/planet-client-python/blob/82bd7e4a80db2634c08c0c4512c58ee13f3191bb/planet/clients/orders.py#L197), [here](https://github.com/planetlabs/planet-client-python/blob/82bd7e4a80db2634c08c0c4512c58ee13f3191bb/planet/clients/orders.py#L230), etc

I created this code to make the errors more readable by a human, that is, in sentence structure. It felt a little bit jolting to be using the CLI to, say, download an order, then get a json blob printed to the string when, for example, the order didn't exist.

Let's revisit and answer the questions:
- is it valuable to parse the json blob returned by the server into sentence structure?
- if so, could it be done more gracefully. for example, is there a way to get the json blob structure from the openAPI schema?

jreiberkyle:
this is done, error messages are raised, unaltered.

PR #370, opened 2022-01-21
Chain order creation exception for layers of detail about errors

sgillies:
Towards resolving #361


planet/clients/orders.py
```diff
@@ -149,30 +149,23 @@ async def create_order(
...
-            msg = msg_json['general'][0]['message']
-            try:
-                # get first error field
-                field = next(iter(msg_json['field'].keys()))
-                msg += ' - ' + msg_json['field'][field][0]['message']
-            except AttributeError:
-                pass
-            raise exceptions.BadQuery(msg)
+        except exceptions.BadQuery as bad_query:
+            raise OrderError("Order creation failed.") from bad_query
```
sgillies, line 166: Here's the gist of it: for the Python API we chain the exceptions. The user will get an OrderError with a brief, unambiguous message and this carries the BadQuery exception with it. And then we pass the responsibility for presenting the BadQuery message to a human onto the caller of create_order. Advantage for the Python API is that we generalize our exception so that it can suit the needs of almost any caller and eliminate code. Advantage for a user is that they are not bound by choices we make.

sgillies: @jreiberkyle what do you think about hiding the request/response details and raising a more general OrderError for create_order?

sgillies:
Ready for some preliminary review @jreiberkyle @kevinlacaille. Tests fail because I haven't yet updated them to expect OrderError, will do that later.

planet/exceptions.py
```diff
@@ -68,3 +68,7 @@ class InvalidIdentity(APIException):
 class RequestCancelled(Exception):
     '''Internal exception when a request is cancelled'''
     pass
+
+
+class OrderError(APIException):
```
jreiberkyle, line 73: +1 to an exception that is specific to each client. I am curious why you chose Error here instead of Exception.

planet/clients/orders.py
jreiberkyle: I am going back and forth here quite a bit. I like the clarity that adding the `OrderError` layer brings in that it can specify that the order creation failed. However, it seems like we are handing the user quite a bit of complication if we are telling them to use `error.__context__.message` to access the real source of the failure. When the two usages are compared, I have to say that the second one seems easier. And there are reasons besides API interactions that can cause a failed order - like, for example, if some of the user input wasn't quite right.

```python
try:
    order = client.create_order(...)
except OrderError as error:
    print("Order creation failed. Service response details: {}".format(error.__context__.message))
```

```python
try:
    order = client.create_order(...)
except APIException as exception:
   print("Order creation failed. Service response details: {}".format(exception.message))
```

planet/scripts/cli.py
```diff
@@ -324,6 +324,11 @@ async def create(ctx, name, ids, bundle, item_type, email, cloudconfig, clip,
         tools=tools)
 
     async with orders_client(ctx) as cl:
-        order = await cl.create_order(request)
+        try:
+            order = await cl.create_order(request)
+        except OrderError as error:
+            service_resp = json.loads(err.__context__.message)
+            LOGGER.info("Order failed: service_resp=%r", service_resp)
+            raise click.ClickException(error.message)
```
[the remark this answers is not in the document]
jreiberkyle: It seems beneficial to give the users some insight into the cause of the failure without them having to access and sift through logs. It would be helpful to tell them, say, whether they entered bad data for the order creation or they are up against quota limits. I would lean toward giving them the raw service response over making them work for the information. Yes, changing the logging level is an option. It just seems like it should be used for debugging issues that are more infrequent than order creation failing.

planet/exceptions.py
sgillies: `*Error` is the Python community consensus favorite. The standard library has `ValueError` not `ValueException`. Attrs and urllib3 are two of the most widely used non-standard libs and those use Error in the names: https://urllib3.readthedocs.io/en/latest/reference/urllib3.exceptions.html, https://github.com/python-attrs/attrs/blob/main/src/attr/exceptions.py. It's recommended in the Google Python style guide, too, which we kinda adhere to because of yapf, right?

planet/clients/orders.py
sgillies: Yes, it's very possible that users will want easier access to the details about the root cause of a server error. I propose we deal with that later and for now go with the chaining, which doesn't lose any information from the server and provides a good base for future improvements.

planet/scripts/cli.py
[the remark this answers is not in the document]
sgillies: I think I see what you mean: that the service response details could get lost in info level messages from our dependencies? It's true that urllib3 is super chatty and httpx or aiohttp might also be. In that case, `echo(service_resp, err=True)` could be the way to go. How about I do that for now if `-v` is on and we iterate on the best approach as we go?

sgillies:
@jreiberkyle thanks for the comments! I'll get the tests to pass before I mark this ready for final review.

Issue #362, opened 2022-01-20
Reorganize the CLI structure


jreiberkyle:
Rename the CLI code directory from scripts to cli and break up the cli script into cli, auth, and orders.

Issue #327, opened 2021-10-07
Convert CLI tests to integration tests


jreiberkyle:
Right now we unit test the CLI and mock out the rest of the planet library. Because we do not perform integration tests with the CLI, we miss errors that occur when the CLI passes unexpected arguments to the library. We could catch these bugs with integration tests that incorporate the CLI. Since the CLI is a thin wrapper around the API, it isn't much of a win to test the CLI alone with unit tests. Therefore, move to integration tests where we don't mock out the rest of the python API but mock the servers.

PR #377, opened 2022-01-22
Refactor CLI

jreiberkyle:
What changed:
- Rename CLI folder from scripts to cli
- Break CLI into modules: main, orders, auth, cmds, and io
- Break up CLI testing into three modules: test_cli_main, test_cli_auth, and test_cli_orders.

What didn't change:
- Functionality of the cli
- Test functionality

Small side effects:
- remove hidden coverage file (why #359 didn't remove this file, I do not know)
- update copyright of affected files to planet labs pbc.
- pin pytest-asyncio version to avoid deprecation warnings (chose to pin vs updating because if we move to aiohttp then we will no longer use pytest-asyncio)
- Add PlanetException root exception to exceptions module
- Move AuthException over to exceptions module

Part of #362


sgillies:
@jreiberkyle pytest-asyncio is very useful with aiohttp and it seems to be a popular combo.

planet/cli/utils.py
```diff
@@ -0,0 +1,58 @@
...
+
+
+def get_auth():
+    try:
+        auth = planet.Auth.from_file()
+    except planet.auth.AuthException:
+        raise click.ClickException(
+            'Auth information does not exist or is corrupted. Initialize '
+            'with `planet auth init`.')
+    return auth
```
sgillies, line 50: Is this function necessary? If I understand the intent of handle_exceptions(), we could use it to wrap any exception from planet.Auth.from_file(), yes?

jreiberkyle:
@sgillies pytest-aiohttp ([pypi](https://pypi.org/project/pytest-aiohttp/), [github](https://github.com/aio-libs/pytest-aiohttp)) is developed by the same organization as aiohttp and is at version 1.0.3 while pytest-asyncio ([pypi](https://pypi.org/project/pytest-asyncio/), [github](https://github.com/pytest-dev/pytest-asyncio)) is at version 0.17.2 and is developed by the pytest organization. It will be worth our time to determine which we would want to go with, should we switch to aiohttp. For now, with that discussion looming, it seems reasonable to punt (with *intention*)

planet/cli/utils.py
sgillies: I'm ever concerned about the potential for utils modules to turn into a "junk drawer" (and I've seen it). It would be good to not add anything that we can do without.

planet/cli/utils.py
```diff
@@ -0,0 +1,58 @@
...
+# https://github.com/pallets/click/issues/85#issuecomment-503464628
+def coro(f):
+    '''Wraps async functions so they can be run sync with Click.'''
+    @wraps(f)
+    def wrapper(*args, **kwargs):
+        return asyncio.run(f(*args, **kwargs))
+    return wrapper
+
+
+def handle_exceptions(f):
```
sgillies, line 33: "handle exceptions" is a little bit too vague to be a good name.

```suggestion
def translate_exceptions(func):
    """Translate internal exceptions to ClickException."""
```

sgillies: And sometimes single char variable names get in the way of debugging. Maybe we could limit use to `i` for looping or `f` for file objects, but not in function declarations?

planet/cli/utils.py
```diff
@@ -0,0 +1,58 @@
...
+    try:
+        auth = planet.Auth.from_file()
+    except planet.auth.AuthException:
+        raise click.ClickException(
+            'Auth information does not exist or is corrupted. Initialize '
+            'with `planet auth init`.')
+    return auth
+
+
+def json_echo(json_dict, pretty):
```
sgillies, line 53:
```suggestion
def echo_json(obj, pretty=False):
```

What would you think of "verb, noun" naming here? Also the first positional argument is handed directly to json.dumps and isn't required to be a dict.
A keyword argument

sgillies: The body of the function could be as small as `click.echo(json.dumps(obj, sort_keys=pretty, indent=(2 * pretty or None)))`. This eliminates an if/else, which is almost always a good thing.

sgillies (changes requested):
@jreiberkyle I've got a couple changes to suggest, see inline comments.

Overall, this is a good module layout. If we put all the commands in a single file we wouldn't need anything like a "utils", but the file could eventually grow to ~20 commands and get unwieldy. One API/service per module feels good to me.

Generally, instead of a single "utils" for shared code, I would like to see them split up into related groups. Like a "cli/cmds.py" for the command decorators and a "cli/io.py" for echo_json() etc.

sgillies:
@jreiberkyle pytest-aiohttp requires pytest-asyncio https://github.com/aio-libs/pytest-aiohttp/blob/master/setup.cfg#L44 and provides features we don't need and won't use in unit testing. In bigger integration type tests, maybe.

planet/cli/utils.py
jreiberkyle: This is an excellent change and docstring, thank you for providing this suggestion! And I'm with you on the single char variable name, good point.

planet/cli/utils.py
jreiberkyle: Yes to the rename and agreed dict-> obj. For the body of the function, I think the if/else approach has the benefit of readability

planet/cli/auth.py
```diff
@@ -0,0 +1,65 @@
...
+    click.echo('Initialized')
+
+
+@auth.command()
+def value():
+    '''Print the stored authentication information'''
+    click.echo(get_auth().value)
+
+
+def get_auth():
```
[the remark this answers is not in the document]
jreiberkyle: ahh yes, coming right up!

jreiberkyle:
Moving the conversation around pytest-asyncio and pytest-aiohttp to #379 so it doesn't derail this PR.

jreiberkyle:
@sgillies 

> Generally, instead of a single "utils" for shared code, I would like to see them split up into related groups. Like a "cli/cmds.py" for the command decorators and a "cli/io.py" for echo_json() etc.

I share your mistrust of utils modules and I like the idea of splitting the contents into cmds.py and io.py. Great suggestion!

planet/cli/utils.py
jreiberkyle: great point! we can put this in handle_exceptions(). will do.

planet/cli/utils.py
jreiberkyle: a note on the `f` var. I was just familiarizing myself with @wraps and noticed that the official Python documentation uses `f` as the argument name for the example usage of `wraps` ([ref]:(https://docs.python.org/3/library/functools.html#functools.wraps))

```python
>>> from functools import wraps
>>> def my_decorator(f):
...     @wraps(f)
...     def wrapper(*args, **kwds):
...         print('Calling decorated function')
...         return f(*args, **kwds)
...     return wrapper
```

planet/cli/auth.py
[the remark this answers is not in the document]
jreiberkyle: oh whew, I don't have to document this function since it's gone now :)

jreiberkyle:
@sgillies this is ready for review again.

As an aside: Something that isn't in this work but is bugging me is the CLI test coverage. Right now, the CLI is tested with unit tests, which mock out the rest of the planet library. This is great for testing the CLI alone but has failed to catch bugs where the CLI calls the python API, but does so erroneously. Since the CLI is a thin wrapper around the API, I am tempted to move the unit tests to integration tests. As integration tests, we won't mock out the python API. I am tempted to do this earlier than later as we are planning some significant work with the CLI. What are your thoughts?

sgillies (approved):

sgillies:
@jreiberkyle using, for example, the actual download_order() (but with a fake service response)? That seems right to me.

jreiberkyle:
@sgillies yep. See #327 for more on the integration tests.

planet/exceptions.py
```diff
@@ -13,8 +13,13 @@
 # limitations under the License.
 
 
-class APIException(Exception):
-    '''General unexpected response'''
+class PlanetException(Exception):
```
jreiberkyle, line 16: @sgillies just want to point out that I added this in here. Nice to have a root exception for the project instead of just extending `Exception` directly.

PR #378, opened 2022-01-22
Change auth client base url to the api url

### LINKED ISSUE #375: change auth base url to the api url
The orders client base url points to the orders API endpoint. Changing the base url replaces the entire url path. Right now, the auth client base url is just the root url, and the relative path to the auth API endpoint is added in the client. Changing the base url only changes the root url. Bring the functionality of the auth client (and, therefore, auth CLI) into alignment with the functionality of the orders api: that is, change the auth base url to the api endpoint.

jreiberkyle:
Changes:
- change auth client base url to the api url instead of the planet services root url
- update auth api integration test to change in base url
- add integration test for auth cli that confirms base url is changed

Side effect:
- copyright planet labs inc to planet labs pbc

Closes #375

### DIFF
--- a/planet/auth.py
+++ b/planet/auth.py
@@ -26,7 +26,7 @@ from . import constants, http, models
 
 LOGGER = logging.getLogger(__name__)
 
-BASE_URL = constants.PLANET_BASE_URL
+BASE_URL = constants.PLANET_BASE_URL + 'v0/auth/'
 ENV_API_KEY = 'PL_API_KEY'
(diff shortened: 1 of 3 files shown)


sgillies (approved):
@jreiberkyle seems like a good change. String format() or f-strings perform better than `str + str`, that's the only possible improvement I see.

jreiberkyle:
Agreed, f-strings are the way to go here as they have both great performance and readability. This is captured in #382. I chose a ticket for this change because, while it is a valuable change, it should also be done in the orders client and it is lower priority than what this PR is blocking.

PR #385, opened 2022-02-01
Convert CLI tests to integration tests, increase overall test coverage, slight change to API url building logistics

### LINKED ISSUE #327: Convert CLI tests to integration tests

### LINKED ISSUE #382: move url creation in the auth and orders clients to f-strings
Currently, the auth and orders clients build urls with concatenation. f-strings are more performant while maintaining readability. Move url creation to f-strings for both the auth and orders clients.

jreiberkyle:
### Changes:
* CLI tests are now integration tests, where none of the planet library is mocked, just server responses are mocked. While many of the integration tests are 1:1 with the unit tests, some new tests were added and some existing tests were broken into multiple integration tests. Overall coverage increased.
* Minimum test coverage to pass tests is increased from 95% to 97%
* Change where slashes are in API url building to match API documentation (example pulled from docs below, note how relative paths start with a slash)
image
* URLs are now built using f-strings instead of string concatenation (closes #382)

Closes #327


tests/unit/test_cli_io.py
```diff
@@ -0,0 +1,30 @@
...
+
+from planet.cli import io
+
+
+@pytest.mark.parametrize(
+    "pretty,expected",
+    [(False, '{"key": "val"}'), (True, '{\n  "key": "val"\n}')])
+def test_cli_echo_json(pretty, expected, monkeypatch):
+    mock_echo = Mock()
+    monkeypatch.setattr(io.click, 'echo', mock_echo)
```
sgillies, line 26: @jreiberkyle what would you think about using mock.patch instead of pytest's monkeypatch?

```suggestion
@pytest.mark.parametrize(
    "pretty,expected",
    [(False, '{"key": "val"}'), (True, '{\n  "key": "val"\n}')])
@mock.patch('io.click.echo')
def test_cli_echo_json(mock_echo, pretty, expected, monkeypatch):
```

sgillies (approved):
I left one suggestion inline for using mock.patch instead of pytest.monkeypatch in the case where we're using mock.Mock.

tests/unit/test_cli_io.py
jreiberkyle: This works beautifully with one change: patch uses the full path name to the module. (see paragraph discussing the *target* arg in the [patch docs](https://docs.python.org/3/library/unittest.mock.html#patch))

PR #389, opened 2022-02-02
Remove reported data from stdout for CLI

### LINKED ISSUE #386: Remove reported data from CLI download
Currently, when requesting a download, the CLI reports the number of files in an order. 

For example, for an order with 12 files here's what one would see to stdout:

```
❯ planet orders download my-oid
Downloaded 12 files.
```

**Proposal:**
After [lengthy discussions](https://github.com/planetlabs/planet-client-python/pull/355#issuecomment-1010215231), I propose that the CLI should remain silent upon a download request. Specifically, the download request should not output to stdout nor stderr, but using the INFO log, which already exists in `body.write()`, to communicate the files downloaded.

**To do:**
1) Remove `click.echo(f'Downloaded {len(filenames)} files.')` in `planet.cli.orders.download()` (line 126)
2) Update the CLI's test in `tests.unit.test_cli_orders.test_cli_orders_download()` to no longer expect "`'Downloaded 4 files.\n'`"

kevinlacaille:
**Overview:**
- Previously, the CLI would report the number of files in an order.
- Now, the CLI remains silent in stdout upon a download request. 

**Work done:**
1) Removed `click.echo(f'Downloaded {len(filenames)} files.')` in `planet.cli.orders.download()`, which reported the number of files in an order. The filenames are still communicated in `body.write()` in `models.py` in `LOGGER.debug()` and `LOGGER.info()`.
2) Updated the CLI's test in `tests.unit.test_cli_orders.test_cli_orders_download()` to no longer expect anything communicated to stdout.

Closes #386

### DIFF
--- a/planet/cli/orders.py
+++ b/planet/cli/orders.py
@@ -118,12 +118,11 @@ async def download(ctx, order_id, quiet, overwrite, dest):
         with planet.reporting.StateBar(order_id=order_id,
                                        disable=quiet) as bar:
             await cl.poll(str(order_id), report=bar.update)
-            filenames = await cl.download_order(
+            _ = await cl.download_order(
                     str(order_id),
                     directory=dest,
                     overwrite=overwrite,
                     progress_bar=not quiet)
-    click.echo(f'Downloaded {len(filenames)} files.')
 
 
 def split_id_list(ctx, param, value):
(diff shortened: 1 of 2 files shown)


jreiberkyle (approved):
Looks great! Gret job! I'd usually say something about documentation changing, but it looks like the CLI documentation is already incredibly sparse and the CLI Interface implementations will beef it up.

Issue #352, opened 2022-01-05
Reporting breaks when quiet=True


kevinlacaille:
**Summary:** While setting `quiet = True` in the `download()` function in the CLI I get an attribute error with `tqdm_asyncio`.

**Replicate error:**
Add `quiet = True` to line 214 on v2 in cli.py in `download()` so that the function looks like this:
```
async def download(ctx, order_id, quiet, overwrite, dest):
    '''Download order by order ID.'''

    # Download the user's order
    async with orders_client(ctx) as cl:
        quiet = True
        with planet.reporting.StateBar(order_id=order_id,
                                       disable=quiet) as bar:
            await cl.poll(str(order_id), report=bar.update)
            filenames = await cl.download_order(
                    str(order_id),
                    directory=dest,
                    overwrite=overwrite,
                    progress_bar=not quiet)
    click.echo(f'Downloaded {len(filenames)} files.')
```

Then, try and download an order:
```
❯ planet orders download 4acfe01d-6fb9-494f-ae11-ee80ab92d58d
```

Associated traceback:
```
Traceback (most recent call last):
  File "/Users/kevin.lacaille/.pyenv/versions/planetSDK/bin/planet", line 33, in <module>
    sys.exit(load_entry_point('planet', 'console_scripts', 'planet')())
  File "/Users/kevin.lacaille/.pyenv/versions/planetSDK/lib/python3.9/site-packages/click/core.py", line 1137, in __call__
    return self.main(*args, **kwargs)
  File "/Users/kevin.lacaille/.pyenv/versions/planetSDK/lib/python3.9/site-packages/click/core.py", line 1062, in main
    rv = self.invoke(ctx)
  File "/Users/kevin.lacaille/.pyenv/versions/planetSDK/lib/python3.9/site-packages/click/core.py", line 1668, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/Users/kevin.lacaille/.pyenv/versions/planetSDK/lib/python3.9/site-packages/click/core.py", line 1668, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/Users/kevin.lacaille/.pyenv/versions/planetSDK/lib/python3.9/site-packages/click/core.py", line 1404, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/Users/kevin.lacaille/.pyenv/versions/planetSDK/lib/python3.9/site-packages/click/core.py", line 763, in invoke
[...]

PR #399, opened 2022-02-24
Rename poll to wait, separate wait and download in cli

### LINKED ISSUE #381: Implement wait interface, rename poll API command
This tickets covers the following changes:
1. Implement `planet orders wait` CLI command
1. Document use cases
1. Rename orders API poll command to wait

### Implement `planet orders wait` CLI command
Implement a new orders cli command, `wait` with the following interface:

```
Usage: planet orders wait [OPTIONS] [ORDER_ID]

  Wait for order to reach the desired state.

  Polls the service for order status and completes when the order has reached
  the desired state. By default, the desired state is any completed state.
  This can be overridden to any order state with the --state option. The
  polling period is one second, this can be overridden with the --period
  option.

  This command outputs the order details at the end of the wait, optionally
  pretty-printed.

Options:
  --state TEXT   Stop waiting when order reaches this state
  --period TEXT  Duration of time for one polling cycle, in seconds.
  --pretty       Format JSON output.
  --help         Show this message and exit.

```

The `wait` command should also use the `--quiet` option (#369) to optionally disable progress bar reporting.

### Document use cases
Document the following use cases:

User Story: As a CLI user I would like to wait for an order to be ready for download so I can send a message to another machine to do the download.

```
$ planet orders wait 65df4eb0-e416-4243-a4d2-38afcf382c30
```

User Story: As a CLI user I would like to wait for an order to be ready for download and then I would like to download the order.

```
$ planet orders wait 65df4eb0-e416-4243-a4d2-38afcf382c30 \
    && planet orders download 65df4eb0-e416-4243-a4d2-38afcf382c30
```

### Rename orders API poll command to wait
To align the python API and CLI interfaces, rename OrdersClient `poll()` to `wait()`

Blocked by #369

jreiberkyle:
This PR implements the following:
- rename `OrdersClient.poll()` to `OrdersClient.wait()`
- in `OrdersClient.wait()`, add `max-attempts` parameter and rename `wait` parameter to `delay`.
- `OrdersClient.wait()` now checks if a specified state has been reached **or passed**. This requires introducing the concept of state sequence and is handled with `orders.ORDER_STATE_SEQUENCE` and `orders.OrderStates`
- in `OrdersClient.wait()`, `report` parameter renamed to `callback`. `callback` no longer takes a keyword argument, state is a required positional argument.  Add `update()` function to `reporting.StateReporter` that fits this pattern.
- add new `planet orders wait` cli command and remove polling/waiting from `planet orders download`
- fix bug in `reporting.StateReporter` when bar is disabled
- all exceptions are either `APIException` or `ClientError`

Closes #381 and #352

### DIFF
--- a/examples/orders_create_and_download_multiple_orders.py
+++ b/examples/orders_create_and_download_multiple_orders.py
@@ -75,8 +75,8 @@ async def create_and_download(order_detail, directory, client):
         order = await client.create_order(order_detail)
         reporter.update(state='created', order_id=order.id)
 
-        # poll
-        await client.poll(order.id, report=reporter.update)
+        # wait for completion
+        await client.wait(order.id, report=reporter.update_state)
 
     # download
     await client.download_order(order.id, progress_bar=True)
--- a/planet/cli/orders.py
+++ b/planet/cli/orders.py
@@ -158,7 +158,7 @@ async def wait(ctx, order_id, delay, max_attempts, quiet, state):
                                   state=state,
                                   delay=delay,
                                   max_attempts=max_attempts,
-                                  report=bar.update_state)
+                                  callback=bar.update_state)
     click.echo(state)
 
 
--- a/planet/cli/orders.py
+++ b/planet/cli/orders.py
@@ -98,13 +98,83 @@ async def cancel(ctx, order_id):
     click.echo('Cancelled')
 
 
+def split_list_arg(ctx, param, value):
+    if value is None:
+        return None
+    elif value == '':
+        # note, this is specifically checking for an empty string
+        click.BadParameter('Entry cannot be an empty string.')
+
+    # split list by ',' and remove whitespace
+    entries = [i.strip() for i in value.split(',')]
+
+    # validate passed entries
+    for e in entries:
+        if not e:
+            raise click.BadParameter('Entry cannot be an empty string.')
+    return entries
+
+
+@orders.command()
+@click.pass_context
+@translate_exceptions
+@coro
+@click.argument('order_id', type=click.UUID)
+@click.option('--delay', type=int, default=5,
+              help='Time (in seconds) between polls.')
+@click.option('--max-attempts', type=int, default=5,
+              help='Maximum number of polls. Set to zero for no limit.')
+@click.option('--quiet', is_flag=True, default=False,
+              help='Disable ANSI control output.')
+@click.option('--states', type=click.STRING, callback=split_list_arg,
+              help=('One or more comma-separated order states that will end '
+                    'polling. Defaults to all completed states.'))
+async def wait(ctx, order_id, delay, max_attempts, quiet, states):
+    """Wait until order reaches one of the specified states.
+
+    Reports the state of the order when waiting is complete.
+
+    This function polls the Orders API to determine the order state, with
+    the specified delay between each polling attempt, until the
+    order reaches one of the set of desired states. If the maximum number
+    of attempts is reached before the order reaches one of the desired
+    states, an exception is raised. Setting --max-attempts to zero will
+    result in no limit on the number of attempts.
+
+    Setting --delay to zero results in no delay between polling attempts.
+    This will likely result in throttling by the Orders API, which has
+    a rate limit of 10 requests per second. If many orders are being
+    polled asynchronously, consider increasing the delay to avoid
+    throttling.
+
+    By default, the set of states that result are polled for is the
+    set of completed states. This can be changed to any set of valid
+    order states with --states. It is not recommended to change this to a
+    subset of completed states, as this opens up the possibility of the the
+    specified state never being reached (i.e. if the order completes in
+    a 'partial' state but only the 'success' state was specified).
+    """
+    if not states:
+        states = planet.clients.orders.ORDERS_STATES_COMPLETE
+
+    async with orders_client(ctx) as cl:
+        with planet.reporting.StateBar(order_id=order_id,
+                                       disable=quiet) as bar:
+            state = await cl.wait(str(order_id),
+                                  states=states,
+                                  delay=delay,
+                                  max_attempts=max_attempts,
+                                  report=bar.update_state)
+    click.echo(state)
+
+
 @orders.command()
 @click.pass_context
 @translate_exceptions
 @coro
 @click.argument('order_id', type=click.UUID)
 @click.option('-q', '--quiet', is_flag=True, default=False,
-              help=('Disable ANSI control output.'))
+              help='Disable ANSI control output.')
 @click.option('-o', '--overwrite', is_flag=True, default=False,
               help=('Overwrite files if they already exist.'))
 @click.option('--dest', default='.',
@@ -112,30 +182,12 @@ async def cancel(ctx, order_id):
               type=click.Path(exists=True, resolve_path=True,
                               writable=True, file_okay=False))
 async def download(ctx, order_id, quiet, overwrite, dest):
-    '''Download order by order ID.'''
-
-    # Download the user's order
+    """Download order by order ID."""
     async with orders_client(ctx) as cl:
-        with planet.reporting.StateBar(order_id=order_id,
-                                       disable=quiet) as bar:
-            await cl.poll(str(order_id), report=bar.update)
-            _ = await cl.download_order(
-                    str(order_id),
-                    directory=dest,
-                    overwrite=overwrite,
-                    progress_bar=not quiet)
-
-
-def split_id_list(ctx, param, value):
-    # split list by ',' and remove whitespace
-    ids = [i.strip() for i in value.split(',')]
-
-    # validate passed ids
-    for iid in ids:
-        if not iid:
-            raise click.BadParameter('id cannot be empty string.')
-
-    return ids
+        await cl.download_order(str(order_id),
+                                directory=dest,
+                                overwrite=overwrite,
+                                progress_bar=not quiet)
 
 
 def read_file_geojson(ctx, param, value):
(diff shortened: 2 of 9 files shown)


examples/orders_create_and_download_multiple_orders.py
```diff
@@ -75,8 +75,8 @@ async def create_and_download(order_detail, directory, client):
         order = await client.create_order(order_detail)
         reporter.update(state='created', order_id=order.id)
 
-        # poll
-        await client.poll(order.id, report=reporter.update)
+        # wait for completion
+        await client.wait(order.id, report=reporter.update_state)
```
sgillies, line 79: @jreiberkyle what would you think about wait taking an instance of Order (like the one from line 75) instead of a string id? I think that could help make the API more consistent and intuitive?

Also, I'd like to understand the reporter feature better. What is it about?

planet/cli/orders.py
```diff
@@ -98,44 +98,96 @@ async def cancel(ctx, order_id):
...
+@click.argument('order_id', type=click.UUID)
+@click.option('--delay', type=int, default=5,
+              help='Time (in seconds) between polls.')
+@click.option('--max-attempts', type=int, default=5,
+              help='Maximum number of polls. Set to zero for no limit.')
+@click.option('--quiet', is_flag=True, default=False,
+              help='Disable ANSI control output.')
+@click.option('--states', type=click.STRING, callback=split_list_arg,
+              help=('One or more comma-separated order states that will end '
+                    'polling. Defaults to all completed states.'))
```
sgillies, line 131: Do you think it's possible to dependably wait on an intermediate state? I don't think it is, so perhaps we could rule those out entirely.

And I sorta feel like all completed states (success or failure) should be the same to the waiter. Allowing users to wait for failure when the process is intended to succeed 99% of the time seems not very useful.

planet/cli/orders.py
```diff
@@ -98,44 +98,96 @@ async def cancel(ctx, order_id):
...
+    return entries
+
+
+@orders.command()
+@click.pass_context
+@translate_exceptions
+@coro
+@click.argument('order_id', type=click.UUID)
+@click.option('--delay', type=int, default=5,
+              help='Time (in seconds) between polls.')
```
sgillies, line 124: I wonder if we might want to remove this option and start with a fixed delay? It might turn out that we want to use an exponential back-off down the road for orders that take a rather long time.

planet/clients/orders.py
```diff
@@ -80,14 +74,12 @@ def __init__(
 
     @staticmethod
     def _check_order_id(oid):
-        msg = f'Order id ({oid}) must be a valid UUID string.'
-        if not isinstance(oid, str):
-            raise OrdersClientException(msg)
-
+        """Raises planet.exceptions.ValueError if oid is not a valid UUID"""
```
sgillies, line 77: Reusing the name of a built-in exception could lead to problems in users' code.

sgillies: How about just an OrderError for almost everything order related? Until we learn that more specific exceptions are really needed.

planet/clients/orders.py
```diff
@@ -324,8 +318,17 @@ async def download_order(
 
         Raises:
             planet.exceptions.APIException: On API error.
+            planet.exceptions.StateError: If the order is not in a completed
+                state.
         """
         order = await self.get_order(order_id)
+        if order.state not in ORDERS_STATES_COMPLETE:
+            raise exceptions.StateError(
```
sgillies, line 326: I feel like StateError is too specific. Maybe just OrderError again? Less is more with exception types, I think.

planet/clients/orders.py
```diff
@@ -337,71 +340,90 @@ async def download_order(
...
-            if not completed:
-                sleep_time = max(wait-(time.time()-t), 0)
+            done = state in states
+            if not done:
+                if max_attempts:
+                    num_attempts += 1
+                    if num_attempts >= max_attempts:
+                        raise exceptions.MaxAttemptsError(max_attempts)
+
+                sleep_time = max(delay-(time.time()-t), 0)
```
sgillies, line 426: I don't understand why we're trying to base this on the execution time of the code. Is this to try to make the user-perceived delay constant? What if we just make the approximation that `time.time() - t` is 0?

planet/clients/orders.py
```diff
@@ -337,71 +340,90 @@ async def download_order(
...
-                report(state=order.state)
-            else:
-                LOGGER.debug(state)
+                report(order.state)
 
-            completed = state in states
-            if not completed:
-                sleep_time = max(wait-(time.time()-t), 0)
+            done = state in states
+            if not done:
```
sgillies, line 420: Two different kinds of conditionals testing `done` is a bit of a red flag for me. I think this code would be easier to understand if it used one `while i < max_attempts` and used `break` to get out early.

planet/clients/orders.py
```diff
@@ -337,71 +340,90 @@ async def download_order(
...
-            Completed state of the order.
+            State of the order.
 
         Raises:
             planet.exceptions.APIException: On API error.
-            OrdersClientException: If order_id is not valid or state is not
-                supported.
+            planet.exceptions.ValueError: If order_id or one or more of the
+                states are not valid.
+            planet.exceptions.MaxAttemptsError: If the maximum number of
```
sgillies, line 396: Yep, we need an error like this. MaxAttemptsReachedError might be more descriptive? Or SickOfWaitingError :laughing:

planet/clients/orders.py
```diff
@@ -337,71 +340,90 @@ async def download_order(
...
+    async def wait(
         self,
         order_id: str,
-        state: str = None,
-        wait: int = 1,
-        report=None
+        states: typing.Iterable[str] = ORDERS_STATES_COMPLETE,
+        delay: int = 5,
+        max_attempts: int = 200,
+        report: typing.Callable[[str], None] = None
```
sgillies, line 349: How generalizing this to be named `callback`? Reporting is one kind of use but not necessarily the only one. And maybe it could be passed the entire order description and not only the status?

planet/cli/orders.py
sgillies: We had a call about this and @jreiberkyle pointed we *can* wait on an intermediate state having been reached, i.e. being at that state or a later state. So let's not rule this out. I'll resolve this conversation.

planet/clients/orders.py
sgillies: Was reminded by @jreiberkyle that an order's state is the only changing attribute, so there's no real need for the entire description in the callback.

planet/cli/orders.py
jreiberkyle: Hmm.. this is a good thought. I think if we get to where we want to add exponential delay, we can expand the option type so that the value can either be an integer or 'exponential'. I believe this option is valuable because users may have different wait sensitivities.

planet/cli/orders.py
jreiberkyle: After our conversation, I believe this could be improved to have a wait on an intermediate state return any time after that state was reached, even if polling didn't catch the state itself. I'll keep this open until I can add this improvement.

planet/clients/orders.py
jreiberkyle: yes, changing it to `callback` is a great idea

planet/clients/orders.py
jreiberkyle: LOL or the more general WTFError. I can change it to MaxAttemptsReachedError.

planet/clients/orders.py
jreiberkyle: Yep, this is to make the delay constant. As for changing this, why would we approximate when we can calculate? It doesn't seem like it's too heavy a lift operationally.

planet/clients/orders.py
jreiberkyle: Yeah, it could lead to problems. I agree. I would actually throw a OEM `ValueError`, but I'm trying to have all exceptions derive from `PlanetError`. My preference here is to have an exception that handles "invalid argument type" cases and is derived from `PlanetError` vs an exception that is specific to the Orders client. Just because this type of thing is likely to show up in all of the clients. What about `ValidationError(PlanetError)`?

planet/clients/orders.py
jreiberkyle: My thinking on adding `StateError` is to indicate situations where a waiter might be useful. For an order, this is when it's not in a completed state. For an image in the Data API, this is when the image is not active. I am not as clear on where/if it happens in the other APIs. As for the less is more approach, it seems to make sense but I just haven't read enough about exception design to know for sure. Do you have any resources or examples that could provide context?

planet/clients/orders.py
jreiberkyle: that makes sense. I'll give it a go.

planet/clients/orders.py
jreiberkyle: Ohh! Found an example in the wild - [botocore.exceptions.ValidationError](https://github.com/boto/botocore/blob/21b06efb39da61fe4101ab9709140dc1e51dc33d/botocore/exceptions.py#L294)

planet/clients/orders.py
jreiberkyle: actually, what about [RetriesExceededError](https://github.com/boto/boto3/blob/c236fa1d230ea2f4738d8d5e5961d587ef006299/boto3/exceptions.py#L70)?

jreiberkyle: just falling back to ClientError for now

planet/clients/orders.py
jreiberkyle: just falling back to ClientError for now

Issue #395, opened 2022-02-04
Include yapf or black in linting


jreiberkyle:
The results from [yapf](https://github.com/google/yapf) using the default style, pep8, seem pretty reasonable. Black is another popular option. Black is not configurable, yet yapf may be non-deterministic ([Hacker News thread](https://news.ycombinator.com/item?id=17155048)). 

Chose which to use, then run it on the `planet` and `tests` directories. Add it to the automatic linting for the project, and make reference to it in the CONTRIBUTING doc.

jreiberkyle:
Increase priority, re: https://github.com/planetlabs/planet-client-python/pull/403#discussion_r817266556

jreiberkyle:
yapf is closest to the existing style, so that's what we will use

PR #403, opened 2022-03-01
Replace Order with an attrs dataclass

### LINKED ISSUE #402: Make Order an attrs dataclass
The project already depends on attrs and this could reduce boilerplate and give us a more tested implementation.

sgillies:
Shift responsibility for turning API responses into instances of Order to OrdersClient.

Previously, the Order class constructor took a blob of JSON directly from the Orders API response and did some computation to come up with an order description object. This coupled the API response directly to our information model. I have a hypothesis that using a less tightly coupled Order data class and shifting the responsibility for translation to the OrdersClient will give us more freedom and fun down the road.

Resolves #402


planet/cli/orders.py
```diff
@@ -63,8 +63,9 @@ def orders(ctx, base_url):
 async def list(ctx, state, limit, pretty):
     '''List orders'''
     async with orders_client(ctx) as cl:
-        orders = await cl.list_orders(state=state, limit=limit, as_json=True)
+        orders = await cl.list_orders(state=state, limit=limit)
 
+    orders = [order.to_dict() for order in orders]
```
sgillies, line 68: @jreiberkyle let's keep it so only the outermost layer of our system, the CLI commands, are concerned about JSON. On the inside, let's use instances of the classes in models.py: Object, ObjectComponent, etc. My reasoning is based on an analogy to Ned Batchelder's ["Unicode Sandwich"](https://nedbatchelder.com/text/unipain.html).

> As we saw with Fact of Life #1, the data coming into and going out of your program must be bytes. But you don’t need to deal with bytes on the inside of your program. The best strategy is to decode incoming bytes as soon as possible, producing unicode. You use unicode throughout your program, and then when outputting data, encode it to bytes as late as possible.
>
> This creates a Unicode sandwich: bytes on the outside, Unicode on the inside.

Also, methods that return one type of thing or another type based on an keyword argument are harder to maintain and refactor in the long run. We'll feel good about a single return type in 6 months or a year, I predict.

@tschaub I'm cc'ing you on this since you're getting back into Python programming. I hope the STAC ecosystem agrees with me about JSON :smile:

planet/cli/orders.py
```diff
@@ -228,4 +229,4 @@ async def create(ctx, name, ids, bundle, item_type, email, cloudconfig, clip,
     async with orders_client(ctx) as cl:
         order = await cl.create_order(request)
 
-    echo_json(order.json, pretty)
+    echo_json(order.to_dict(), pretty)
```
sgillies, line 232: If we're encoding to JSON as late as possible this can just be an ordinary function and not a property that we might cache or whatever.

planet/clients/orders.py
```diff
@@ -20,11 +21,13 @@
...
 
+import arrow
+
 from .. import exceptions
 from ..constants import PLANET_BASE_URL
 from ..http import Session
-from ..models import Order, Orders, Request, Response, StreamingBody
-
+from ..models import (Order, OrderComponent, Orders, Request, Response,
+                      StreamingBody)
```
sgillies, line 30: yapf did this.

planet/clients/orders.py
```diff
@@ -98,21 +98,34 @@ def _stats_url(self):
...
+                expiration = arrow.get(expiration).datetime
+            results.append(
+                OrderComponent(delivery=res.get("delivery", None),
+                               name=res.get("name", None),
+                               location=res.get("location", None),
+                               expires_at=expiration))
+
+        return Order(id=contents["id"],
+                     state=contents["state"],
+                     results=results)
```
sgillies, line 118: @jreiberkyle here is the heart of this PR. Previously, the Order class constructor took a blob of JSON from the API response and did some computation to come up with an order description object. This coupled the API response directly to our information model. I have a hypothesis that using a less tightly coupled Order data class and shifting the responsibility for translation to the OrdersClient will give us more freedom and fun down the road.

planet/clients/orders.py
```diff
@@ -98,21 +98,34 @@ def _stats_url(self):
...
-    ) -> Response:
+    def _order_factory(self, contents: dict) -> Order:
+        """Convert API response contents to an instance of Order."""
+        links = contents.get("_links", [])
+        results = []
+
+        for res in links.get("results", []):
+            expiration = res.get("expires_at", None)
+            if expiration:
+                expiration = arrow.get(expiration).datetime
```
sgillies, line 109: Python's datetime module doesn't parse the API's expiration dates, so I'm bringing in the Arrow project.

planet/clients/orders.py
```diff
@@ -326,24 +326,23 @@ async def download_order(
             planet.exceptions.APIException: On API error.
         """
         order = await self.get_order(order_id)
-        locations = order.locations
+        locations = [component.location for component in order.results]
```
sgillies, line 329: Consequence of adding a new OrderComponent class (alluded to in https://developers.planet.com/docs/orders/reference/#operation/getOrder).

planet/clients/orders.py
```diff
@@ -401,23 +399,19 @@ async def poll(
...
-    ) -> typing.Union[typing.List[Order], dict]:
+    async def list_orders(self,
+                          state: str = None,
+                          limit: int = None) -> typing.List[Order]:
         """Get all order requests.
 
         Parameters:
             state: Filter orders to given state.
             limit: Limit orders to given limit.
-            as_json: Return orders as a json dict.
```
sgillies, line 420: This method will be much easier to maintain if it only returns a list of Orders.

planet/models.py
```diff
@@ -256,6 +257,28 @@ def _get_random_filename(content_type=None):
...
+    name: str
+    location: str
+    expires_at: datetime
+
+
+@define
+class Order:
+    id: str
+    state: str
+    results: typing.List[OrderComponent]
```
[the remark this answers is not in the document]
sgillies: Order is only a data class now. I think this is a great match for its role in our information model: it's only a snapshot of the state of an order, the truth of which exists entirely in our API server.

planet/models.py
```diff
@@ -340,71 +363,16 @@ def _next_link(self, page):
...
 class Orders(Paged):
     '''Asynchronous iterator over Orders from a paged response describing
     orders.'''
     LINKS_KEY = '_links'
     NEXT_KEY = 'next'
     ITEMS_KEY = 'orders'
 
+    def __init__(self, request, do_request_fcn, order_factory, limit=None):
+        super().__init__(request, do_request_fcn, limit=limit)
+        self._order_factory = order_factory
```
sgillies, line 375: We inject a factory for Orders.

tests/integration/test_orders_api.py
```diff
@@ -166,23 +165,6 @@ async def test_list_orders_limit(order_descriptions, session):
     assert oids == ['oid1']
 
 
-@respx.mock
-@pytest.mark.asyncio
-async def test_list_orders_asjson(order_descriptions, session):
```
sgillies, line 171: No longer needed.

tests/integration/test_orders_cli.py
```diff
@@ -65,7 +65,11 @@ def test_cli_orders_list_basic(invoke, order_descriptions):
 
     result = invoke(['list'])
     assert not result.exception
-    assert [order1, order2, order3] == json.loads(result.output)
```
sgillies, line 68: The decoupling I've done means that CLI output is no longer exactly the same as the API response.

Specifically: 

* Order properties other than the ones specified in the class definition are gone from the output JSON. They were already hidden by the Python API (only id and state were surfaced).
* `_links.results` has been flattened to `results`.

sgillies:
Tests are passing, this is ready for review.

planet/cli/orders.py
jreiberkyle: Huh. This is an interesting proposal and I can understand it, especially the analogy, and I still wonder about the tradeoff between working with JSON internally and working with data structures which are pretty much just wrapped JSON. Does the Python library user prefer the user experience of managing data structures vs managing JSON? I don't know the answer, though my original assumption was that they would prefer JSON. I would love to learn more about this ux decision in the wild. I will check out the python libraries associated with aws and gcloud - any others I could check?

> Also, methods that return one type of thing or another type based on an keyword argument are harder to maintain and refactor in the long run. We'll feel good about a single return type in 6 months or a year, I predict.

This I full-heartedly agree with 😄

planet/cli/orders.py
jreiberkyle: The property `json` could be a function that calls e.g. `to_dict()`, so it seems the change would be based on whether it improves ux, right?

planet/clients/orders.py
jreiberkyle: This really points to the need for us to standardize on a formatter. I will elevate the priority of this (see #395). I wish we didn't have a lot of formatting-only changes in a PR as it makes it tougher to wade through, but I am also open to doing the wading. Is there any chance you ran yapf right at the end and can base this PR on the previous commit?

jreiberkyle (commented):
First off, thank you for this work. There are some considerable gems in here and I am appreciating the conversation this initiates. As for my review, I am finding it difficult to focus on the large number of actual proposed changes in light of the equally large number of formatting changes. To help me get traction, I will focus on commenting on the high level changes I see proposed here:

1. Moving managing Order over to the OrdersClient does make sense from a separation of concerns point of view.
2. I would further advocate for managing (paged)Orders to the OrdersClient as well, and possibly spending some time on design of a `pager` that, similar to the `waiter`, can be abstracted across all clients (as aws does)
3. I am curious why you are advocating for having `models` manage the Order dataclass. Could OrdersClient do the same just as well? While it does seem nice to have one 'go to' for a data structure (aka `models` here), it seems to tightly couple OrdersClient to models, as evidenced by the `orders` module importing from `models` and the `models` tests import from the `orders` module.
4. Regarding the proposal to use JSON as IO in the CLI and data structures as IO for the Python library, I think this is an excellent discussion point and may be evidence that we could be well served by going through the Python library public interface, similar to what we've done for the CLI, and coming to a consensus.

planet/models.py
```diff
@@ -265,6 +266,29 @@ def _get_random_filename(content_type=None):
     return name
 
 
+@define
+class OrderComponent:
+    delivery: str
+    expires: datetime
```
sgillies, line 272: @jreiberkyle the API response has `expires_at`, but that's an idiosyncracy. I think users will be less surprised to see `created` since that's a standard metadata term: https://www.dublincore.org/specifications/dublin-core/dcmi-terms/.

planet/models.py
```diff
@@ -265,6 +266,29 @@ def _get_random_filename(content_type=None):
...
+
+
+@define
+class Order:
+    created: datetime
+    error_hints: typing.List[str]
+    id: str
+    last_message: str
+    last_modified: datetime
+    request: dict
```
sgillies, line 284: `request` primarily exists to contain the user-provided order and delivery criteria. In this implementation it will also contain the entire order description sent back by the API.

planet/clients/orders.py
```diff
@@ -94,6 +97,45 @@ def _stats_url(self):
...
+        else:
+            last_modified = None
+
+        return Order(id=contents["id"],
+                     state=contents["state"],
+                     results=results,
+                     created=created,
+                     last_modified=last_modified,
+                     last_message=contents["last_message"],
+                     error_hints=contents["error_hints"],
```
sgillies, line 136: New attributes that fully cover the API-provided and dynamic attributes.

planet/clients/orders.py
```diff
@@ -94,6 +97,45 @@ def _stats_url(self):
...
+                     request=contents)
```
sgillies, line 137: `request` is the user-provided order and delivery criteria. Essential, the data that we post to the API when we create an order.

planet/cli/orders.py
sgillies: Right. We're only going to call `Order.to_dict()` once in a CLI command, just before we print the JSON-encoded result, and I think that'll very likely be true in user code.

Also I think `json.dumps(order.to_dict())` makes more sense that `json.dumps(order.json)`. The second one reads like we're JSON-encoding twice.

sgillies:
@jreiberkyle I rebased on the newly formatted v2 branch and fleshed out the Order class with more API-generated attributes.

I wouldn't be opposed to moving Order, OrderComponent, and Orders into planet.clients.orders. Could we do that in another PR?

planet/models.py
```diff
@@ -276,6 +286,31 @@ class OrderComponent:
...
+            provided.
+        last_message (str): Some info on the current order state. API
+            provided.
+        last_modified (str): The UTC date this Order request was last
+            modified. API provided.
+        request (dict): Product and delivery criteria. User provided.
+        results (list of OrderComponent): The canonical links to this
+            Order's results. API provided.
+        state (str): The current state of this Order request. API
+            provided.
```
sgillies, line 311: All the attribute descriptions come straight from https://developers.planet.com/docs/orders/reference/#operation/getOrder.

PR #405, opened 2022-03-04
add yapf linting with one modifier

jreiberkyle:
Add yapf linting. yapf is installed as a part of the tests sub-package. Running yapf as a linter is included in nox automated testing (which is also what is run in github actions). yapf is directed to ignore all hidden directories.

Modify pep8 style with `split_all_top_level_comma_separated_values=True`.

See #406 to see the effect of removing this modifier.

### DIFF
--- a/planet/models.py
+++ b/planet/models.py
@@ -48,19 +48,19 @@ class Request():
     :raises RequestException: When provided `body_type` is not a subclass of
         :py:class:`planet.api.models.Body`
     '''
+
     def __init__(self, url, params=None, data=None, json=None, method='GET'):
         if data or json:
             headers = {'Content-Type': 'application/json'}
         else:
             headers = None
 
-        self.http_request = httpx.Request(
-            method,
-            url,
-            params=params,
-            data=data,
-            json=json,
-            headers=headers)
+        self.http_request = httpx.Request(method,
+                                          url,
+                                          params=params,
+                                          data=data,
+                                          json=json,
+                                          headers=headers)
 
     @property
     def url(self):
@@ -84,6 +84,7 @@ class Response():
     :param http_response: Response that was received from the server
     :type http_response: :py:Class:`requests.models.Response`
     '''
+
     def __init__(self, request, http_response):
         self.request = request
         self.http_response = http_response
@@ -118,6 +119,7 @@ class StreamingBody():
     :param response: Response that was received from the server
     :type response: :py:Class:`requests.models.Response`
         '''
+
     def __init__(self, response):
         self.response = response.http_response
         self.url = response.request.url
@@ -132,9 +134,8 @@ class StreamingBody():
         :returns: name of this resource
         :rtype: str
         '''
-        name = (_get_filename_from_headers(self.response.headers) or
-                _get_filename_from_url(self.url) or
-                _get_random_filename(
+        name = (_get_filename_from_headers(self.response.headers)
+                or _get_filename_from_url(self.url) or _get_random_filename(
                     self.response.headers.get('content-type')))
         return name
 
@@ -181,7 +182,9 @@ class StreamingBody():
             True.
         :type progress_bar: boolean, optional
         '''
+
         class _LOG():
+
             def __init__(self, total, unit, filename, disable):
                 self.total = total
                 self.unit = unit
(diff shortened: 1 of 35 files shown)


planet/models.py
```diff
@@ -132,9 +134,8 @@ def name(self):
         :returns: name of this resource
         :rtype: str
         '''
-        name = (_get_filename_from_headers(self.response.headers) or
-                _get_filename_from_url(self.url) or
-                _get_random_filename(
+        name = (_get_filename_from_headers(self.response.headers)
+                or _get_filename_from_url(self.url) or _get_random_filename(
```
jreiberkyle, line 138: I'm not sure this is an improvement from a readability point of view.

planet/specs.py
```diff
@@ -17,11 +17,19 @@
...
+    'band_math',
+    'clip',
+    'composite',
+    'coregister',
+    'file_format',
+    'reproject',
+    'tile',
+    'toar',
+    'harmonize'
+]
```
jreiberkyle, line 32: This change is because of the modifier. It goes away when the modifier is removed. I'm ambivalent to the change here.

setup.py
```diff
@@ -64,26 +58,27 @@
...
-          'lint': lint_requires,
-          'docs': doc_requires,
-          'dev': test_requires + lint_requires + doc_requires},
-      entry_points={
-          'console_scripts': [
-              'planet=planet.cli.cli:main',
-          ],
-        },
-      )
+    ],
```
jreiberkyle, line 61: This is the section that prompted the use of the modifier. Without it, these classifiers are clumped together one or two to a line. It's just difficult to understand in that format. [This](https://github.com/planetlabs/planet-client-python/pull/406/files#r819820037) is what happens without the modifier.

tests/conftest.py
```diff
@@ -87,54 +89,30 @@ def geom_geojson():
...
-      }
+        "type":
+        "Polygon",
+        "coordinates":
+        [[[37.791595458984375, 14.84923123791421],
+          [37.90214538574219, 14.84923123791421],
+          [37.90214538574219, 14.945448293647944],
+          [37.791595458984375, 14.945448293647944],
+          [37.791595458984375, 14.84923123791421]]]
+    }  # yapf: disable
```
jreiberkyle, line 100: Yapf kind mangles this so I just disable it here.

tests/integration/test_orders_cli.py
```diff
@@ -343,58 +341,76 @@ def test_cli_orders_create_basic_item_type_invalid(invoke):
...
-        ])
+        '--name',
+        'test',
+        '--id',
+        '',
+        '--bundle',
+        'analytic',
+        '--item-type',
+        'invalid'
+    ])
```
jreiberkyle, line 360: What do we do with cli options that are given as a list, where [OPTION NAME] and [OPTION VALUE] are separate entries? yapf wants to put them all on different lines. This seems to be a step back in readability, but I am meh to okay about letting yapf have it's way just to minimize any special cases we have in the code.

tests/integration/test_orders_cli.py
```diff
@@ -320,12 +317,13 @@ def test_cli_orders_create_basic_success(
         '--id', id_string,
         '--bundle', 'analytic',
         '--item-type', 'PSOrthoTile'
-        ])
+        ])  # yapf: disable
```
jreiberkyle, line 320: What do we do with cli options that are given as a list, where [OPTION NAME] and [OPTION VALUE] are separate entries? yapf wants to put them all on different lines. See what happens when it gets its way [here](https://github.com/planetlabs/planet-client-python/pull/405/files#r819824573). I had decided to just go along with yapf but accidentally left this (and another, below) disable here. It's a good discussion point so I'll leave it for now, but my inclination would be to remove it.

tests/unit/test_models.py
```diff
@@ -87,62 +87,70 @@ def test_StreamingBody_name():
...
+    'content-type': 'image/tiff',
+    'content-length': '57350256',
+    'content-disposition': 'attachment; filename="open_california.tif"'
+}
+
+
+@pytest.mark.parametrize('headers,expected',
+                         [(OPEN_CALIFORNIA_HEADERS, 'open_california.tif'),
+                          (NO_NAME_HEADERS, None),
+                          ({}, None)])  # yapf: disable
```
jreiberkyle, line 113: I think it helps comprehension here to have each 'test' on its own line. yapf would like to join the last two lines. I couldn't find a way around it, so just disabled it.

tests/unit/test_models.py
```diff
@@ -161,14 +169,9 @@ async def _aiter_bytes():
...
-    p1 = {'links': {'next': 'blah'},
-          'items': [1, 2]}
-    p2 = {'links': {},
-          'items': [3, 4]}
-    responses = [
-        mock_http_response(json=p1),
-        mock_http_response(json=p2)
-    ]
+    p1 = {'links': {'next': 'blah'}, 'items': [1, 2]}
+    p2 = {'links': {}, 'items': [3, 4]}
```
jreiberkyle, line 173: For readability, I like the format before this change. However, I don't like it enough to create a special case by disabling yapf.

tests/conftest.py
sgillies: Yeah, there's nothing to do about readability of GeoJSON arrays :rofl:

tests/integration/test_orders_cli.py
sgillies: @jreiberkyle we could switch to `'--item-type=invalid'` if we had to, that's a standard and click supports it.

sgillies:
For me, adopting a formatting tool was a challenge. I had my own style that was pretty much consistent with PEP 8. And black had different ideas. But I found that learning to read the tool's style was a small price to pay for not thinking about line breaks any more. Yapf is more configurable than black, so maybe we can yet find some config that feels comfortable.

tests/integration/test_orders_cli.py
jreiberkyle: @sgillies ohh, great suggestion! I'll do that.

jreiberkyle:
@sgillies I fully agree. Despite all of my comments highlighting issues, I am pretty happy with the result of yapf with this modifier. If you are happy as well, give it a LGTM and I'll merge this (after the CLI commands change), thus closing the book (for now) on the linting issue.

tests/unit/test_models.py
sgillies: FWIW, black would put them each on their own line if the list's text was wider than  some value. See https://black.readthedocs.io/en/stable/the_black_code_style/current_style.html#how-black-wraps-lines.

sgillies (approved):

tests/unit/test_models.py
jreiberkyle: Well, I'm with black on this one. 😆

PR #406, opened 2022-03-04
revert yapf linting to pure pep8

jreiberkyle:
Removes `split_all_top_level_comma_separated_values=True` modifier.

This is prepared for discussion around whether to use the `split_all_top_level_comma_separated_values=True` modifier.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -37,9 +37,7 @@ test_requires = [
 lint_requires = ['flake8', 'yapf']
 
 doc_requires = [
-    'mkdocs==1.1',
-    'mkdocs-click==0.4.0',
-    'mkdocs-material',
+    'mkdocs==1.1', 'mkdocs-click==0.4.0', 'mkdocs-material',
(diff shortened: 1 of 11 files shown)


setup.py
```diff
@@ -49,14 +47,11 @@
...
-        'Environment :: Console',
+        'Development Status :: 2 - Pre-Alpha', 'Environment :: Console',
         'Intended Audience :: Developers',
         'License :: OSI Approved :: Apache Software License',
-        'Operating System :: OS Independent',
-        'Programming Language :: Python',
-        'Topic :: Scientific/Engineering',
-        'Topic :: Software Development',
+        'Operating System :: OS Independent', 'Programming Language :: Python',
+        'Topic :: Scientific/Engineering', 'Topic :: Software Development',
```
jreiberkyle, line 54: This is the section that prompted the use of the modifier. This does't seem very readable.

jreiberkyle:
Modifier accepted, won't go this route.

PR #409, opened 2022-03-09
Fix CLI docs build, add docs build test to CI

jreiberkyle:
Fix broken CLI docs build and add docs build test to CI

**NOTE:** adds new CI and default Nox check


jreiberkyle:
@sean fyi

PR #418, opened 2022-03-22
Verbosity handling1 363

mkshah605:
Changing verbosity handling to incorporate strings as options rather than v's. Added a test for this new functionality.


jreiberkyle (changes requested):
From a high level, 41 changed files is too many files for a PR. There is probably some sort of git untangling that needs to take place. Let's check this out together in our meeting tomorrow.

PR #429, opened 2022-04-20
Report cancelled order

### LINKED ISSUE #364: Update cancel order to report the details of the cancelled order
Currently, when an order is cancelled, the CLI reports 'Cancelled' and the python API returns nothing. However, when an order is cancelled, the Orders API [returns the order details](https://developers.planet.com/docs/orders/reference/#operation/cancelOrder). To align with the api, change these both to return the details of the cancelled order. 

Also, update the SDK and CLI documentation accordingly.

CLI documentation:
```
async def cancel(ctx, order_id, pretty):
    """Cancel order.

    This command outputs the cancelled order details, optionally
    pretty-printed.
    """
```

Blocked by #362

kevinlacaille:
- Report back the details of the cancelled order in a `dict` format, as per our [API docs](https://developers.planet.com/docs/orders/reference/#operation/cancelOrder). 
- Updated documentation in the CLI and the SDK.

Closes #364

### DIFF
--- a/planet/cli/orders.py
+++ b/planet/cli/orders.py
@@ -96,9 +96,9 @@ async def get(ctx, order_id, pretty):
 async def cancel(ctx, order_id):
     '''Cancel order by order ID.'''
     async with orders_client(ctx) as cl:
-        await cl.cancel_order(str(order_id))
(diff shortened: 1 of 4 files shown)


jreiberkyle:
LGTM!

Issue #380, opened 2022-01-27
Update list order interface, change limit behavior


jreiberkyle:
- Update the CLI list order interface to match the interface given below
- Change the behavior of the limit option (in the API as well as the CLI). The limit option should to default to 100, and when user sets it to 0 it should not limit at all.
- Change the output to output each order individually in a sequence. If --pretty is specified, pretty-print each order json blob

```
Usage: planet orders list [OPTIONS]

List orders

This command outputs a sequence of the returned order descriptions.
If --pretty is specified, each order description is pretty-printed.

Options:
  --state TEXT     Filter orders to given state.
  --limit INTEGER  Maximum number of results to return. Default is 100. A
                   value of 0 means no maximum.
  --pretty         Format JSON output.
  --help           Show this message and exit.
```

kevinlacaille:
I think `--state TEXT` introduces ambiguity, as it's not immediately obvious that the options are and doesn't tell them where to find the options. Although it's not as clean, I propose we keep it as a `click.Choice()`, which will indicate the options to the user (i.e., `queued|running|failed|success|partial|cancelled`). Thoughts @jreiberkyle @cholmes @sgillies ?

kevinlacaille:
I propose we change `--state TEXT` to `--state` + a list of all choices available to the user given by `click.Choice()`. i.e., `--state [queued|running|failed|success|partial|cancelled]`.

kevinlacaille:
I think we can get away with only updating the behaviour of the limit option in API (instead of both API and CLI), as the CLI calls the API.

CLI code (`planet.cli.orders.list()`)
```
async with orders_client(ctx) as cl:
    orders = await cl.list_orders(state=state, limit=limit)
    orders_list = [o async for o in orders]
```

As the CLI calls `cl.list_orders()`, I believe we'd only need to influence the behaviour of the limit option in said function (`planet.clients.orders.OrdersClient.list_orders()`)

jreiberkyle:
I agree with using Click.choice for the option in the cli. The api can (and should) also check to ensure that the provided argument is one of the list of valid choices. That list should be an environment variable in the client module and be used to populate click.choice.

Issue #325, opened 2021-10-06
The locations property in models.py breaks when order fails


kevinlacaille:
On branch v2-polling-output-310, in models.py, `locations()` returns a TypeError when trying to unpack all download locations when `self.results` is a NoneType, i.e., when an order fails. Specifically, this occurs at line 380 in the following return statement:
```
return list(r[self.LOCATION_KEY] for r in self.results)
```

Below is the code used for the request, poll, and download the order, taken from docs/guide.md:
```
import asyncio
from planet import Session
from planet import OrdersClient

# Build request
request = {
  "name": "test_order",
  "products": [
    {
      "item_ids": [
        "20170614_113217_3163208_RapidEye-5"
      ],
      "item_type": "REOrthoTile",
      "product_bundle": "analytic"
    }
  ],
  "tools": [
    {
      "toar": {
        "scale_factor": 10000
      }
    },
    {
      "reproject": {
        "projection": "WSG84",
        "kernel": "cubic"
      }
    },
    {
      "tile": {
        "tile_size": 1232,
        "origin_x": -180,
        "origin_y": -90,
        "pixel_size": 2.7056277056e-05,
        "name_template": "C1232_30_30_{tilex:04d}_{tiley:04d}"
      }
    }
  ]
}    

# Polling & Downloading
from planet import reporting

async def create_poll_and_download():
    async with Session() as sess:
        cl = OrdersClient(sess)
        with reporting.StateBar(state='creating') as bar:
            # create order
            order = await cl.create_order(request)
            bar.update(state='created', order_id=order.id)

            # poll
            await cl.poll(order.id, report=bar.update)

        # download
        await cl.download_order(order.id)

asyncio.run(create_poll_and_download())
```

Below is the traceback for the aforementioned error:

```
TypeError                                 Traceback (most recent call last)
<ipython-input-12-401800b0f64e> in <module>
     13         await cl.download_order(order.id)
     14
---> 15 asyncio.run(create_poll_and_download())

/usr/local/Cellar/python@3.9/3.9.6/Frameworks/Python.framework/Versions/3.9/lib/python3.9/asyncio/runners.py in run(main, debug)
     42         if debug is not None:
     43             loop.set_debug(debug)
---> 44         return loop.run_until_complete(main)
     45     finally:
     46         try:

[...]

PR #451, opened 2022-04-27
v2: manage and return an order as json instead of an object

### LINKED ISSUE #434: change orders python api response from Order object to json description of the order
Change orders python api response from Order object to json description of the order

This affects:
- list orders (remove as_json option)
- get order
- cancel order

jreiberkyle:
This PR implements the following:

- moves `Orders` class from `models` to `orders` modules
- removes `Order` class
- manages and returns orders as json instead of instance of `Order`

Closes #434
Blocks #380 and impacts #325


sgillies (approved):
Woo, less code!

PR #456, opened 2022-04-27
Try geojson instead of shapely

sgillies:
Without shapely, we don't need GEOS, and therefore don't need a C library build chain, or docker.

### DIFF
--- a/planet/geojson.py
+++ b/planet/geojson.py
@@ -11,10 +11,12 @@
 # WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
 # License for the specific language governing permissions and limitations under
 # the License.
+
 """Functionality for interacting with GeoJSON."""
+
 import logging
 
-import shapely.geometry as sgeom
+import geojson as gj
 
 LOGGER = logging.getLogger(__name__)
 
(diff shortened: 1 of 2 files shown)


planet/geojson.py
```diff
@@ -107,15 +109,18 @@ def validate_geom(data: dict):
         GeoJSONException: If data is not a valid GeoJSON geometry.
     """
     if 'type' not in data:
-        raise GeoJSONException('Missing \'type\' key.')
-    elif 'coordinates' not in data:
-        raise GeoJSONException('Missing \'coordinates\' key.')
+        raise GeoJSONException("Missing 'type' key.")
```
sgillies, line 112: No need for escaping if we use double quotes.

sgillies:
@jreiberkyle let's merge this and continue without shapely until we have a real need for shapely. Use of the geojson module is good enough as far as the tests are concerned. And without shapely, there's no need for the GEOS library, which looks to be the motivation for the Dockerfile and build system proposed in another PR.

Or... just remove planet/geojson.py? Its functions aren't used in the client or CLI.

sgillies:
I'm going to merge this 4/28 unless there are objections before then. It'll make contributing easier and we can always bring shapely back if we need it in the future.

jreiberkyle:
I am +1 to removing the shapely dependency if it is causing issues. Thank you for looking into that. I am surprised to hear the geojson module is not used - I believe I remember it being used to validate inputs upon creating an order request. The shapely piece is used to validate the actual GeoJSON geometry, something we discussed moving from a step that is done prior to request submission to after receiving a failed request. The benefit being that it can give the user more info about why it failed. I am comfortable with removing that minor benefit from the codebase if shapely is causing issues. The other part of the validation the geojson module does - validating geojson schema - seems valuable as it would reduce the number of pings to the api that are destined to fail. There is a ticket around geojson in which, I believe, we discuss this in more detail.

sgillies:
Actually, VS Code let me down and missed a reference to as_geom in planet/cli/orders.py. So validate_geom is being used.

```
$ grep -R as_geom planet
planet/cli/orders.py:    geo = planet.geojson.as_geom(json_value)
planet/geojson.py:def as_geom(data: dict) -> dict:
planet/geojson.py:    geom = as_geom(data)
planet/geojson.py:            ret = as_geom(data['geometry'])
planet/geojson.py:            ret = as_geom(features[0])
```

PR #458, opened 2022-04-28
fix doc build

jreiberkyle:
adjust doc dep pins to fix doc build


jreiberkyle:
@kevinlacaille @sgillies @mkshah605 FYI 🥳

Issue #453, opened 2022-04-27
Update readme for Orders CLI MVP milestone


cholmes:
So I think we need a few updates to [the README](https://github.com/planetlabs/planet-client-python/blob/v2/README.md) for v2, in order to achieve the Orders CLI MVP milestone.

I don't think we need a very advanced 'install' story, it needs to be good enough for a motivated person who is fairly comfortable with the command line and ok to check out the repo. I _think_ at the least it needs to say `pip install .` instead of `pip install planet`. Though with #408 and a cut release we may be able to do `pip install planet2`? I could see putting `pip install .` in 'contributing' under a 'build from source' type of directions, but I think we need to at least link from there in the readme.

Other thing the readme needs / will need:

* [x] Link to the v1 branch, after #368 and #367 
* [x] Explain the situation of v1 vs v2, especially the state of stability / completeness of v2.
* [x] Update to the quickstart list - we should have it say orders, data and subscriptions, and perhaps link to the milestones with the 'unimplemented', and then mention that we plan in future versions to support analytics, basemaps and tasking.
* [ ] Quickstart should show a CLI example, not a python API example (perhaps move that down to a directory / to another file and then link to 'python API examples', as we want to focus on the CLI.
* [x] Link to the 'docs' directory, instead of just saying `docs` and making people click through the source code.

Issue #368, opened 2022-01-20
Rename "master" branch to "v1"


sgillies:
This issue goes with #367. I don't think we want any more feature work or pull requests on the branch that is currently named master. @jreiberkyle yes? no?

cholmes:
+1 here.

jreiberkyle:
this sounds great!

sgillies:
Done!

Issue #367, opened 2022-01-20
Rename "v2" branch to "main"


sgillies:
@jreiberkyle we're not planning to merge v2 back to master, correct? What would you think about renaming it? Or at least making it the default branch?

jreiberkyle:
I'm for making it the default branch. For one, it will allow us to [automatically close issues from PRs](https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue). Additionally, it is the active development branch. Further, I think this can be done without much fuss.

As for renaming the branch, I am also for renaming it to "main" once @sarasafavi gives the go-ahead from a project/customer/communication management standpoint.

sarasafavi:
+1 for v2 becoming default branch, woot

If we're also moving from _master_ to _main_ then let's split that move into a separate issue, so we can make sure we consider documentation changes it would require, etc.

JuLeeAtPlanet:
On the up side, making main the default branch may correct the version number in [the V2 doc builds](https://planet-sdk-for-python.readthedocs.io/en/latest/).

Whereas the V1.4.9 docs on [developers.planet.com](https://developers.planet.com/docs/pythonclient/), do use an iframe and pull in the github.io V1.4.9:

`<iframe src="https://planetlabs.github.io/planet-client-python/index.html"`

Not sure how github.io websites are generated, but this may be an issue to investigate.

cholmes:
+1 here. Is there any reason to not do this soon?

sgillies:
As far as I can tell, this is the current doc situation:

The v1 docs are built using GitHub Pages from the gh-pages branch and then are slurped into the devcenter site. They won't be affected by a v2 -> main rename.

It looks like the [v2 docs](https://planet-sdk-for-python.readthedocs.io/en/latest/) are built on Read the Docs from config in mkdocs.yml. I'm almost certain that we'll need to get into our RTD account and turn on builds for "main" and make it the primary build after switching over.

sgillies:
We don't have integration between GitHub and RTD set up, so there are no automated doc builds. Last one was 6 months ago and I guess it was manually triggered.

I didn't find any references to the v2 branch in our code or docs outside of a v2_earlyaccess.md file. I'll make the edits there and rename now.

sgillies:
v2_earlyaccess.md updated in fded5d3. Next: renaming using the project settings console.

sgillies:
Done!

[...]

PR #460, opened 2022-04-28
readme updates for #453

cholmes:
This does most of the changes needed for #453

* [x] Link to the v1 branch, after #368 and #367 
* [x] Explain the situation of v1 vs v2, especially the state of stability / completeness of v2.
* [x] Update to the quickstart list - we should have it say orders, data and subscriptions, and perhaps link to the milestones with the 'unimplemented', and then mention that we plan in future versions to support analytics, basemaps and tasking.
* [x] Link to the 'docs' directory, instead of just saying `docs` and making people click through the source code.

The one it doesn't is 

* [ ] Quickstart should show a CLI example, not a python API example (perhaps move that down to a directory / to another file and then link to 'python API examples', as we want to focus on the CLI.

I can probably take that on, but probably not till next week, so just issuing this PR for now to hopefully get merged in soon.

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,11 +4,31 @@
 
 [Planet](https://planet.com) Software Development Kit (SDK) for Python.
 
+## Versions and Stability
+
+The default branch (main) of this repo is for the [Planet SDK for 
+Python](https://github.com/planetlabs/planet-client-python/projects/2),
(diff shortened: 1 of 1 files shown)


jreiberkyle (commented):
looks great!

PR #461, opened 2022-04-28
test package install with nox and in CI, fix breaking install

### LINKED ISSUE #454: change nox to do production install
`pip install .` is throwing un error that is not captured by ci. change nox from `pip install -e .` to `pip install .`

jreiberkyle:
This moves the data dir into the `planet` dir and includes the data dir in the package build. It also updates nox to test the install of the package and includes the test in the CI. Removes unnecessary `--upgrade` flag in nox session.install, as it is unnecessary when using the latest nox package.

Closes #454


jreiberkyle:
@sgillies FYI. @cholmes this should fix your problem with `python -m pip install .`

cholmes:
Sweet! Will try it out.

PR #462, opened 2022-04-29
data api quick search, create search, stats - 289

jreiberkyle:
Implements data client quick_search, create_search, and stats.

Part of #289


jreiberkyle:
@sgillies this seems like an adequately large chunk of code to warrant a PR. Just a part of the data api, but a start!

planet/clients/data.py
```diff
@@ -0,0 +1,221 @@
...
+
+class Items(Paged):
+    '''Asynchronous iterator over items from a paged response.'''
+    LINKS_KEY = '_links'
+    NEXT_KEY = '_next'
+    ITEMS_KEY = 'features'
+
+
+class DataClient():
+    """High-level asynchronous access to Planet's data API.
```
sgillies, line 40: @jreiberkyle wouldn't it be more correct to say "low level"? We're returning raw API JSON responses right now and that's about as low as we can get. High level, to me, implies Search, Stats, &c classes.

planet/clients/data.py
```diff
@@ -0,0 +1,221 @@
...
+        '''Submit a request and get response.
+
+        Parameters:
+            request: request to submit
+        '''
+        return await self._session.request(request)
+
+    async def quick_search(self,
+                           item_types: typing.List[str],
+                           search_filter: dict,
```
sgillies, line 89:
```suggestion
                           filter: dict,
```

I think "search" goes without saying?

sgillies: Nevermind! filter conflicts with the name of the built-in function. search_filter is fine.

tests/integration/test_data_api.py
```diff
@@ -0,0 +1,264 @@
...
+    mock_resp1 = httpx.Response(HTTPStatus.OK, json=page1_response)
+    respx.post(quick_search_url).return_value = mock_resp1
+
+    # if the sort parameter is not used correctly, the client will not send
+    # the request to the mocked endpoint and this test will fail
+    cl = DataClient(session, base_url=TEST_URL)
+    items = await cl.quick_search(['PSScene'], search_filter, sort=sort)
+
+    # run through the iterator to actually initiate the call
+    [i async for i in items]
```
sgillies, line 126: @jreiberkyle this makes me wonder about the API. I think it would be better if quick_search() failed right away if the sort parameter is wrong. Failing fast is almost always better. Waiting to fail until we iterate over results could be surprising to users who call quick_search, pass the result to another function, and then iterate there after doing other work.

sgillies (approved):
Got some comments for you @jreiberkyle .

planet/clients/data.py
```diff
@@ -0,0 +1,221 @@
...
+        self._session = session
+
+        self._base_url = base_url or BASE_URL
+        if self._base_url.endswith('/'):
+            self._base_url = self._base_url[:-1]
+
+    def _searches_url(self):
+        return f'{self._base_url}{SEARCHES_PATH}'
+
+    def _stats_url(self):
```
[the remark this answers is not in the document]
sgillies: cc @mkshah605 @kevinlacaille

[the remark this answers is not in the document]
jreiberkyle: I agree with the rule of 3. In the case of `_searches_url`,  case, there will be more search API calls using this function when the client is complete. This is only a subset of the client.  I believe we will get to the rule of 3 before the client is done. For `_stats_url`, I do not believe we will get to the rule of 3 so I will refactor.

tests/integration/test_data_api.py
jreiberkyle: Agreed that is an issue with generators. This usage paradigm and the tradeoffs would be well served by an intentional discussion with @cholmes . I will make note of it. For now, with CLI being the deliverable, let's move forward with the API as it was planned.

planet/clients/data.py
jreiberkyle: yep, fully agree

PR #464, opened 2022-04-29
Update list order interface, change limit behaviour

kevinlacaille:
Closing #380.

### DIFF
--- a/planet/cli/orders.py
+++ b/planet/cli/orders.py
@@ -57,14 +57,23 @@ def orders(ctx, base_url):
               help='Filter orders to given state.',
               type=click.Choice(planet.clients.orders.ORDER_STATE_SEQUENCE,
                                 case_sensitive=False))
-@click.option('-l',
-              '--limit',
+@click.option('--limit',
               help='Filter orders to given limit.',
               default=None,
               type=int)
 @pretty
 async def list(ctx, state, limit, pretty):
-    '''List orders'''
+    '''List orders
+
+    This command outputs a sequence of the returned order descriptions.
+    If --pretty is specified, each order description is pretty-printed.
+    '''
+    # Set default limit to 100
+    if limit is None:
+        limit = 100
+    # If limit is set to 0, do not limit results
+    elif limit == 0:
+        limit = None
     async with orders_client(ctx) as cl:
         orders = await cl.list_orders(state=state, limit=limit)
         orders_list = [o async for o in orders]
--- a/planet/cli/orders.py
+++ b/planet/cli/orders.py
@@ -59,7 +59,7 @@ def orders(ctx, base_url):
                                 case_sensitive=False))
 @click.option('--limit',
               help='Maximum number of results to return. Default is 100. A '
-                   'value of 0 means no maximum.',
+                   'value of 0 means no limit.',
               default=None,
               type=int)
 @pretty
--- a/planet/cli/orders.py
+++ b/planet/cli/orders.py
@@ -71,10 +71,7 @@ async def list(ctx, state, limit, pretty):
     async with orders_client(ctx) as cl:
         orders = await cl.list_orders(state=state, limit=limit)
         orders_list = [o async for o in orders]
-
-    print(limit)
-    print(len(orders_list))
-    # echo_json(orders_list, pretty)
+    echo_json(orders_list, pretty)
 
 
 @orders.command()
--- a/planet/cli/orders.py
+++ b/planet/cli/orders.py
@@ -68,17 +68,13 @@ async def list(ctx, state, limit, pretty):
     This command outputs a sequence of the returned order descriptions.
     If --pretty is specified, each order description is pretty-printed.
     '''
-    # Set default limit to 100
-    if limit is None:
-        limit = 100
-    # If limit is set to 0, do not limit results
-    elif limit == 0:
-        limit = None
     async with orders_client(ctx) as cl:
         orders = await cl.list_orders(state=state, limit=limit)
         orders_list = [o async for o in orders]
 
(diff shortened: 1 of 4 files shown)


kevinlacaille:
To test the new functionality, we need to still update:
- `tests.integration.test_orders_cli.test_cli_orders_list_limit()`
- `tests.integration.test_orders_api.test_list_orders_limit()` 

New functionality:
- No limit given -> only list up to 100 orders
- Limit set to 0 -> list all orders (limitless)

planet/cli/orders.py
```diff
@@ -57,18 +57,21 @@ def orders(ctx, base_url):
               help='Filter orders to given state.',
               type=click.Choice(planet.clients.orders.ORDER_STATE_SEQUENCE,
                                 case_sensitive=False))
-@click.option('-l',
-              '--limit',
-              help='Filter orders to given limit.',
+@click.option('--limit',
+              help='Maximum number of results to return. Default is 100. A '
+                   'value of 0 means no maximum.',
```
sgillies, line 62:
```suggestion
                   'value of 0 means no limit.',
```

planet/cli/orders.py
```diff
@@ -57,18 +57,21 @@ def orders(ctx, base_url):
...
               default=None,
               type=int)
 @pretty
 async def list(ctx, state, limit, pretty):
-    '''List orders'''
+    '''List orders
+
+    This command outputs a sequence of the returned order descriptions.
+    If --pretty is specified, each order description is pretty-printed.
```
sgillies, line 70: The text in line 70 could go into the help string for the --pretty option?

sgillies, line 70:
```suggestion
    This command prints a sequence of the returned order descriptions, optionally pretty-printed.
```

planet/clients/orders.py
```diff
@@ -422,6 +422,13 @@ async def list_orders(self, state: str = None, limit: int = None):
         """
         url = self._orders_url()
 
+        # Set default limit to 100
+        if limit is None:
+            limit = 100
+        # Set no limit
+        elif limit == 0:
+            limit = None
```
sgillies, line 430: If we change the function definition to `list_orders(self, state: str = None, limit: int = 100)` then we can delete lines 425-427 and avoid the limit/None switcheroo.

tests/integration/test_orders_api.py
```diff
@@ -149,26 +149,36 @@ async def test_list_orders_state_invalid_state(session):
...
 @respx.mock
 @pytest.mark.asyncio
-async def test_list_orders_limit(order_descriptions, session):
+@pytest.mark.parametrize("limit,limitted_list_length", [(None, 100), (0, 102),
+                                                        (1, 1)])
+async def test_list_orders_limit(order_descriptions,
+                                 session,
+                                 limit,
+                                 limitted_list_length):
     nono_page_url = TEST_ORDERS_URL + '?page_marker=OhNoNo'
```
sgillies, line 158: @kevinlacaille I found the problem! The "next" link evaluates to `True` on all the result pages, so we have infinitely many pages and the client never breaks out of the paging loop.

```suggestion
    nono_page_url = None
```

sgillies: If you really want to see what's going on, try replacing some of your test code with this:

```python
    cl = OrdersClient(session, base_url=TEST_URL)

    # import pdb; pdb.set_trace()
    orders = await cl.list_orders(limit=limit)

    async def iter_orders(orders):
        i = 1
        # import pdb; pdb.set_trace()
        async for page in orders._get_pages():
            for item in page["orders"]:
                yield item
                i += 1
                print(f"item {i}")
                if (orders.limit and i > orders.limit) or i > 200:
                    print(f"Reached limit {i}")
                    break
            else:
                continue
            break

    assert len([o async for o in iter_orders(orders)]) == limitted_list_length
```

planet/cli/orders.py
kevinlacaille: Yes, good call!

planet/clients/orders.py
kevinlacaille: Jeez ya, that's way better 🥇

tests/integration/test_orders_api.py
```diff
@@ -149,26 +149,36 @@ async def test_list_orders_state_invalid_state(session):
 
 @respx.mock
 @pytest.mark.asyncio
-async def test_list_orders_limit(order_descriptions, session):
-    nono_page_url = TEST_ORDERS_URL + '?page_marker=OhNoNo'
+@pytest.mark.parametrize("limit,limitted_list_length", [(None, 100), (0, 102),
```
sgillies, line 152: Nit: limited instead of limitted.

sgillies (approved):
Great work @kevinlacaille ! I made two suggestions.

planet/cli/orders.py
```diff
@@ -57,18 +57,21 @@ def orders(ctx, base_url):
               help='Filter orders to given state.',
               type=click.Choice(planet.clients.orders.ORDER_STATE_SEQUENCE,
                                 case_sensitive=False))
-@click.option('-l',
-              '--limit',
-              help='Filter orders to given limit.',
+@click.option('--limit',
+              help='Maximum number of results to return. Default is 100. A '
+              'value of 0 means no limit.',
               default=None,
```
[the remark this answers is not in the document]
kevinlacaille: Oh whoopsie, nice find!

tests/integration/test_orders_api.py
kevinlacaille: IDEs need to also have spell check lol - good find, thanks!

PR #465, opened 2022-04-29
Always use installed code, not editable installs, for nox test session

sgillies:
Testing installed code is the best check since users will be running installed code. Coverage sort of requires an editable install, but we can restrict ourselves to a single Python version for coverage since we don't have different code branches for different versions.

This is a follow up to #461.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -45,14 +45,14 @@ Python and required development packages.
 Run Nox with the default sessions (same checks as CI):
 
 ```console
-    $ nox
+nox
 ```
 
 If no changes have been made to the Nox environment since it was last run,
 speed up the run by reusing the environment:
 
 ```console
-    $ nox -r
+nox -r
 ```
 
 The configuration for Nox is given in `noxfile.py`. See the Nox link above for
@@ -118,7 +118,7 @@ Command-line arguments can be passed to pytest within Nox. For example, to only
 run the tests on a certain file, use:
 
 ```console
-    $ nox -- [file]
+nox -- [file]
 ```
 
 By default, Nox runs tests on all supported Python versions along with other
@@ -127,7 +127,15 @@ CI checks. However, Nox can run a test on a single Python version.
 To run tests on python 3.7:
 
 ```console
-    $ nox -s test-3.7
+nox -s test-3.7
+```
+
+## Code coverage
+
+To measure code coverage and see a report:
+
+```console
+nox -s coverage
 ```
 
 ## Linting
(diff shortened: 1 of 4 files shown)


noxfile.py
```diff
@@ -3,40 +3,43 @@
 nox.options.stop_on_first_error = True
 nox.options.reuse_existing_virtualenvs = False
 
-nox.options.sessions = ['test', 'lint', 'docs', 'install']
+nox.options.sessions = ['lint', 'test', 'coverage', 'docs']
```
sgillies, line 6: Linting first, since that's the most likely to fail in my experience.

setup.cfg
```diff
@@ -6,11 +6,6 @@ license_file: LICENSE
 
 [tool:pytest]
 addopts =
-    --cov=planet
-    --cov=tests
-    --cov-report=term-missing
-    --cov-report=xml
-    --cov-fail-under 98
```
sgillies, line 13: Removing these since coverage is now its own nox session.

noxfile.py
```diff
@@ -3,40 +3,43 @@
...
+                '--no-summary',
+                '--no-cov',
+                '--ignore',
+                'examples/')
+    session.run('coverage', 'report')
 
-    options = session.posargs
-    if '-k' in options:
-        options.append('--no-cov')
-    session.run('pytest', '--ignore', 'examples/', '-v', *options)
```
jreiberkyle, line 19: The options are needed here to allow just running a test on one file or one file's test, aka `nox -s test-3.7 -- tests/unit/test_specs.py`. Please reinstate them.

noxfile.py
```diff
@@ -3,40 +3,43 @@
...
-@nox.session
-def install(session):
-    # just make sure it can be installed
-    session.install(".")
-
-
 @nox.session
 def docs_test(session):
     session.install("-e", ".[docs]")
-
```
jreiberkyle, line 39: Is there a reason for these proposed changes beyond personal preference in formatting?

CONTRIBUTING.md
```diff
@@ -26,13 +26,13 @@ This repository uses two primary tools for development:
 Install Nox in your local dev environment:
 
 ```console
-    $ pip install nox
```
jreiberkyle, line 29: This proposed change takes the CONTRIBUTING.md doc out of alignment with the other docs. Please propose a format change at a higher level so it can be addressed across the documentation.

jreiberkyle (changes requested):
This is an excellent approach. I am on-board with the code test changes. Please remove the formatting changes (feel free to propose them as a separate effort) and reinstate the options for the testing session.

noxfile.py
sgillies: Sometimes an empty line makes code easier to read, but in this case it made it harder for me to read.

PR #469, opened 2022-05-02
Refactored quiet and verbosity tests

kevinlacaille:
Simplify and clarify planet --quiet and --verbosity tests.

### DIFF
--- a/tests/unit/test_cli_main.py
+++ b/tests/unit/test_cli_main.py
@@ -31,6 +31,7 @@ def debug_input():
 
 @pytest.mark.parametrize("option,quiet", [("", False), ("--quiet", True)])
 def test_cli_orders_quiet(option, quiet):
+    """Check that --quiet is passed in context to subcommands."""
 
     # dummy command so we can invoke cli
     @click.command()
@@ -50,6 +51,7 @@ def test_cli_orders_quiet(option, quiet):
                           ("--verbosity=debug", logging.DEBUG)])
 @mock.patch('planet.cli.cli.logging.basicConfig')
 def test_cli_info_verbosity(mock_config, verbosity, log_level):
+    """Check that main command configures logging with the proper level."""
 
     def configtest(stream, level, format):
         assert level == log_level
@@ -68,6 +70,7 @@ def test_cli_info_verbosity(mock_config, verbosity, log_level):
 
 
 def test_cli_invalid_verbosity():
+    """Get a BadParameter error for invalid --verbosity."""
 
     # dummy command so we can invoke cli
     @click.command()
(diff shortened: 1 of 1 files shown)


tests/unit/test_cli_main.py
```diff
@@ -28,28 +29,32 @@ def debug_input():
     return ['debug', ' debug ', 'debu', 45]
 
 
-def test_cli_orders_quiet():
+@pytest.mark.parametrize("option,quiet", [("", False), ("--quiet", True)])
+def test_cli_orders_quiet(option, quiet):
```
sgillies, line 33:
```suggestion
def test_cli_orders_quiet(option, quiet):
    """Check that --quiet is passed in context to subcommands."""
```

tests/unit/test_cli_main.py
```diff
@@ -28,28 +29,32 @@ def debug_input():
...
-    for quiet_input in invalid_quiet_inputs:
-        invalid_result = runner.invoke(cli.main, args=[quiet_input, 'orders'])
-        assert invalid_result.exception
-        assert invalid_result.exit_code != 0
+@pytest.mark.parametrize("verbosity,log_level",
+                         [("", logging.WARNING),
+                          ("--verbosity=info", logging.INFO),
+                          ("--verbosity=debug", logging.DEBUG)])
+@mock.patch('planet.cli.cli.logging.basicConfig')
+def test_cli_info_verbosity(mock_config, verbosity, log_level):
```
sgillies, line 52:
```suggestion
def test_cli_info_verbosity(mock_config, verbosity, log_level):
    """Check that main command configures logging with the proper level."""
```

tests/unit/test_cli_main.py
```diff
@@ -58,27 +63,21 @@ def test():
...
 
-    def configtest(stream, level, format):
-        nonlocal log_level
-        log_level = level
+    runner = CliRunner()
+    runner.invoke(cli.main, args=[verbosity, 'test'], catch_exceptions=False)
 
-    monkeypatch.setattr(cli.logging, 'basicConfig', configtest)
+
+def test_cli_invalid_verbosity():
```
sgillies, line 70:
```suggestion
def test_cli_invalid_verbosity():
    """Get a BadParameter error for invalid --verbosity."""
```

sgillies:
The key things here:

- Asserting in "helper" functions or commands can improve clarity and eliminate the need for Python tricks. It requires `catch_exceptions=False` for CliRunner.invoke, otherwise the AssertionErrors aren't seen by pytest.
- When patching, we need to patch the module that our unit under test is defined in.
- Avoid shoehorning multiple tests into one function because a) that reduces readability and clarity, and b) we lose the ability to run the tests in random order. Tests that run in a fixed order are a great place for bugs to hide.

PR #471, opened 2022-05-02
Change version to 2.0dev

sgillies:
This will help prevent confusion in environments between pre-releases and the final release.


jreiberkyle (approved):
Thank you for doing this!

Issue #432, opened 2022-04-21
Add checksum match to download order


jreiberkyle:
Add checksum match to download order.

Update the CLI interface to match the following, and change the API interface accordingly:

```
Usage: planet orders download [OPTIONS] ORDER_ID

  Download order

  If --checksum is provided, the associated checksums given in the manifest
  are compared against the downloaded files to verify that they match.

Options:
  --checksum [MD5|SHA1]  Verify that checksums match.
  --directory DIRECTORY  Root directory for file download.
  --overwrite            Overwrite files if they already exist.
  --help                 Show this message and exit.
```

Implement checksum match verification for download order for the CLI and API. In the API, add a new parameter, `checksum` to `orders.download_order()` and implement the functionality to check the checksum of the downloaded files against the downloaded manifest. Add a new exception that is thrown upon match validation failure. In the CLI, pass on the `checksum` variable and handle the thrown exception.

mkshah605:
@jreiberkyle the manifest json files have MD5 and SHA256 keys. Do we still want SHA1 to be an option here, or should we just keep it MD5?

jreiberkyle:
You can just update the option to SHA256 instead of SHA1

PR #474, opened 2022-05-03
Checksum match orders #432

mkshah605:


planet/clients/orders.py
```diff
@@ -313,6 +338,26 @@ async def download_order(self,
...
+            # Save each filename and respective hash in a dict as a key-value pair
+            with open(manifest_json, 'rb') as manifest:
+                manifest_data = json.load(manifest)
+                file_key_pairs = {}
+                for json_entry in manifest_data['files']:
+                    file_name = json_entry['path'].split('/')[-1]
+                    origin_hash = json_entry['digests'][checksum.lower()]
+                    file_key_pairs[file_name] = origin_hash
+            # For each file (not including manifest json file), calculate hash on its contents
+            filenames_loop = [x for x in filenames if not x.endswith('manifest.json')]
```
kevinlacaille, line 354: For your test, you currently only provide a `manifest.json`, so this will return an empty string and you'll never run `self.calculate_checksum()`.

tests/integration/test_orders_api.py
```diff
@@ -536,6 +536,37 @@ async def test_download_asset_md(tmpdir, session):
     assert Path(filename).name == 'metadata.json'
 
 
+@respx.mock
+@pytest.mark.asyncio
+async def test_checksum_success(tmpdir, order_description, oid, session):
+    # Mock an HTTP response for download
+    order_description['state'] = 'success'
```
kevinlacaille, line 543: If you mock 2 responses (i.e., mock an order with 2 files), then you can test the functionality of your checksum function.

```
order_description['state'] = 'success'
dl_url1 = TEST_DOWNLOAD_URL + '/1?token=IAmAToken'
dl_url2 = TEST_DOWNLOAD_URL + '/2?token=IAmAnotherToken'
order_description['_links']['results'] = [{
    'location': dl_url1
}, {
    'location': dl_url2
}]

get_url = f'{TEST_ORDERS_URL}/{oid}'
mock_resp = httpx.Response(HTTPStatus.OK, json=order_description)
respx.get(get_url).return_value = mock_resp

mock_resp1 = httpx.Response(HTTPStatus.OK,
                            json={"name":"","files":[{"path":Path(tmpdir, "manifest.json").name,
                                                        "digests":{"md5":"anOriginMd5",
                                                                    "sha256":"anOriginSha256"}}]},
                            headers={
                                'Content-Type':
                                'application/json',
                                'Content-Disposition':
                                'attachment; filename="manifest.json"'
                            })
respx.get(dl_url1).return_value = mock_resp1

mock_resp2 = httpx.Response(HTTPStatus.OK,
                            json={"name":"","files":[{"path":Path(tmpdir, "manifest.json").name,
                                                        "digests":{"md5":"aReturnMd5",
                                                                    "sha256":"aReturnSha256"}}]},
                            headers={
                                'Content-Type':
                                'application/json',
                                'Content-Disposition':
                                'attachment; filename="someotherfile.json"'
                            })
respx.get(dl_url2).return_value = mock_resp2
```

planet/clients/orders.py
```diff
@@ -313,6 +338,26 @@ async def download_order(self,
...
+            # filenames_loop = ' '.join([x for x in filenames if not x.endswith('manifest.json')])
+            for filename in filenames_loop:
+                downloaded_file_name = filename.split('/')[-1]
+                self.calculate_checksum(origin_hash= file_key_pairs[downloaded_file_name],
```
kevinlacaille, line 358: On line 352 you store the file name and MD5 key for `manifest.json` in `file_key_pairs`. And on line 357 `downloaded_file_name` is the name of the a file that's not `manifest.json`, so as of right now `file_key_pairs[downloaded_file_name]` won't work. However, a simple solution would be to change from `file_key_pairs[downloaded_file_name]` to `file_key_pairs['manifest.json']`, so you're really just trying to pull the key from the `manifest.json`.

planet/clients/orders.py
```diff
@@ -313,6 +338,26 @@ async def download_order(self,
                                       progress_bar=progress_bar)
             for location in locations
         ]
+        if checksum:
```
kevinlacaille, line 341: This only works if you're expecting `checksum` to not be `None`, but your code looks for `checksum` to either be `MD5` or `SHA256`, so you can replace this with `checksum == 'md5' or 'sha256'` for it to work.

planet/clients/orders.py
```diff
@@ -271,18 +273,41 @@ async def download_asset(self,
                              progress_bar=progress_bar)
         return dl_path
 
+    @staticmethod # should this be an async function?
+    def calculate_checksum(origin_hash: str,
+                        filename: str,
+                        checksum: str):
+        """Calculate checksum and validate that it passes."""
+        if checksum == 'MD5':
```
kevinlacaille, line 281: I think you should only compare `checksum` to strings without a care of its case (i.e., uppercase or lowercase). Here this fails because on line 351 you compare to `checksum.lower()`, which is great! (i.e., `origin_hash = json_entry['digests'][checksum.lower()]`). Let's either do `checksum.lower()` everywhere, or maybe ever more robust, we can assign `checksum = checksum.lower()` at the beginning and now you know `md5` and `sha256` will always be lowercase. Or, another option would be to compare without caring about cases, which would be like `checksum.casefold() == 'MD5'.casefold()`.

tests/integration/test_orders_api.py
```diff
@@ -536,6 +536,37 @@ async def test_download_asset_md(tmpdir, session):
...
+                                    'application/json',
+                                    'Content-Disposition':
+                                    'attachment; filename="manifest.json"'
+                                })
+    respx.get(dl_url1).return_value = mock_resp1
+    cl = OrdersClient(session, base_url=TEST_URL)
+    filenames = await cl.download_order(oid, directory=str(tmpdir), checksum=True)
+    import pdb; pdb.set_trace()
+    # Check that the checksum passes and doesn't raise an exception
+    assert json.load(open(filenames[0])) == {'md5': '22e054aad3e6cb33d31a4c919ccadcb2'}
```
kevinlacaille, line 568: Instead, you should see if a ClientError was raised. A way of doing this is:
```
with pytest.raises(exceptions.ClientError):
    cl.download_order(oid, directory=str(tmpdir), checksum=True)
```

Then you can get rid of the assert message and this will take care of all of the comparisons.

However, this code will only pass if the exception is raised, so this would be for a failing test.

planet/clients/orders.py
```diff
@@ -271,18 +273,41 @@ async def download_asset(self,
...
+            hash_type = hashlib.md5
+        elif checksum == 'SHA256':
+            hash_type = hashlib.sha256
+        # Calculate returned hash
+        with open(filename, 'rb') as file_to_check:
+            json_data = file_to_check.read()
+            returned_hash = hash_type(json_data).hexdigest()
+            # Compare original hashkey in dict with calculated
+            if origin_hash != returned_hash:
```
kevinlacaille, line 290: Here you're comparing the `origin_hash`, as it's written in the JSON file, to the `returned_hash`, which has been put through `hexdigest()`, so it's no comparing apples to apples. 

For instance, if you put `origin_hash = 'anOriginMd5'` through `hash_type(json_data).hexdigest()` you'll get `'39f5b5db96c26ca61e13b280c8b98b6e'`, so you're comparing `'anOriginMd5 == '39f5b5db96c26ca61e13b280c8b98b6e'`, which will return false.

So, you'll need to compare `hash_type(origin_json_data).hexdigest() == hash_type(otherfile_json_data).hexdigest()`

kevinlacaille: Here's an idea:

In `download_order()` you can replace a bunch of your code with just a few lines, as now you're going to pass along a lot of the heavy lifting to `calculate_checksum()`. Try:

```
if checksum == 'md5':
    # Checksum Implementation
    # Get manifest filepath
    manifest_json = ' '.join([x for x in filenames if x.endswith('manifest.json')])
    with open(manifest_json, 'rb') as manifest:
        manifest_data = manifest.read()

    # For each file (not including manifest json file), calculate hash on its contents
    filenames_loop = [x for x in filenames if not x.endswith('manifest.json')]
    # filenames_loop = ' '.join([x for x in filenames if not x.endswith('manifest.json')])
    for filename in filenames_loop:
        self.calculate_checksum(origin_json_data= manifest_data,
                            filename= filename,
                            checksum= checksum)
```

And in `calculate_checksum()` try this:
```
def calculate_checksum(origin_json_data: str,
                    filename: str,
                    checksum: str):
    """Calculate checksum and validate that it passes."""
    if checksum == 'md5':
        hash_type = hashlib.md5
    elif checksum == 'sha256':
        hash_type = hashlib.sha256
    # Calculate returned hash
    with open(filename, 'rb') as file_to_check:
        json_data = file_to_check.read()
        origin_hash = hash_type(origin_json_data).hexdigest()
        returned_hash = hash_type(json_data).hexdigest()
        # Compare original hashkey in dict with calculated
        if origin_hash != returned_hash:
            print('origin_hash: ', origin_hash)
            print('returned_hash: ', returned_hash)
            raise exceptions.ClientError(
                'Checksum failed. File not correctly downloaded.')
        return f'({checksum}) checksum succesful: ({filename})'
```

kevinlacaille: Now you're comparing apples to apples! 🍎 🍎

planet/clients/orders.py
mkshah605: `manifest.json` has several keys inside it for different files- I don't actually want to use the `manifest.json` itself (it has no keys tied to it) which is why it's left out.

planet/clients/orders.py
mkshah605: hmmmm i'm not sure the hexdigest is the issue? I think it's different with strings, but that's needed to read the encoded data in the file so that I can see it as a string and compare it that way. Also, i'm trying to keep the checksum function as simple as possible (not incorporating the part of the function that loads the manifest) to keep things as simple as possible for testing as i was having issues with that previously.

mkshah605:
She's finally ready for review!! @kevinlacaille @sgillies @jreiberkyle would love your feedback. 
The lint is currently failing due to the hashkey, which exceeds the required line length. But when I split the hashkey into multiple lines using '\' it ends up changing the value and the checksum fails. Any advice on how to get around this?

planet/clients/orders.py
```diff
@@ -271,18 +274,72 @@ async def download_asset(self,
                              progress_bar=progress_bar)
         return dl_path
 
+    @staticmethod  # should this be an async function?
```
sgillies, line 277: @mkshah605 really good question! Async doesn't usually help as much with local file reads as it does with network requests (the benefits depend on the system's hardware and OS). And we can only hash one string of bytes at a time asynchronously, so no speed up can be had there. To get a real speed up we would need to read files and hash them in different threads, and there we would have to watch out for memory usage.

planet/clients/orders.py
```diff
@@ -271,18 +274,72 @@ async def download_asset(self,
...
+                           checksum: str):
+        """Calculate checksum and validate that it passes.
+
+        Parameters:
+            manifest_data: The contents of the opened manifest.json file.
+            filenames: List of downloaded assets.
+            checksum: The type of checksum hash- 'MD5' or 'SHA256'.
+
+        Returns:
+            Message confirming checksums were succesful.
```
sgillies, line 289: @mkshah605 in my experience, having functions return message strings isn't very useful. And at line 380 we're not doing anything with the message. It would be fine for this method to `return None` (Python's default if there is no return statement) on success and raise an exception (as you do) on failure.

Also, I feel like the `calculate` in the name of the method implies that we are calculating checksums so that they would be returned and used by other code. That's not the case here, so how about `validate_checksums` instead?

planet/clients/orders.py
```diff
@@ -271,18 +274,72 @@ async def download_asset(self,
...
+                json_data = file_to_check.read()
+                returned_hash = hash_type(json_data).hexdigest()
+                # Compare original hashkey in dict with calculated
+                if origin_hash != returned_hash:
+                    print("origin hash: ", origin_hash)
+                    print("returned hash: ", returned_hash)
+                    raise exceptions.ClientError(
+                        f'Checksum failed. File ({filename}) not correctly \
+                        downloaded.')
+                print(f'({checksum}) checksum succesful: ({filename})')
```
sgillies, line 326: Remember to remove the debugging prints.

planet/clients/orders.py
```diff
@@ -271,18 +274,72 @@ async def download_asset(self,
...
+            checksum = checksum.lower()
+        elif checksum == 'SHA256':
+            hash_type = hashlib.sha256
+            checksum = checksum.lower()
+        file_key_pairs = {}
+        for json_entry in manifest_data['files']:
+            file_name = Path(json_entry['path']).name
+            origin_hash = json_entry['digests'][checksum]
+            file_key_pairs[file_name] = origin_hash
+        # For each downloaded file, retrieve origin hashkey from dict
```
sgillies, line 307: This is an excellent comment and helps explain the code!

planet/clients/orders.py
```diff
@@ -313,6 +369,17 @@ async def download_order(self,
                                       progress_bar=progress_bar)
             for location in locations
         ]
+        if checksum:
+            # Checksum Implementation
+            # Get manifest filepath
+            manifest_json = ' '.join(
+                [x for x in filenames if x.endswith('manifest.json')])
+            # Open manifest file and pass to checksum function
```
sgillies, line 377: Just a little nit: you could delete these comments because the code is already very clear. Comments are good when needed but can be distracting when not needed.

tests/integration/test_orders_api.py
```diff
@@ -536,6 +536,222 @@ async def test_download_asset_md(tmpdir, session):
...
 
 
+@respx.mock
+@pytest.mark.asyncio
+async def test_checksum_success(tmpdir, order_description, oid, session):
+    # Mock an HTTP response for download
+    order_description['state'] = 'success'
+    dl_url1 = TEST_DOWNLOAD_URL + '/1?token=IAmAToken'
+    dl_url2 = TEST_DOWNLOAD_URL + '/2?token=IAmAnotherToken'
+    dl_url3 = TEST_DOWNLOAD_URL + '/2?token=IAmYetAnotherToken'
```
sgillies, line 546: The test isn't about tokens, so I find the endings of the URLs a bit misleading. How about this?

```suggestion
    dl_url1 = TEST_DOWNLOAD_URL + '/asset1'
    dl_url2 = TEST_DOWNLOAD_URL + '/assert2'
    dl_url3 = TEST_DOWNLOAD_URL + '/manifest'
```

tests/integration/test_orders_api.py
```diff
@@ -536,6 +536,222 @@ async def test_download_asset_md(tmpdir, session):
...
+                              "8092eb7d64064d4c0c2136d4cdb61eaa",
+                              "sha256":
+                              "0a6874c3e4a75c45828fc3c2ae5f6c238b9f99bfcdcb020332b02e651c5c2873"
+                          }
+                      }]
+        },
+        headers={
+            'Content-Type': 'application/json',
+            'Content-Disposition': 'attachment; filename="mock_manifest.json"'
+        })
```
sgillies, line 617: Excellent!

tests/integration/test_orders_api.py
```diff
@@ -536,6 +536,222 @@ async def test_download_asset_md(tmpdir, session):
...
+                              "md5":
+                              "8092eb7d64064d4c0c2136d4cdb61eaa",
+                              "sha256":
+                              "0a6874c3e4a75c45828fc3c2ae5f6c238b9f99bfcdcb020332b02e651c5c2873"
+                          }
+                      }]
+        },
+        headers={
+            'Content-Type': 'application/json',
+            'Content-Disposition': 'attachment; filename="mock_manifest.json"'
```
sgillies, line 616: How about just manifest.json? I believe this is the standard name for our API and that the client won't encounter anything else.

```suggestion
            'Content-Disposition': 'attachment; filename="manifest.json"'
```

tests/integration/test_orders_api.py
```diff
@@ -536,6 +536,222 @@ async def test_download_asset_md(tmpdir, session):
...
+        })
+    respx.get(dl_url3).return_value = mock_manifest
+    # Test Checksum
+    for checksum in ['MD5', 'SHA256']:
+        cl = OrdersClient(session, base_url=TEST_URL)
+        filenames = await cl.download_order(oid, directory=str(tmpdir))
+        cl.calculate_checksum(manifest_data=mock_manifest.json(),
+                              filenames=filenames,
+                              checksum=checksum)
```
sgillies, line 625: @mkshah605 instead of calling calculate_checksum explicitly, this test should call download_order with the `checksum` keyword argument.

tests/integration/test_orders_api.py
```diff
@@ -536,6 +536,222 @@ async def test_download_asset_md(tmpdir, session):
...
+    respx.get(get_url).return_value = mock_asset
+
+    mock_asset1 = httpx.Response(HTTPStatus.OK,
+                                 json={
+                                     "md5":
+                                     "1975b76cc6904ff84f432be2ef1fea2e",
+                                     "sha256":
+                                     "300ebdfd62614f82ce6ba45b298bf797\
+                                           f1ea595d9214bcf23899f3fd2267c5dd"
+                                 },
```
sgillies, line 565: @mkshah605 so that we don't have too many checksums flying around in the test, what would you think about simplifying the content? `json={"asset": 1}` will suffice.

tests/integration/test_orders_api.py
```diff
@@ -536,6 +536,222 @@ async def test_download_asset_md(tmpdir, session):
...
+    mock_manifest = httpx.Response(
+        HTTPStatus.OK,
+        json={
+            "name":
+            "",
+            "files": [{
+                "path": "asset1.json",
+                "digests": {
+                    "md5":
+                    "9a967dfe809c2eb25d990ad547feadf0",
```
sgillies, line 599: It would also be okay to compute the digest values directly in here. Like this:

```suggestion
                    hashlib.md5(mock_asset1.content).hexdigest(),
```

Then it would automatically stay up to date if we ever changed the mock response JSON.

BTW, I verified that we can use the `content` attribute of a response like that.

```pycon
>>> resp = Response(200, json={"asset": 1})
>>> resp
<Response [200 OK]>
>>> resp.content
b'{"asset": 1}'
```

It's always bytes, so no need to encode it.

sgillies (commented):
@mkshah605 this is very good work! I made some suggestions that will make your tests a bit more clear.

tests/integration/test_orders_api.py
sgillies: I think my suggestion will solve flake8's problem with the long line, too.

tests/integration/test_orders_api.py
sgillies: This could also help make flake8 happy.

planet/clients/orders.py
mkshah605: Makes sense. Thanks Sean!!

planet/clients/orders.py
mkshah605: Agree! Will do.

tests/integration/test_orders_api.py
```diff
@@ -536,6 +537,169 @@ async def test_download_asset_md(tmpdir, session):
     assert Path(filename).name == 'metadata.json'
 
 
+@respx.mock
+@pytest.mark.asyncio
+async def test_checksum_success(tmpdir, order_description, oid, session):
+    # Mock an HTTP response for download
+    order_description['state'] = 'success'
+    dl_url1 = TEST_DOWNLOAD_URL + '/asset1'
+    dl_url2 = TEST_DOWNLOAD_URL + '/assert2'
```
sgillies, line 546:
```suggestion
    dl_url2 = TEST_DOWNLOAD_URL + '/asset2'
```

tests/integration/test_orders_api.py
```diff
@@ -536,6 +537,169 @@ async def test_download_asset_md(tmpdir, session):
...
+    order_description['_links']['results'] = [{
+        'location': dl_url1
+    }, {
+        'location': dl_url2
+    }, {
+        'location': dl_url3
+    }]
+    get_url = f'{TEST_ORDERS_URL}/{oid}'
+    mock_asset = httpx.Response(HTTPStatus.OK, json=order_description)
+    respx.get(get_url).return_value = mock_asset
```
sgillies, line 557: @mkshah605 I'm suggested a name change since the order description isn't an asset and also isn't a mock, it's a real response object.

```suggestion
    get_order_response = httpx.Response(HTTPStatus.OK, json=order_description)
    respx.get(get_url).return_value = get_order_response
```

tests/integration/test_orders_api.py
```diff
@@ -536,6 +537,169 @@ async def test_download_asset_md(tmpdir, session):
...
+                                     "md5": "mock_key1_md5",
+                                     "sha256": "mock_key1_sha256"
+                                 },
+                                 headers={
+                                     'Content-Type':
+                                     'application/json',
+                                     'Content-Disposition':
+                                     'attachment; filename="asset1.json"'
+                                 })
+    respx.get(dl_url1).return_value = mock_asset1
```
sgillies, line 570: Or... (an idea I had while running on Sunday) it could be helpful to simulate a TIFF asset instead of a JSON asset.

```suggestion
    # The first asset in the order is a 1 byte binary file containing the value "1".
    asset1_content = b"1"
    asset1_response = httpx.Response(HTTPStatus.OK,
                                 content=asset1_content, 
                                 headers={
                                     'Content-Type':
                                     'image/tiff',
                                     'Content-Disposition':
                                     'attachment; filename="asset1.tif"'
                                 })
    respx.get(dl_url1).return_value = asset1_response
```

planet/clients/orders.py
```diff
@@ -271,18 +274,67 @@ async def download_asset(self,
                              progress_bar=progress_bar)
         return dl_path
 
+    @staticmethod
+    def validate_checksum(manifest_data: dict, filenames: list, checksum: str):
```
sgillies, line 278: @mkshah605 after thinking about this a little more, I believe we will be doing ourselves and our users a favor if we remove this static method from the OrdersClient class and turn it into a stand-alone function in the orders module.

The implementation doesn't require a client object, it only requires a manifest document and a directory of downloaded assets. The function might be useful outside the context of downloading an order, for example: a user might want to re-validate their data after copying their downloaded assets to another computer.

tests/integration/test_orders_api.py
mkshah605: This makes sense, especially since the downloaded assets will include both json and TIFF files.

tests/integration/test_orders_api.py
```diff
@@ -536,6 +537,171 @@ async def test_download_asset_md(tmpdir, session):
...
+    # Test Checksum
+    for checksum in ['MD5', 'SHA256']:
+        cl = OrdersClient(session, base_url=TEST_URL)
+        await cl.download_order(oid, directory=str(tmpdir), checksum=checksum)
+
+
+@respx.mock
+@pytest.mark.asyncio
+async def test_checksum_failure(tmpdir, order_description, oid, session):
+    # Note: the hashkeys in the mock manifest below were changed
```
jreiberkyle, line 622: these kinds of notes are so helpful. thank you.

planet/clients/orders.py
jreiberkyle, line 278: @sgillies had a great point but somehow with the passage of time it looks like the original comment is gone. @sgillies did you change your mind on recommending this fcn moves out of `OrdersClient` while staying in the module?

This was my response to that recommendation:
> good point @sgillies and I agree. 
> 
> @mkshah605 if it feels like an easy change, go for it. If you think it will take you a while (>1 hour) to implement, make a ticket for moving the function out of the client, link to that ticket here, and feel free to continue with the PR.

jreiberkyle (changes requested):
Awesome work Mansi! It is so great to see this functionality added to the API. I am fully on-board with approving this when we get cleared up on the functionality for determining the path to the `manifest.json` file. All other suggestions are just possibilities for improvement.

planet/clients/orders.py
```diff
@@ -313,6 +364,15 @@ async def download_order(self,
                                       progress_bar=progress_bar)
             for location in locations
         ]
+        if checksum:
+            # Checksum Implementation
+            manifest_json = ' '.join(
```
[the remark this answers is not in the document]
mkshah605: As far as I know, there should only be one manifest.json file in each order, that has info about all the items in that order

tests/integration/test_orders_api.py
```diff
@@ -536,6 +537,171 @@ async def test_download_asset_md(tmpdir, session):
...
+                                     headers={
+                                         'Content-Type':
+                                         'image/tiff',
+                                         'Content-Disposition':
+                                         'attachment; filename="asset1.tif"'
+                                     })
+    respx.get(dl_url1).return_value = asset1_response
+
+    asset2_response = httpx.Response(HTTPStatus.OK,
+                                     json={
```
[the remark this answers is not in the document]
mkshah605: This is just an example, it could be anything. As an example, the first asset used in this test is a tif file (so not even json). This works because the hashkey is calculated based on the contents of the file, regardless of what it it.

tests/integration/test_orders_api.py
```diff
@@ -536,6 +537,171 @@ async def test_download_asset_md(tmpdir, session):
...
+                              },
+                              headers={
+                                  'Content-Type':
+                                  'application/json',
+                                  'Content-Disposition':
+                                  'attachment; filename="manifest.json"'
+                              })
+    respx.get(dl_url3).return_value = manifest
+    # Test Checksum
+    for checksum in ['MD5', 'SHA256']:
```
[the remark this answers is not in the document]
mkshah605: Yes absolutely!!

planet/clients/orders.py
mkshah605: @jreiberkyle I can't seem to find this comment anymore either, but @sgillies  and I talked about it separately- I wasn't really sure what it would involve/if there were other considerations we needed to take into account as a group given our current timeline, but I agree it's a great idea and am happy to implement. Sean made a [new ticket about it here](https://github.com/planetlabs/planet-client-python/issues/521). 

Could (either of you) clarify on what you mean by "out of the Orders Client into the current module"? Where/what file would this function ideally sit in so that it could be more generally available to users?

jreiberkyle: The function `validate_checksum` would still be in `orders.py` but would not be a method of the `OrdersClient` class.

tests/integration/test_orders_api.py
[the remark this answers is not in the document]
jreiberkyle: ok, cool. thanks. would you be willing to change the json to something like `{'foo': 'bar'}` to highlight the fact that the content of the json doesn't matter? That will be a little clearer than having the content be related to checksums.

planet/clients/orders.py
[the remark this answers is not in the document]
jreiberkyle: ok then what about doing something like this vs the `' '.join()` fcn:

```python
manifest_files = [x for x in filenames if x.endswith('manifest.json')])
if len(manifest_files) > 1:
    raise ClientError('error msg')
manifest_json = manifest_files[0]
```

planet/clients/orders.py
sgillies: Right, I moved my comment to #521 and deleted it here so it wouldn't be a distraction.

jreiberkyle: awesome!

planet/clients/orders.py
```diff
@@ -343,6 +343,8 @@ async def download_order(self,
             planet.exceptions.APIError: On API error.
             planet.exceptions.ClientError: If the order is not in a final
                 state.
+            planet.exceptions.ClientError: If more than one manifest file
+                per order.
```
sgillies, line 347: Unless this has ever been observed to happen I think it would be okay to omit this check for now. Getting no manifest or an incomplete manifest feels way more likely to me.

planet/cli/orders.py
```diff
@@ -190,14 +194,17 @@ async def wait(ctx, order_id, delay, max_attempts, state):
               is_flag=True,
               default=False,
               help=('Overwrite files if they already exist.'))
-async def download(ctx, order_id, overwrite, directory):
-    """Download order by order ID."""
+async def download(ctx, order_id, overwrite, directory, checksum):
+    """Download order by order ID. \n
```
jreiberkyle, line 198:
```suggestion
    """Download order by order ID.
    
```

planet/clients/orders.py
mkshah605: Hmm yes that's what I was wondering too. @jreiberkyle thoughts? Should I just revert to what I had before or would you prefer to keep this in?

jreiberkyle: Agreed those two are much more likely. In fact, the entire name of the file is `manifest.json` so instead of searching for a file that ends with `manifest.json` (which is a broader search and could possibly come up with multiple files), restrict the search to a file that is named `manifest.json`. This could be done pretty easily with a check for a match on `Path.name` ([ref](https://docs.python.org/3.7/library/pathlib.html#pathlib.PurePath.name)). And agreed that we would also want to wrap missing or corrupt `manifest.json` files in a `ClientError`

planet/clients/orders.py
```diff
@@ -433,3 +458,48 @@ async def list_orders(self, state: str = None, limit: int = None):
 
         request = self._request(url, 'GET', params=params)
         return Orders(request, self._do_request, limit=limit)
+
+
+def validate_checksum(manifest_data: dict, filenames: list, checksum: str):
```
jreiberkyle, line 463: This looks great! Since the user can call this function directly, as it is not a private function just being used by `download_order` anymore, I would consider usage scenarios and look at adding a bit of defensive logic. For example, I am curious what happens if `checksum` is an invalid entry like `checksum='foobar'`. Additionally, the user could call this function with arbitrary entries in `filenames`. I am curious what happens if one or more of the entries in `filenames` is not contained within `manifest_data`. Also what happens if an entry in `filenames` does not exist? These scenarios may be good candidates for raising an exception. Since this logic is no longer contained within a client, it may make sense to use a `PlanetError` instead of a `ClientError` (and catch and re-raise in `download_order` as a `ClientError`). 

It also may make sense to add tests for these scenarios. The tests can test `validate_checksum` directly.

jreiberkyle (changes requested):

planet/clients/orders.py
jreiberkyle: @mkshah605 My recommendation here is to separate these efforts, go back to `validate_checksum` being a function within the client (I would make it private at this point so only `download_order` uses it) and merge the work for checksum matching in download orders. The work for checksum matching in download_orders is ready to be merged now and is required for the Orders CLI milestone, which is tomorrow (Wed). The effort to elevate `validate_checksum` involves more work and is not high priority.

mkshah605: Thank you @jreiberkyle sounds good! I've included this info for added detail/future reference in [this ticket](https://github.com/planetlabs/planet-client-python/issues/521)

planet/clients/orders.py
```diff
@@ -271,18 +274,69 @@ async def download_asset(self,
                              progress_bar=progress_bar)
         return dl_path
 
+    @staticmethod
+    def __validate_checksum(manifest_data: dict,
```
jreiberkyle, line 278: private - or, more specifically, non-public - functions start with just one underscore, so this would be `_validate_checksum()` ([ref](https://docs.python.org/3.7/tutorial/classes.html#private-variables))

```suggestion
    def _validate_checksum(manifest_data: dict,
```

planet/clients/orders.py
```diff
@@ -313,6 +369,24 @@ async def download_order(self,
...
+            if manifest_count > 1:
+                raise exceptions.ClientError(
+                    'Only 1 manifest.json expected per order.\
+                                             Recieved: {manifest_count}')
+            elif manifest_count is None:
+                raise exceptions.ClientError('No manifest file found.')
+            manifest_json = manifest_files[0]
+            with open(manifest_json, 'rb') as manifest:
+                manifest_data = json.load(manifest)
+                self.__validate_checksum(manifest_data=manifest_data,
```
jreiberkyle, line 387:
```suggestion
                self._validate_checksum(manifest_data=manifest_data,
```

PR #482, opened 2022-05-03
Remove empty () after class definitions

### LINKED ISSUE #481: `class Paged:` instead of `class Paged():` &c
Dropping the empty parentheses when there's no base class is more common Python style and might work better with some IDEs (I don't know for sure, I only use Vim and VS Code).

sgillies:
Resolves #481

Partially this was an exercise in learning how to do more refactoring in VS Code. I used the search tool to find occurrences of `class (\w+)\(\)` (as a regex) and replace them with `class $1`. 7 instances were replaced.

### DIFF
--- a/planet/clients/orders.py
+++ b/planet/clients/orders.py
@@ -45,7 +45,7 @@ class Orders(Paged):
     ITEMS_KEY = 'orders'
 
 
-class OrderStates():
+class OrderStates:
     SEQUENCE = ORDER_STATE_SEQUENCE
 
     @classmethod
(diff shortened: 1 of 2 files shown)


jreiberkyle (approved):
Hah! This a great catch and is a great learning point for me. Thanks!

sgillies:
This will be almost invisible to work in other PRs out there.

sgillies:
Doh! My first search was accidentally restricted to files that were open in the editor. I found more outside. Still on the VS Code learning curve.

Issue #279, opened 2021-04-30
implement retry with exponential backoff for TOO_MANY_REQUESTS


jreiberkyle:
In #276, it was discovered that httpx retry is not for status codes but instead for connection errors. Therefore, retry with [exponential backoff](https://en.wikipedia.org/wiki/Exponential_backoff) for TOO_MANY_REQUESTS needs to be implemented in our codebase.

This will fix #349

PR #483, opened 2022-05-04
download order cli: increase number of poll attempts to a reasonable number

jreiberkyle:
The poll attempts (5) is too low a number for a reasonable polling time. This increases this number to 200, which is the default for the python API.


jreiberkyle:
@sgillies @cholmes FYI

cholmes:
Cool. Does this close https://github.com/planetlabs/planet-client-python/issues/279? Or are you also going to make an PR for the exponential backoff?

jreiberkyle:
this is unrelated so I'll make a separate pr for exponential backoff

PR #490, opened 2022-05-04
Fix errors reported by mypy

### LINKED ISSUE #489: Use mypy to check type hints
We're using type hints in a few places. If we don't check or test them they could be erroneous.

Here are the results of running mypy on our code today:

```
$ mypy planet
planet/reporting.py:17: error: Skipping analyzing "tqdm.asyncio": module is installed, but missing library stubs or py.typed marker
planet/reporting.py:94: error: "None" has no attribute "postfix"
planet/reporting.py:102: error: "None" has no attribute "set_description_str"
planet/reporting.py:104: error: "None" has no attribute "refresh"
planet/geojson.py:18: error: Skipping analyzing "geojson": module is installed, but missing library stubs or py.typed marker
planet/geojson.py:18: note: See https://mypy.readthedocs.io/en/stable/running_mypy.html#missing-imports
planet/models.py:23: error: Skipping analyzing "tqdm.asyncio": module is installed, but missing library stubs or py.typed marker
planet/order_request.py:73: error: Incompatible types in assignment (expression has type "int", target has type "Sequence[Collection[Any]]")
planet/order_request.py:76: error: Incompatible types in assignment (expression has type "Dict[Any, Any]", target has type "Sequence[Collection[Any]]")
planet/order_request.py:79: error: Incompatible types in assignment (expression has type "Dict[Any, Any]", target has type "Sequence[Collection[Any]]")
planet/order_request.py:83: error: Incompatible types in assignment (expression has type "Optional[str]", target has type "Sequence[Collection[Any]]")
planet/order_request.py:117: error: List item 1 has incompatible type "Optional[str]"; expected "str"
planet/auth.py:85: error: Argument 1 to "APIKeyAuth" has incompatible type "Optional[str]"; expected "str"
planet/auth.py:132: error: "Auth" has no attribute "to_dict"
planet/http.py:123: error: Argument "auth" to "AsyncClient" has incompatible type "planet.auth.Auth"; expected "Union[Tuple[Union[str, bytes], Union[str, bytes]], Callable[[Request], Request], httpx._auth.Auth, None]"
planet/clients/data.py:182: error: Incompatible types in assignment (expression has type "bool", target has type "Collection[Any]")
planet/clients/__init__.py:17: error: Type of __all__ must be "Sequence[str]", not "List[Type[object]]"
planet/__init__.py:20: error: Type of __all__ must be "Sequence[str]", not "List[object]"
Found 17 errors in 9 files (checked 21 source files)
```

I'm going to see about adding mypy to our nox configuration and look into some of those errors. Some of them look like easily fixable hinting bugs or nits. In planet/reporting.py it looks like mypy has found a potential bug that our tests don't trigger. Static analysis for the win?

sgillies:
Resolves #489

### DIFF
--- a/planet/__init__.py
+++ b/planet/__init__.py
@@ -18,10 +18,10 @@ from .auth import Auth
 from .clients import DataClient, OrdersClient
 
 __all__ = [
-    "Auth",
-    "DataClient"
-    "order_request",
-    "OrdersClient",
-    "reporting",
-    "Session",
+    'Auth',
+    'DataClient'
+    'OrdersClient',
+    'order_request',
+    'reporting',
+    'Session',
 ]
--- a/planet/__init__.py
+++ b/planet/__init__.py
@@ -17,4 +17,11 @@ from .__version__ import __version__  # NOQA
 from .auth import Auth
 from .clients import DataClient, OrdersClient
 
-__all__ = [Session, OrdersClient, order_request, reporting, Auth, DataClient]
+__all__ = [
+    "Auth",
+    "DataClient"
+    "order_request",
+    "OrdersClient",
+    "reporting",
+    "Session",
+]
--- a/noxfile.py
+++ b/noxfile.py
@@ -3,9 +3,16 @@ import nox
 nox.options.stop_on_first_error = True
 nox.options.reuse_existing_virtualenvs = False
 
-nox.options.sessions = ['lint', 'test', 'coverage', 'docs']
+nox.options.sessions = ['lint', 'analyze', 'test', 'coverage', 'docs']
 
 source_files = ("planet", "examples", "tests", "setup.py", "noxfile.py")
+
+
+@nox.session
+def analyze(session):
+    session.install(".[lint]")
+
+    session.run("mypy", "planet")
 
 
 @nox.session
--- a/planet/clients/__init__.py
+++ b/planet/clients/__init__.py
@@ -15,6 +15,6 @@ from .data import DataClient
 from .orders import OrdersClient
 
 __all__ = [
-    DataClient,
-    OrdersClient,
+    'DataClient',
+    'OrdersClient',
 ]
--- a/setup.py
+++ b/setup.py
@@ -34,7 +34,7 @@ test_requires = [
     'pytest', 'pytest-asyncio==0.16', 'pytest-cov', 'respx==0.16.3'
 ]
 
-lint_requires = ['flake8', 'yapf']
+lint_requires = ['flake8', 'mypy', 'yapf']
 
 doc_requires = [
     'mkdocs==1.3',


planet/clients/data.py
```diff
@@ -175,11 +175,12 @@ async def create_search(self,
         url = self._searches_url()
 
         # TODO: validate item_types
-        request_json = {
+        request_json: typing.Dict[str, typing.Any] = {
             'name': name, 'filter': search_filter, 'item_types': item_types
         }
+        # QUESTION: can we just set the value to True or False above?
```
sgillies, line 181: We have a test that doesn't expect `__daily_email_enabled: False`, but I wonder if the API would handle this?

planet/clients/data.py
```diff
@@ -175,11 +175,12 @@ async def create_search(self,
         url = self._searches_url()
 
         # TODO: validate item_types
-        request_json = {
+        request_json: typing.Dict[str, typing.Any] = {
```
sgillies, line 178: Without this hint, mypy diagnoses request_json's type as `Dict[str, Collection[Any]]` because name, search_filter, and item_types are all collections, and then complains when we try to add a bool item, which is not a collection.

planet/auth.py
```diff
@@ -113,14 +115,18 @@ def from_login(email: str, password: str, base_url: str = None) -> Auth:
...
         pass
 
     @property
     @abc.abstractmethod
     def value(self):
         pass
 
+    @abc.abstractmethod
+    def to_dict(self) -> dict:
+        pass
```
sgillies, line 128: Resolves 

```
planet/auth.py:132: error: "Auth" has no attribute "to_dict"
```

planet/auth.py
```diff
@@ -30,12 +30,14 @@
 BASE_URL = f'{PLANET_BASE_URL}/v0/auth'
 ENV_API_KEY = 'PL_API_KEY'
 
+AuthType = httpx.Auth
```
sgillies, line 33: Resolves 

```
planet/auth.py:85: error: Argument 1 to "APIKeyAuth" has incompatible type "Optional[str]"; expected "str"
planet/http.py:123: error: Argument "auth" to "AsyncClient" has incompatible type "planet.auth.Auth"; expected "Union[Tuple[Union[str, bytes], Union[str, bytes]], Callable[[Request], Request], httpx._auth.Auth, None]"
```

planet/auth.py
```diff
@@ -80,7 +82,7 @@ def from_env(variable_name: str = None) -> Auth:
             variable_name: Alternate environment variable.
         '''
         variable_name = variable_name or ENV_API_KEY
-        api_key = os.getenv(variable_name)
+        api_key = os.getenv(variable_name, '')
```
sgillies, line 85: Resolves

```
planet/auth.py:85: error: Argument 1 to "APIKeyAuth" has incompatible type "Optional[str]"; expected "str"
```

The APIKeyAuth constructor does not take None as an argument (`key: str`) but we might pass it none if there's nothing in the environment.

planet/auth.py
```diff
@@ -93,7 +93,9 @@ def from_env(variable_name: str = None) -> AuthType:
         return auth
 
     @staticmethod
-    def from_login(email: str, password: str, base_url: str = None) -> AuthType:
+    def from_login(email: str,
+                   password: str,
+                   base_url: str = None) -> AuthType:
```
sgillies, line 98: Carried over formatting.

planet/order_request.py
```diff
@@ -67,7 +67,7 @@ def build_request(name: str,
         planet.specs.SpecificationException: If order_type is not a valid
             order type.
     '''
-    details = {'name': name, 'products': products}
+    details: Dict[str, Any] = {'name': name, 'products': products}
```
sgillies, line 70: mypy requires us to be more careful about dict typing.

planet/order_request.py
```diff
@@ -79,8 +79,8 @@ def build_request(name: str,
         details['notifications'] = notifications
 
     if order_type:
-        order_type = specs.validate_order_type(order_type)
-        details['order_type'] = order_type
+        validated_order_type = specs.validate_order_type(order_type)
+        details['order_type'] = validated_order_type
```
sgillies, line 83: In general, mypy discourages reassigning names, and not without reason because it can be a source of bugs.

planet/order_request.py
```diff
@@ -108,18 +108,18 @@ def product(item_ids: List[str],
...
+    validated_product_bundle = specs.validate_bundle(product_bundle)
+    item_type = specs.validate_item_type(item_type, validated_product_bundle)
 
     if fallback_bundle is not None:
-        fallback_bundle = specs.validate_bundle(fallback_bundle)
-        specs.validate_item_type(item_type, fallback_bundle)
-        product_bundle = ','.join([product_bundle, fallback_bundle])
+        validated_fallback_bundle = specs.validate_bundle(fallback_bundle)
+        specs.validate_item_type(item_type, validated_fallback_bundle)
+        validated_product_bundle = ','.join([validated_product_bundle, validated_fallback_bundle])
```
sgillies, line 117: Here, though, mypy's analyzer can't find that anything on the right side will be None, so it doesn't care if we reassign.

planet/reporting.py
```diff
@@ -90,15 +90,19 @@ def update_state(self, state: str):
...
 
         if order_id:
             self.order_id = order_id
-            self.bar.set_description_str(self.desc, refresh=False)
+            if self.bar is not None:
+                self.bar.set_description_str(self.desc, refresh=False)
 
-        self.bar.refresh()
+        if self.bar is not None:
+            self.bar.refresh()
```
sgillies, line 108: mypy noticed some cases where self.bar could be None and we would try to call its methods.

sgillies:
@jreiberkyle @mkshah605 @kevinlacaille this is 6 commits, 9 files. It might look like a long review, but it's only a few lines per file. The TLDR: a few commits fix subtle bugs that our tests were not triggering (static analysis for the win!), the rest of the commits fix improper use of type hints or reconcile our code with mypy's stricter view of typing. I think we should merge this right away if we're going to continue to use type hints in the project. And I think we probably will, users increasingly expect them.

planet/__init__.py
```diff
@@ -15,6 +15,13 @@
...
 
-__all__ = [Session, OrdersClient, order_request, reporting, Auth, DataClient]
+__all__ = [
+    'Auth',
+    'DataClient'
+    'OrdersClient',
+    'order_request',
+    'reporting',
+    'Session',
+]
```
sgillies, line 27: `__all__` requires strings, not objects.

planet/reporting.py
kevinlacaille: That is cool as heck. I should try mypy

planet/clients/data.py
```diff
@@ -175,11 +175,12 @@ async def create_search(self,
...
             'name': name, 'filter': search_filter, 'item_types': item_types
         }
+        # QUESTION: can we just set the value to True or False above?
         if enable_email:
-            request_json['__daily_email_enabled'] = True
+            request_json['__daily_email_enabled'] = enable_email
```
jreiberkyle, line 183: I'm curious about this change. What does it improve?

planet/clients/data.py
jreiberkyle: ahh interesting! good catch. I wonder if we want to make this a custom type, eg `JSONType` and move it over, along with `AuthType` to a new `types` module?

planet/order_request.py
jreiberkyle: Oh! yeah, I can see that. nice catch.

planet/reporting.py
jreiberkyle: awesome catch. this would probably have bubbled up as a bug eventually.

jreiberkyle (approved):
its great to have static typing in the python library and turned out to be much easier than I had expected.

planet/order_request.py
sgillies: I reassign names a lot and it's a hard habit to break :laughing:

planet/clients/data.py
sgillies: I'm addressing both of these questions by setting `__daily_email_enabled: enable_email` in the literal dict expression at line 179.

jreiberkyle:
@sgillies  can you please also add how to use mypy into CONTRIBUTING.md?

sgillies:
Done in ff57b41.

PR #525, opened 2022-05-10
CLI Interface Docs

cholmes:
This ports the CLI Interface documents from internal google docs to public / markdown.


cholmes:
Merging this in as it's pretty minor and won't break any code, and I need it as a reference for issues.

PR #527, opened 2022-05-10
Data api skeleton

### LINKED ISSUE #493: write interface skeleton for data api functions
define data api function interfaces and each function will currently raise `NotImplementedError`

asset functions:
* list_assets
* get_asset
* get_asset_type
* activate_asset
* wait_asset
* download_asset

item functions:
* get_item
* get_item_type
* list_item_types

search functions:
* list_searches
* delete_search
* get_search
* update_search
* run_search

blocks #511 #510 #495 #509 #508 #499

jreiberkyle:
This specifies the data python api interfaces.

Closes #493

### DIFF
--- a/planet/clients/data.py
+++ b/planet/clients/data.py
@@ -26,6 +26,9 @@ STATS_PATH = '/stats'
 
 STATS_INTERVAL = ('hour', 'day', 'week', 'month', 'year')
 
+WAIT_DELAY = 5
+WAIT_MAX_ATTEMPTS = 200
+
 LOGGER = logging.getLogger(__name__)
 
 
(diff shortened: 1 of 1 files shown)


planet/clients/data.py
```diff
@@ -216,3 +307,207 @@ async def get_stats(self,
...
+            planet.exceptions.APIError: On API error.
+            planet.exceptions.ClientError: If asset type identifier is not
+            valid.
+        """
+        # NOTE: this is not an API endpoint
+        # this is getting an asset by name from the dict returned by
+        # list_item_assets()
+        raise NotImplementedError
+
+    async def activate_asset(self, asset: dict):
```
jreiberkyle, line 439: I went back and forth on having the input for asset functions be `asset` (description of the asset) or `[item_id, item_type_id, asset_type_id]`. The latter has the benefit of simpler parameters and not requiring the user to run `get_asset()`.

I ultimately went with `asset` because it will cut down on each function having to call `get_asset()`. One concern with this is that the user may be providing a description of an asset that is no longer in sync with what is on the server (an example being that the user provides a description of an asset that has a download location url, but the asset has since been inactivated so the download url is no longer valid).

planet/clients/data.py
```diff
@@ -185,11 +188,99 @@ async def create_search(self,
...
+    async def list_searches(
+            self,
+            sort: str = 'created desc',
+            search_type: str = 'any') -> typing.AsyncIterator[dict]:
+        """List all saved searches available to the authenticated user.
+
+        NOTE: the term 'saved' is overloaded here. We want to list saved
+        searches that are 'quick' or list saved searches that are 'saved'? Do
+        we want to introduce a new term, 'stored' that encompasses 'saved' and
+        'quick' searches?
```
jreiberkyle, line 219: I just want to highlight this and raise it for consideration of review of the Data API docs (my source for most of the docstrings).

PR #537, opened 2022-05-11
Exponential backoff for retries

### LINKED ISSUE #279: implement retry with exponential backoff for TOO_MANY_REQUESTS

jreiberkyle:
Changes retry backoff from constant to exponential backoff with random jitter.

Introduces `MAX_RETRY_BACKOFF` and `MAX_RETRIES` global vars and removes them from `Session` class variables.

Closes #279

### DIFF
--- a/planet/http.py
+++ b/planet/http.py
@@ -24,8 +24,8 @@ from .auth import Auth, AuthType
 from . import exceptions, models
 from .__version__ import __version__
 
-RETRY_COUNT = 5
-MAX_BACKOFF = 64  # seconds
+MAX_RETRIES = 5
+MAX_RETRY_BACKOFF = 64  # seconds
 
 LOGGER = logging.getLogger(__name__)
 
@@ -137,7 +137,6 @@ class Session(BaseSession):
         self._client.event_hooks['response'] = [
             alog_response, araise_for_status
         ]
-        self.retry_count = RETRY_COUNT
 
     async def __aenter__(self):
         return self
@@ -171,7 +170,7 @@ class Session(BaseSession):
                 resp = await func(*a, **kw)
                 break
             except exceptions.TooManyRequests as e:
-                if num_tries > RETRY_COUNT:
+                if num_tries > MAX_RETRIES:
                     raise e
                 else:
                     LOGGER.debug(f'Try {num_tries}')
@@ -193,7 +192,7 @@ class Session(BaseSession):
         """
         random_number_milliseconds = random.randint(0, 1000) / 1000.0
         calc_wait = 2**num_tries + random_number_milliseconds
-        return min(calc_wait, MAX_BACKOFF)
+        return min(calc_wait, MAX_RETRY_BACKOFF)
 
     async def request(self,
                       request: models.Request,
--- a/planet/http.py
+++ b/planet/http.py
@@ -16,6 +16,7 @@ from __future__ import annotations  # https://stackoverflow.com/a/33533514
 import asyncio
 from http import HTTPStatus
 import logging
+import random
 
 import httpx
 
@@ -24,7 +25,7 @@ from . import exceptions, models
 from .__version__ import __version__
 
 RETRY_COUNT = 5
-RETRY_WAIT_TIME = 1  # seconds
+MAX_BACKOFF = 64  # seconds
 
 LOGGER = logging.getLogger(__name__)
 
@@ -136,7 +137,6 @@ class Session(BaseSession):
         self._client.event_hooks['response'] = [
             alog_response, araise_for_status
         ]
-        self.retry_wait_time = RETRY_WAIT_TIME
         self.retry_count = RETRY_COUNT
 
     async def __aenter__(self):
@@ -151,18 +151,13 @@ class Session(BaseSession):
     async def _retry(self, func, *a, **kw):
         """Run an asynchronous request function with retry.
 
+        Retry uses exponential backoff
+
         Raises:
             planet.exceptions.TooManyRequests: When retry limit is exceeded.
         """
-        # TODO: retry will be provided in httpx v1 [1] with usage [2]
-        # 1. https://github.com/encode/httpcore/pull/221
-        # 2. https://github.com/encode/httpx/blob/
-        # 89fb0cbc69ea07b123dd7b36dc1ed9151c5d398f/docs/async.md#explicit-transport-instances # noqa
         # TODO: if throttling is necessary, check out [1] once v1
         # 1. https://github.com/encode/httpx/issues/984
-        retry_count = self.retry_count
-        wait_time = self.retry_wait_time
-
         num_tries = 0
         while True:
             num_tries += 1
@@ -170,15 +165,29 @@ class Session(BaseSession):
                 resp = await func(*a, **kw)
                 break
             except exceptions.TooManyRequests as e:
-                if num_tries > retry_count:
+                if num_tries > RETRY_COUNT:
                     raise e
                 else:
                     LOGGER.debug(f'Try {num_tries}')
+                    wait_time = self._calc_exponential_wait(num_tries)
                     LOGGER.info(f'Too Many Requests: sleeping {wait_time}s')
-                    # TODO: consider exponential backoff
-                    # https://developers.planet.com/docs/data/api-mechanics/
                     await asyncio.sleep(wait_time)
         return resp
+
+    @staticmethod
+    def _calc_exponential_wait(num_tries):
+        """Calculates exponential wait
+
+        Introduces some randomness to stagger multiple requests and avoid
+        waves. Also thresholds maximum wait value to MAX_BACKOFF.
+
+        Ref:
+        * https://developers.planet.com/docs/data/api-mechanics/
+        * https://cloud.google.com/iot/docs/how-tos/exponential-backoff
+        """
+        random_number_milliseconds = random.randint(0, 1000) / 1000.0
+        calc_wait = 2**num_tries + random_number_milliseconds
+        return min(calc_wait, MAX_BACKOFF)
 
     async def request(self,
                       request: models.Request,
--- a/planet/http.py
+++ b/planet/http.py
@@ -156,6 +156,12 @@ class Session(BaseSession):
         Raises:
             planet.exceptions.TooManyRequests: When retry limit is exceeded.
         """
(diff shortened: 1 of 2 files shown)


planet/http.py
```diff
@@ -151,35 +150,50 @@ async def aclose(self):
...
-        # 89fb0cbc69ea07b123dd7b36dc1ed9151c5d398f/docs/async.md#explicit-transport-instances # noqa
+        # NOTE: if we need something more fancy, consider the following libs:
+        # * https://pypi.org/project/retry/
+        # * https://pypi.org/project/retrying/
+        # TODO: consider increasing the scope of retryable exceptions, ex:
+        # https://github.com/googleapis/python-storage/blob/1dc6d647b86466bf133
+        # fe03ec8d76c541e19c632/google/cloud/storage/retry.py#L23-L30
         # TODO: if throttling is necessary, check out [1] once v1
         # 1. https://github.com/encode/httpx/issues/984
-        retry_count = self.retry_count
```
jreiberkyle, line 163: There doesn't seem to be any benefit from having these be `Session` properties. Just manage them as global vars.

tests/unit/test_http.py
```diff
@@ -102,31 +109,48 @@ async def test_session_stream(mock_request):
...
         assert route.call_count == 2
 
 
 @respx.mock
 @pytest.mark.asyncio
-async def test_session_retry(mock_request):
-    async with http.Session() as ps:
+async def test_session__retry(monkeypatch):
+    """A unit test for the _retry function"""
+    # decorator form of patch seems to conflict with respx.mock
```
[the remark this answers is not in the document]
sgillies: @jreiberkyle it looks to me like this is intended behavior for patch. Under https://docs.python.org/3/library/unittest.mock.html#unittest.mock.patch it says

> If [patch()](https://docs.python.org/3/library/unittest.mock.html#unittest.mock.patch) is used as a decorator and new is omitted, the created mock is passed in as an extra argument to the decorated function.

Someone opened a ticket about this at https://bugs.python.org/issue44052, but they're looking at it from the other side: they don't want to add a new argument when they use `new`.

planet/http.py
```diff
@@ -151,35 +150,50 @@ async def aclose(self):
...
+
+        Introduces some randomness to stagger multiple requests and avoid
+        waves. Also thresholds maximum wait value to MAX_BACKOFF.
+
+        Ref:
+        * https://developers.planet.com/docs/data/api-mechanics/
+        * https://cloud.google.com/iot/docs/how-tos/exponential-backoff
+        """
+        random_number_milliseconds = random.randint(0, 1000) / 1000.0
+        calc_wait = 2**num_tries + random_number_milliseconds
```
kevinlacaille, line 194: This is so smart, I love this! You mentioned you found this idea in some resource, which one was that?

tests/unit/test_http.py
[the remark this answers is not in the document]
sgillies: That sleep gets called with a particular value is kind of an implementation detail, tho? Often we want to avoid testing these because they get in the way of refactoring down the road. The backoff timing logic could be checked by testing _calc_exponential_wait directly, yes?

Here's an idea: your test_func could assert that requests were backing off by recording some timestamps. That's not an implementation detail and sorta mimics what we're trying to do for the server?

planet/http.py
jreiberkyle: Its the resource referenced on line 191 :)

planet/http.py
```diff
@@ -151,35 +150,50 @@ async def aclose(self):
...
+        """Calculates exponential wait
+
+        Introduces some randomness to stagger multiple requests and avoid
+        waves. Also thresholds maximum wait value to MAX_BACKOFF.
+
+        Ref:
+        * https://developers.planet.com/docs/data/api-mechanics/
+        * https://cloud.google.com/iot/docs/how-tos/exponential-backoff
+        """
+        random_number_milliseconds = random.randint(0, 1000) / 1000.0
```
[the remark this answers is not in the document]
jreiberkyle: yah, great point! That absolutely seems like something that would work.

tests/unit/test_http.py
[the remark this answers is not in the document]
sgillies: I didn't observe any conflict with respx.mock. What I saw was that patch works but doesn't pass the mock to the function if the new keyword argument is used. And then pytest mistakes the argument for a fixture and fails.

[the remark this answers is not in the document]
jreiberkyle: Ok, well it seems we are stuck with this way for mocking an async function since we support pre-3.8 and only at 3.8 does it return an AsyncMock by default.

tests/unit/test_http.py
```diff
@@ -102,31 +110,59 @@ async def test_session_stream(mock_request):
...
-async def test_session_retry(mock_request):
-    async with http.Session() as ps:
+async def test_session__retry():
+    """A unit test for the _retry function"""
+
+    async def test_func():
+        # directly trigger the retry logic
+        raise exceptions.TooManyRequests
+
+    with patch('planet.http.Session._calculate_wait') as mock_wait:
```
jreiberkyle, line 139: @sgillies here is where I have moved from mocking `asyncio.wait` to mocking `_calculate_wait` (hopefully a little less into the implementation weeds). `_calculate_wait` isn't async so we don't need to override the default with the `new` parameter in the patch fcn. However, I could still use help figuring out how to move patch from a context manager to a decorator.

planet/http.py
```diff
@@ -136,8 +137,9 @@ async def araise_for_status(*args, **kwargs):
         self._client.event_hooks['response'] = [
             alog_response, araise_for_status
         ]
-        self.retry_wait_time = RETRY_WAIT_TIME
-        self.retry_count = RETRY_COUNT
+
+        self.max_retries = MAX_RETRIES
```
jreiberkyle, line 141: Conversation with @sgillies revealed it may be beneficial to have `max_retries` and `max_retry_backoff` be instance states, so I restored them after briefly removing them.

planet/http.py
jreiberkyle: restored the instance state variables. 👍

planet/http.py
```diff
@@ -151,35 +153,57 @@ async def aclose(self):
...
+    @staticmethod
+    def _calculate_wait(num_tries, max_retry_backoff):
+        """Calculates retry wait
+
+        Base wait period is calculated as a exponential based on the number of
+        tries. Then, a random jitter of up to 1s is added to the base wait
+        to avoid waves of requests in the case of multiple requests. Finally,
+        the wait is thresholded to the maximum retry backoff.
+
+        Because threshold is applied after jitter, calculations that hit
```
jreiberkyle, line 195: This is a little catch that could potentially be an issue if max_retry_backoff is sufficiently small so the threshold gets hit on the first retry. In that case, we will still have 'stampeding requests'

jreiberkyle:
ok @sgillies this is ready for another round of review

tests/unit/test_http.py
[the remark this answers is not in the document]
sgillies: @jreiberkyle here's an example of using unittest.mock.patch and respx.mock together. I'm patching OrderStates in test_checksum_success and printing the value of OrderStates in orders.py to show that it's a MagicMock. The test passes and I see "MagicMock" printed, so apparently unittest.mock.patch and respx.mock are both working.

```diff
diff --git a/planet/clients/orders.py b/planet/clients/orders.py
index 6b8c7c8..bcb906f 100644
--- a/planet/clients/orders.py
+++ b/planet/clients/orders.py
@@ -348,6 +348,7 @@ class OrdersClient:
         """
         order = await self.get_order(order_id)
         order_state = order['state']
+        print(OrderStates)
         if not OrderStates.is_final(order_state):
             raise exceptions.ClientError(
                 'Order cannot be downloaded because the order state '
diff --git a/tests/integration/test_orders_api.py b/tests/integration/test_orders_api.py
index bf8bd16..108dcb9 100644
--- a/tests/integration/test_orders_api.py
+++ b/tests/integration/test_orders_api.py
@@ -19,10 +19,12 @@ import math
 import os
 from pathlib import Path
 from unittest.mock import call, create_autospec
+from unittest import mock
 import hashlib
 
 import httpx
 import pytest
+# from requests import patch
 import respx
 
 from planet import OrdersClient, exceptions, reporting
@@ -551,7 +553,8 @@ async def test_download_asset_md(tmpdir, session):
 @respx.mock
 @pytest.mark.asyncio
 @pytest.mark.parametrize("checksum", [("MD5"), ("SHA256")])
-async def test_checksum_success(tmpdir,
+@mock.patch("planet.clients.orders.OrderStates", autospec=True)
+async def test_checksum_success(mock_states, tmpdir,
                                 order_description,
                                 oid,
                                 session,
```

tests/unit/test_http.py
jreiberkyle: @sgillies I tried to follow the example you gave that worked for you, but still can't get a successful mock, whereas I'm successful with the context manager

Here's what I tried (the commented patch is one of my attempts too):
```python
@respx.mock
@pytest.mark.asyncio
# @patch("planet.http.Session._calculate_wait", autospec=True)
@patch.object(http.Session, '_calculate_wait', autospec=True)
async def test_session__retry(mock_wait):
    """A unit test for the _retry function"""
    # import pdb; pdb.set_trace()
    async def test_func():
        # directly trigger the retry logic
        raise exceptions.TooManyRequests

    # let's not actually introduce a wait into the tests
    mock_wait.return_value = 0

    async with http.Session() as ps:
        with pytest.raises(exceptions.TooManyRequests):
            await ps._retry(test_func)

    calls = mock_wait.call_args_list
    args = [c[0] for c in calls]
    assert args == [(1, 64), (2, 64), (3, 64), (4, 64), (5, 64)]
```

When I run this, I still see the fcn trying to wait for the wait times the original `_calculate_wait()` would return: 2, 4, 8, etc. instead of my mock, which returns 0.

jreiberkyle:
this looks ready to merge. we can always improve the test to use the decorator version of patch at a future date.

Issue #534, opened 2022-05-11
implement interface skeleton for data api support for creating a data search filter


jreiberkyle:
Blocks #533 and #535

PR #553, opened 2022-05-13
Data filter python api

### LINKED ISSUE #533: implement data api support for creating a data search filter
similar to `order_request` ([ref](https://github.com/planetlabs/planet-client-python/blob/main/planet/order_request.py)), implement `data_filter`, which supports creation of a data search filter. This is the backend for `planet data filter` CLI command

jreiberkyle:
Closes #533 and #534

### DIFF
--- a/docs/guide.md
+++ b/docs/guide.md
@@ -189,22 +189,22 @@ the context of a `Session` with the `OrdersClient`:
 
 ```
 
-#### Polling and Downloading an Order
+#### Waiting and Downloading an Order
 
 Once an order is created, the Orders API takes some time to create the order
 and thus we must wait a while before downloading the order.
-We can use polling to watch the order creation process and find out when the
+We can use waiting to watch the order creation process and find out when the
 order is created successfully and ready to download.
 
-With polling and download, it is often desired to track progress as these
+With wait and download, it is often desired to track progress as these
 processes can take a long time. Therefore, in this example, we use a progress
-bar from the `reporting` module to report poll status. `download_order` has
+bar from the `reporting` module to report wait status. `download_order` has
 reporting built in.
 
 ```python
 from planet import reporting
 
->>> async def create_poll_and_download():
+>>> async def create_wait_and_download():
 ...     async with Session() as sess:
 ...         cl = OrdersClient(sess)
 ...         with reporting.StateBar(state='creating') as bar:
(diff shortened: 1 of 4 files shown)


planet/data_filter.py
```diff
@@ -0,0 +1,118 @@
...
+    """
+    return {'type': 'NotFilter', 'config': nested_filter}
+
+
+def _field_filter(ftype: str, field_name: str, config: Union[dict,
+                                                             list]) -> dict:
+    return {'type': ftype, 'field_name': field_name, 'config': config}
+
+
+def date_range_filter(field_name: str,
```
jreiberkyle, line 72: @sgillies I'm curious about your thoughts on this parameter structure for the optional conditionals. Data API ref: https://developers.planet.com/docs/apis/data/searches-filtering/#daterangefilter

planet/data_filter.py
```diff
@@ -0,0 +1,118 @@
...
+    }
+
+    LOGGER.warning(config)
+
+    return _field_filter('DateRangeFilter',
+                         field_name=field_name,
+                         config=config)
+
+
+def datetime_to_rfc3339(value: datetime) -> str:
```
jreiberkyle, line 112: @sgillies what do you think of this way of managing an RFC339 date string? Alternatively, we could use an external library. I remember you lamenting adding dependencies for libraries of which one only uses a small portion.

tests/unit/test_data_filter.py
```diff
@@ -0,0 +1,105 @@
...
+            'lte': '2022-06-01T01:00:00Z',
+        }
+    }
+    assert res == expected
+
+
+def test_date_range_filter_timezone():
+    # Make a concrete tzinfo class for testing
+    # https://docs.python.org/3.7/library/datetime.html#datetime.tzinfo
+    class TestTZ(tzinfo):
```
jreiberkyle, line 73: this seems like a lot of extra code for a test - it always makes me a bit unsure if I'm introducing more bugs in the test than I'm squashing by having a test.

docs/guide.md
```diff
@@ -221,6 +221,71 @@ from planet import reporting
 >>> asyncio.run(create_poll_and_download())
 ```
 
+### Data Client
```
jreiberkyle, line 224: @cholmes check out this section for proposed data api interface and usage.

jreiberkyle:
@cholmes and @sgillies this is ready for a review. I'll go ahead and merge on Friday if no comments or requests for more time are raised.

planet/data_filter.py
sgillies: Do you mean my thoughts on assembling a config dict from named keyword args instead of taking a config dict as an arg? I feel like what you're doing shapes the input in a positive way and is user-friendly. If you wanted to future proof these filters against changes in operators, I guess you could add an optional `config: dict = None` arg and merge it with the other options?

cholmes (approved):
Generally looks good to me, but I'm not a python dev, so most of this is a bit beyond me.

planet/data_filter.py
```diff
@@ -0,0 +1,118 @@
...
+
+    One or more of the conditional parameters `gt`, `lt`, `gte`, `lte` must be
+    specified.
+
+    Parameters:
+        field_name: Name of field to filter on
+        gt: Filter to field timestamp later than this value
+        lt: Filter to field timestamp earlier than this value
+        gte: Filter to field timestamp at or later than this value
+        lte: Filter to field timestamp at or earlier than this value
```
sgillies: @jreiberkyle is the result the logical conjunction of all of these filters, or the disjunction? I wonder if the doc should say something about that.

sgillies: For example, what will be the result of `date_range_filter(lt=x, gt=x)`?

jreiberkyle: yeah it's a good idea to be explicit. fyi in response to your question, that query would get no results.

planet/data_filter.py
jreiberkyle: Thanks for your thoughts. That is a great suggestion. For now I feel fine with not worrying too much about future proofing, my primary concern is how user-friendly it is. If we're generally ok with it, I'll keep it like it is.

planet/data_filter.py
jreiberkyle: Hah! I just went to make this change and it looks like it was changed to a private fcn at some point between this commit and the latest.

jreiberkyle:
Thank you for your reviews!

PR #555, opened 2022-05-18
implement search-quick functionality

mkshah605:


tests/integration/test_data_cli.py
```diff
@@ -5,16 +5,60 @@
...
+    """Test for planet data search_quick. Test should fail as filter
+    does not contain valid JSON."""
+    runner = CliRunner()
+    item_type = 'SkySatScene'
+    result = invoke(["search_quick", filter, item_type], runner=runner)
+    assert result.exit_code == 2
+
+
+@pytest.mark.asyncio
+async def test_data_search_quick_filter_success(invoke):
```
sgillies, line 45: The search-quick command provides its own event loop (from the `@coro` decorator). Tests that use the CliRunner will not be async.

```suggestion
def test_data_search_quick_filter_success(invoke):
```

tests/integration/test_data_cli.py
```diff
@@ -5,16 +5,60 @@
...
+    filter = {
+        "type": "DateRangeFilter",
+        "field_name": "acquired",
+        "config": {
+            "gt": "2019-12-31T00:00:00Z", "lte": "2020-01-31T00:00:00Z"
+        }
+    }
+    runner = CliRunner()
+    item_type = 'SkySatScene'
+    result = invoke(["search_quick", filter, item_type], runner=runner)
```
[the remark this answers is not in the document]
mkshah605: wow thank you! Yes I think I was getting confused because the dictionaries did look like json

tests/integration/test_data_cli.py
```diff
@@ -5,16 +5,59 @@
...
+
+@pytest.fixture
+def invoke():
+
+    def _invoke(extra_args, runner=None):
+        runner = runner or CliRunner()
+        args = ['data', '--base-url', TEST_URL] + extra_args
+        return runner.invoke(cli.main, args=args)
+
+    return _invoke
```
sgillies, line 21: @jreiberkyle I had a couple thoughts about this fixture while pairing with @mkshah605. First, it hides the `catch_exceptions=False` option of CliRunner.invoke, which is really handy when developing. Second, it puts us in the position of faking the response to a fake URL, which is, in my opinion, one step too far.

I suggest we lean towards configuring the respx mock like `respx.post('https://api.planet.com/data/v1/quick-search').return_value = mock_resp`, using the real endpoint URLs whenever we can. Thoughts?

planet/cli/data.py
```diff
@@ -23,83 +45,66 @@ def parse_item_types(ctx, param, value: str) -> List[str]:
...
-                                        name=name):
-        echo_json(feature, pretty=pretty)
+    """
+    async with data_client(ctx) as cl:
+        items = await cl.quick_search(name=name,
+                                      item_types=item_types,
+                                      search_filter=filter,
+                                      limit=limit,
+                                      sort=None)
+        async for item in items:
```
sgillies, line 106: @jreiberkyle remember our discussion about having quick_search fail fast when there are no result items? I had a related thought while working on the above with Mansi.

My first take at iterating over the items was this:

```python
    async with data_client(ctx) as cl:
        items = await cl.quick_search(name=name,
                                      item_types=item_types,
                                      search_filter=filter,
                                      limit=limit,
                                      sort=None)
    async for item in items:
        echo_json(item, pretty)
```

I tried to iterate outside of the data_client block and that did not work and had a not obvious error. I suspect that users might stumble on this too. The `items` object looks like it is independent of the client, but it is not (on account of the do_request_fn).

What would you think about changes to allow the fool-proof usage below?

```python
async with data_client(ctx) as cl:
    async for item in cl.quick_search(...):
        echo_json(item)
```

mkshah605:
Thank you @sgillies for your help!! Ready for review after some more changes, i'm wondering if the `test_data_search_quick_filter_fail` test should have an Assertion, or is it enough to keep it as-is since it's wrapped in the `pytest.raises(TypeError):`? In the cli.py code, it should hit `click.ClickException('File does not contain valid json.')` when this happens so wondering if I need to put that in somehow.

tests/integration/test_data_cli.py
```diff
@@ -2,19 +2,86 @@
...
+    Test should fail as filter does not contain valid JSON."""
+    mock_resp = httpx.Response(HTTPStatus.OK,
+                               json={'features': [{
+                                   "key": "value"
+                               }]})
+    respx.post(TEST_QUICKSEARCH_URL).return_value = mock_resp
+
+    runner = CliRunner()
+    item_types = 'SkySatScene'
+    with pytest.raises(TypeError):
```
sgillies, line 52: @mkshah605 to answer your question: pytest.raises is just as good as an assert statement when checking that code raises an exception.

However, we intend that the CLI does not let exceptions propagate all the way out. In the integration tests we should assert `assert result.exit_code == 1` when we are testing for failure. If a TypeError is being raised, it's likely that there is a bug.

mkshah605: @sgillies here's the TypeError I'm getting when I give bad inputs like {"foo"} or {''} to the quick-search function filter argument: `TypeError: Object of type set is not JSON serializable`. When I run the code (as a user) this way: `planet data search-quick PSScene {'foo'}  ` I get what I expect: `Error: File does not contain valid json.`, which I've put in the `parse_filter` function in `planet/cli/data.py`.

sgillies: @mkshah605 I found the problem! I will put another comment inline.

tests/integration/test_data_cli.py
```diff
@@ -2,19 +2,86 @@
...
+        result = invoke(["search-quick", item_types, json.dumps(filter)],
+                        runner=runner)
```
sgillies, line 54: And here we no longer need json.dumps. `filter` is already (bad) JSON.

```suggestion
        result = invoke(["search-quick", item_types, filter], runner=runner)
```

planet/cli/data.py
```diff
@@ -23,83 +45,66 @@ def parse_item_types(ctx, param, value: str) -> List[str]:
...
-@click.group()
-@click.pass_context
-def data(ctx):
-    """A group of commands for interacting with the Data API."""
+    # read filter using raw json
+    if value.startswith('{'):
+        try:
+            json_value = json.loads(value)
+        except json.decoder.JSONDecodeError:
+            raise click.ClickException('File does not contain valid json.')
```
sgillies, line 53: @mkshah605 I recommend the following. It will cause click to print out a helpful error message about the problems with the input. I think it would be good if these callback functions only raise BadParameter.

```suggestion
        except json.decoder.JSONDecodeError as err:
            raise click.BadParameter(err, ctx=ctx, param=param)
```

mkshah605: hmmm even with this, I'm seeing the error when I run the user-side code, but it's not registering when I run the test. The test is failing because it's expecting an error to be raised but that's not currently happening. Somehow it looks like the error is not registering?

sgillies: @mkshah605 correct! When there's a bad parameter, we don't want to propagate an exception. Instead we want to print a message and set an exit code of 2. And that's what click does when it encounters a BadParameter. See number 2 under https://click.palletsprojects.com/en/8.1.x/exceptions/?highlight=badparameter#where-are-errors-handled.

sgillies: Also, I strongly recommend passing the JSON decode error to BadParameter as I show above. The error message is very specific.

tests/integration/test_data_cli.py
```diff
@@ -1,20 +1,85 @@
...
+    Test should fail as filter does not contain valid JSON."""
+    mock_resp = httpx.Response(HTTPStatus.OK,
+                               json={'features': [{
+                                   "key": "value"
+                               }]})
+    respx.post(TEST_QUICKSEARCH_URL).return_value = mock_resp
+
+    runner = CliRunner()
+    with pytest.raises(click.exceptions.BadParameter):
+        result = invoke(["search-quick", item_types, filter], runner=runner)
```
sgillies, line 52:
```suggestion
result = invoke(["search-quick", item_types, filter], runner=runner)
assert result.exit_code == 2
```

mkshah605: thank you!! It turns out I was getting the wrong exit code because the test input was a dict instead of a string.

mkshah605:
@sgillies @kevinlacaille @jreiberkyle let me know if y'all have additional thoughts/feedback before this can be merged. I'm also working on `search-create` this sprint which uses a lot of the same building blocks, so can develop that more as soon as this is merged.

planet/cli/data.py
```diff
@@ -53,6 +53,10 @@ def parse_filter(ctx, param, value: str) -> dict:
             raise click.BadParameter('Filter does not contain valid json.',
                                      ctx=ctx,
                                      param=param)
+        if json_value == {}:
+            raise click.BadParameter('Filter does not contain valid json.',
```
sgillies, line 57: `'{}'` is valid JSON, though... how about 'Filter is empty'?

That said, I honestly don't know if empty filters are a problem. If someone searches with an empty filter, should they expect nothing or everything? I feel like the word "filter" implies the latter. Filters reduce information.

planet/cli/data.py
```diff
@@ -23,83 +45,74 @@ def parse_item_types(ctx, param, value: str) -> List[str]:
...
+                                     ctx=ctx,
+                                     param=param)
+        if json_value == {}:
+            raise click.BadParameter('Filter does not contain valid json.',
+                                     ctx=ctx,
+                                     param=param)
+        return json_value
+    # read filter using click pipe option
+    elif value == '-' or Path(value).name.endswith('.json'):
```
sgillies, line 62: @mkshah605 @kevinlacaille we can't rely on file extensions. Some people use .json, some use .geojson, some use no extension at all.

I think `else:` would be fine.

planet/cli/data.py
```diff
@@ -23,83 +45,74 @@ def parse_item_types(ctx, param, value: str) -> List[str]:
...
+            with click.open_file(value) as f:
+                json_value = json.load(f)
+        except json.decoder.JSONDecodeError:
+            raise click.BadParameter('Filter does not contain valid json.',
+                                     ctx=ctx,
+                                     param=param)
+        return json_value
+    else:
+        raise click.ClickException(
+            'Please pass filter using filename or STDIN.')
```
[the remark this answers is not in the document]
sgillies: According to https://github.com/planetlabs/planet-client-python/runs/6557612468?check_suite_focus=true#step:5:51, those lines aren't executed, so it'll be fine to remove them for now.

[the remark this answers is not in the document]
sgillies: Good work overall @mkshah605 !

planet/cli/data.py
mkshah605: Agreed!! I'll get rid of the empty json bringing up an error.

mkshah605: Actually, looks like we need something in the filter. If not, we get this error: `Error: {"field": {"filter.type": [{"message": "'type' is a required property"}]}, "general": []}`

sgillies (approved):
Great, @mkshah605 !

planet/cli/data.py
sgillies: Thanks for looking that up!

PR #559, opened 2022-05-19
Match folder structure in order description for download

### LINKED ISSUE #322: RFC: Match folder structure of Orders API and manifest upon order download
To enable catching missing or incomplete file downloads after an order has been downloaded, match the folder structure of Orders API and `manifest.json` upon order download.

The manifest (`manifest.json`) included in an order provides paths and checksums for every other in the order. Matching the folder structure of `manifest.json` (and, similarly, the Orders API), would provide for comparison between the local filesystem and the manifest.

This is actually pretty easy to accomplish, as the Orders API gives the folder structure for each result.

Here’s an excerpt from an order description returned by the Orders API:
```python
"results": [
      {
        "delivery": "success",
        "expires_at": "2021-09-24T17:34:31.400Z",
        "location": "https://api.planet.com/compute/ops/download/?token=<LongString>",
        "name": "ccc38057-6523-4d75-916d-0de033174abc/PSScene4Band/20200909_182524_1014_3B_AnalyticMS_clip.tif"
      },
...
```
See how there is a `name` entry in that description? It includes the order name as a subfolder but also includes the item-type as a nested folder. Currently, the SDK is only making use of the `location` value, it doesn't even look at name at all! This turns out to be problematic when it comes to comparing to the manifest.

Here is the corresponding entry in the downloaded `manifest.json` file:
```python
{
      "path": "PSScene4Band/20200909_182524_1014_3B_AnalyticMS_clip.tif",
      "media_type": "image/tiff",
      "size": 35254380,
      "digests": {
        "md5": "<string>",
        "sha256": "<string>"
      },
      "annotations": {
        "planet/asset_type": "analytic",
        "planet/bundle_type": "analytic",
        "planet/item_id": "20200909_182524_1014",
        "planet/item_type": "PSScene4Band"
      }
    }
```
Note that `path` is pre-built by the Orders API and is going to be that value regardless of what we chose to actually name the downloaded file. It only differs from the order `name` value in that it doesn't include the subfolder named after the order id. But if you are already in the <order_id> folder and looking at `manifest.json` trying to determine if you got all the files, it would be great if the folder structure was the same as is given in the manifest.

So, how can we follow the Orders API order description exactly?

By having `download_order()` pass the value of the order result `name` as well as the order result `location` to `download_asset()`. We can change `download_asset()` to have `path` as an input instead of `directory` and `filename`, and `path` includes both the name and the directory. If `path` isn't specified, then `download_asset()` defaults to downloading to `os.path.join('.', body.name)`.

Con: With following this convention, we lose the 'freedom' to download all files into the local directory. This could be resolved by providing a `flatten` Bool parameter that would result in `download_order` passing `path=None` to `download_asset()`, which would trigger the current behavior, that is `os.path.join('.', body.name)`

(Side note: According to the Orders API [delivery documentation](https://developers.planet.com/docs/orders/delivery/), the manifest is the last file downloaded, this isn't necessarily the case with the SDK, do we want it to be? Is it possible with the async download framework?)

Usability:

Download an order and check against the manifest:
```python
...
# not sure if we want check_manifest to be a separate function call or included as an option
# A separate function call is cleaner, but how do we know what subfolder the manifest was downloaded
# into?
orders_client.download_order(<order_id>, directory=<my_downloads_dir>, check_manifest=True)
```

Match old behavior:
```python
...
# If flatten is true do we even want to download the manifest?
orders_client.download_order(<order_id>, directory=<my_downloads_dir>, flatten=True)
```
Here, `flatten` and `check_manifest` both set to `True` is nonsense ;)

kevinlacaille:
Closes #322

### DIFF
--- a/planet/clients/orders.py
+++ b/planet/clients/orders.py
@@ -243,6 +243,7 @@ class OrdersClient:
 
     async def download_asset(self,
                              location: str,
+                             order_description_path: str,
                              filename: str = None,
                              directory: str = None,
                              overwrite: bool = False,
@@ -251,6 +252,7 @@ class OrdersClient:
 
         Parameters:
             location: Download location url including download token.
+            order_description_path: Path to file in order description
             filename: Custom name to assign to downloaded file.
             directory: Base directory for file download.
             overwrite: Overwrite any existing files.
@@ -266,7 +268,10 @@ class OrdersClient:
 
         async with self._session.stream(req) as resp:
             body = StreamingBody(resp)
-            dl_path = os.path.join(directory or '.', filename or body.name)
+            dl_dir = os.path.join(directory or '.', order_description_path)
+            if not os.path.isdir(dl_dir):
+                os.mkdir(dl_dir)
+            dl_path = os.path.join(dl_dir, filename or body.name)
             await body.write(dl_path,
                              overwrite=overwrite,
                              progress_bar=progress_bar)
@@ -355,11 +360,13 @@ class OrdersClient:
                 'Consider using wait functionality before '
                 'attempting to download.')
         locations = self._get_order_locations(order)
+        order_description_path = self._get_order_description_path(order)
         LOGGER.info(
             f'downloading {len(locations)} assets from order {order_id}')
 
         filenames = [
             await self.download_asset(location,
+                                      order_description_path,
                                       directory=directory,
                                       overwrite=overwrite,
                                       progress_bar=progress_bar)
@@ -393,8 +400,19 @@ class OrdersClient:
             return list(r['location'] for r in results if r)
         except TypeError:
             LOGGER.warning(
-                'order does not have any locations, will not download any ' +
+                'Order does not have any locations, will not download any ' +
                 'files.')
+            return []
+
+    @staticmethod
+    def _get_order_description_path(order):
+        links = order['_links']
+        results = links.get('results', [])
+        try:
+            # Currently, this returns the order ID, but this may change
+            return os.path.dirname(results[0]['name'])
+        except TypeError:
+            LOGGER.warning('Order does not contain any files.')
             return []
 
     async def wait(self,


kevinlacaille:
Editing tests to get `download_assert()` to take in the argument `order_description_path`, but the order description path 
for these tests is empty, namely the key `name` is set to `name1` in `tests/data/order_description_b0cb3448-0a74-11eb-92a1-a3d779bb08e0.json`. 

Thoughts on adding a nested path to the value `name1`? Specifically, I propose adding the order ID in front like so:
`"name": "b0cb3448-0a74-11eb-92a1-a3d779bb08e0/name1"`
as that's the current format returned by the Orders API (see here: https://developers.planet.com/docs/orders/ordering/#download-an-order)

cc: @sgillies @jreiberkyle

jreiberkyle:
I do not believe a new parameter is needed, I was thinking `directory` and `filename` could be used for this.

For example, I just downloaded order `193e5bd1-dedc-4c65-a539-6bc70e55d928`. I got all the files in the local dir. One of the entries in the `manifest.json` file is
```json
    {
      "path": "PSScene3Band/20200909_182525_1014_3B_Analytic.tif",
      "media_type": "image/tiff",
      "size": 140516607,
      "digests": {
        "md5": "ce9d41384139e520c3c0666205b9e780",
        "sha256": "5c3d87a84b6251737bd41da54fedb30d4aeec11542e83081bc8a3abe687aac89"
      },
```
so `download_asset()` should be called with 
```python
filename = '20200909_182525_1014_3B_Analytic.tif'
directory = Path('193e5bd1-dedc-4c65-a539-6bc70e55d928', 'PSScene3Band')
download_asset(<location_url>, filename=filename, directory=directory, ... )
```
which would result in the desired nested structure for the download.

kevinlacaille:
Closed by #570 .

Issue #536, opened 2022-05-11
Implement planet collect CLI


cholmes:
As specified in https://github.com/planetlabs/planet-client-python/blob/main/design-docs/CLI-Core.md#collect

PR #563, opened 2022-05-25
implement collect python api and cli

### LINKED ISSUE #538: implement collect python api
Implement python functionality that takes the output of e.g. `DataClient.quick_search()` and creates a single FeatureCollection, or the output of `OrdersClient.list()` and returns a single JSON blob.

Blocks #536

jreiberkyle:
Implement `collect` python api and cli commands.

Specifics:
- introduce `io` module to hold `collect` command
- introduce `collect` cli command
- move exceptions housed in `geojson` module to `exceptions.GeoJSONError` module
- add functionality for identifying if a JSON blob is an instance of a GeoJSON Feature and functionality for generating a GeoJSON FeatureClass from GeoJSON Features
  - Add `Feature` schema and move data dir specification from `specs` module to `constants`
- Fix inaccurate name `featureclass`, change to `FeatureCollection`

Closes #538 and #536


jreiberkyle:
@cholmes - check out the usage in the docs and lmk how they look.

@sgillies - your feedback on the placement of the collect command in the cli and api would be greatly appreciated, along with anything else you see.

TY!

docs/guide.md
```diff
@@ -304,6 +331,22 @@ for passing into a Docker instance:
...
+Some API calls, such as searching for imagery and listing orders, return a
+varying, and potentially large, number of results. These API responses are
+paged. The SDK manages paging internally and the associated cli commands
+output the results as a sequence. These results can be converted to a JSON blob
+using the `collect` command. When the results
+represent GeoJSON features, the JSON blob is a GeoJSON FeatureCollection.
+Otherwise, the JSON blob is a list of the individual results.
+
+```console
+$ planet orders list | planet collect -
```
cholmes, line 346: I'd probably make the main example to be the data search-quick one, since that'll be the most common.

Also why is it with a '-'? Might be worth to explain that somewhere. I think we probably should have a 'CLI primer' somewhere that covers piping results and the like, as I've had to level up a bit to use some of these things. I like it overall, it's quite powerful, but we need to lead people to it.

cholmes (approved):
Looks great! One minor comment from me.

docs/guide.md
jreiberkyle: good points. I made a ticket for the cli primer. The `-` for stdin is the format we've been using. Here, and elsewhere, we can have the `-` be the default value so it's not required. I'll make the move to `planet data search-quick`

jreiberkyle:
@sgillies I'm going to merge but feel free to open a discussion / ticket if any suggestions come up

planet/io.py
```diff
@@ -0,0 +1,42 @@
...
+LOGGER = logging.getLogger(__name__)
+
+
+async def collect(
+    values: typing.AsyncIterator[dict]
+) -> typing.Union[typing.List[dict], dict]:
+    """Collect a sequence into JSON.
+
+    If the items in values are GeoJSON features, the JSON blob is a GeoJSON
+    FeatureCollection. Otherwise, the JSON blob is a list of the items.
```
sgillies, line 29: @jreiberkyle @cholmes I feel that functions that have one kind of output are easier to explain to users and easier to maintain than functions that have different kinds of output. So maybe collect could return only a feature collection?

docs/guide.md
```diff
@@ -88,6 +88,33 @@ environment variable:
...
+
+```python
+>>> import asyncio
+>>> from planet import collect, OrdersClient, Session
+>>>
+>>> async def main():
+...     async with Session() as sess:
+...         client = OrdersClient(sess)
+...         orders = client.list_orders()
+...         orders_list = collect(orders)
```
sgillies, line 111: @jreiberkyle @cholmes when I try to imagine that I'm reading that example without knowing any of the details, I wonder "Why doesn't list_orders return a list? Why do I need to call collect?" :phone:

In the Python standard library, I believe list_orders would be named iter_orders to communicate that it's an iterator that must be materialized into a list. And the canonical way of doing that would be to use Python's built in list function. Like `orders_list = list(iter_orders)`.

Perhaps we can emphasize the feature collection use case here instead of lists? Like `orders_collection = collect(orders)`?

planet/cli/collect.py
```diff
@@ -0,0 +1,50 @@
...
+    Output can be pretty-printed with --pretty option.
+    """
+
+    # make an AsyncGenerator from the input lines
+    async def _entries_aiter():
+        for line in input:
+            yield json.loads(line)
+
+    entries = _entries_aiter()
+    collected = await planet.collect(entries)
```
sgillies, line 49: @jreiberkyle if you wanted, you could replace lines 42-49 with 

`collected = await planet.collect(json.loads(line) for line in input)`

I think that's equally clear, maybe a bit more clear.

Issue #431, opened 2022-04-21
Skeleton of Subscriptions CLI


sgillies:
Write the minimal subscriptions CLI module so that our example tests can run and so that "planet subscriptions" shows up under "planet --help".

cholmes:
Any update on this @sgillies ?

sgillies:
Work on #430 and #565 is coming first as an exercise in working backwards.

PR #565, opened 2022-05-25
Initial skeleton of subscription CLI and its tests

### LINKED ISSUE #430: Create example tests for Subscriptions CLI
I'm proposing to create at least a couple tests for the Subscriptions CLI and mark them as expected to fail. Before we have our first pre-release we'll hook them up to the client module, remove the xfail marker, and get them to pass.

cc @jreiberkyle @kevinlacaille @mkshah605 .

sgillies:
This is test-driven to the extreme. The skeleton of the CLI begins in the bodies of the tests. There's no planet.cli.subscriptions at the start.

Resolves #430 and #431

### DIFF
--- a/planet/cli/subscriptions.py
+++ b/planet/cli/subscriptions.py
@@ -13,6 +13,7 @@ from planet.exceptions import PlanetError
 # A collection of fake subscriptions for testing. Tests will
 # monkeypatch this attribute.
 _fake_subs = None
+_fake_sub_results = None
 
 
 # For test use.
@@ -33,6 +34,10 @@ def _get_fake_sub(sub_id):
     sub = _fake_subs[sub_id].copy()
     sub.update(id=sub_id)
     return sub
+
+
+def _get_fake_sub_results(sub_id):
+    return _fake_sub_results[sub_id]
 
 
 @click.group()
@@ -190,6 +195,36 @@ async def update_subscription(ctx, subscription_id, request, pretty):
     # *will* raise PlanetError like this.
     try:
         sub = _update_fake_sub(subscription_id, **request)
+    except KeyError:
+        raise PlanetError(f"No such subscription: {subscription_id!r}")
+
+    # End fake subscriptions service. After we refactor we will get
+    # the "sub" from a method in planet.clients.subscriptions (which
+    # doesn't exist yet).
+
+    echo_json(sub, pretty)
+
+
+@subscriptions.command(name='describe')
+@click.argument('subscription_id')
+@click.option('--pretty', is_flag=True, help='Pretty-print output.')
+@click.pass_context
+@translate_exceptions
+@coro
+async def describe_subscription(ctx, subscription_id, pretty):
+    """Cancels a subscription and prints the API response.
+
+    This implementation is only a placeholder. To begin, instead
+    of mocking calls to the Subscriptions API, we'll use a
+    collection of fake subscriptions (the all_subs object).
+    After we refactor we will change to mocking the API.
+
+    """
+    # Begin fake subscriptions service. Note that the Subscriptions
+    # API will report missing keys differently, but the Python API
+    # *will* raise PlanetError like this.
+    try:
+        sub = _get_fake_sub(subscription_id)
     except KeyError:
         raise PlanetError(f"No such subscription: {subscription_id!r}")
 
--- a/planet/cli/subscriptions.py
+++ b/planet/cli/subscriptions.py
@@ -1,0 +1,126 @@
+"""Subscriptions CLI"""
+
+import itertools
+import json
+
+import click
+
+import planet
+from planet.cli.cmds import coro, translate_exceptions
+from planet.cli.io import echo_json
+from planet.exceptions import PlanetError
+
+# A collection of fake subscriptions for testing. Tests will
+# monkeypatch this attribute.
+_fake_subs = None
+
+
+# For test use.
+def _count_fake_subs():
+    return len(_fake_subs)
+
+
+@click.group()
+@click.pass_context
+def subscriptions(ctx):
+    ctx.obj['AUTH'] = planet.Auth.from_file()
+
+
+# We want our command to be known as "list" on the command line but
+# don't want to clobber Python's built-in "list". We'll define the
+# command function as "list_subscriptions".
+@subscriptions.command(name="list")
+@click.option('--pretty', is_flag=True, help='Pretty-print output.')
+@click.option(
+    '--status',
+    type=click.Choice(["created", "queued", "processing", "failed",
+                       "success"]),
+    multiple=True,
+    default=None,
+    help="Select subscriptions in one or more states. Default is all.")
+@click.option('--limit',
+              type=int,
+              default=100,
+              help='Maximum number of results to return. Defaults to 100.')
+@click.pass_context
+@translate_exceptions
+@coro
+async def list_subscriptions(ctx, status, limit, pretty):
+    """Prints a sequence of JSON-encoded Subscription descriptions.
+
+    This implementation is only a placeholder. To begin, instead
+    of mocking calls to the Subscriptions API, we'll use a
+    collection of fake subscriptions (the all_subs fixture).
+    After we refactor we will change to mocking the API.
+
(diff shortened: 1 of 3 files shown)


sgillies:
Commit 9f4cfb0 has 3 goals.

1. Set the stage for future sprints on the Subscriptions CLI and API.
2. Demonstrate how to start testing a complicated project by pinning down the CLI basics of input, output, and options without prematurely sweating over the tricky implementation details. Establishing a solid foothold where roughly correct inputs produce the right exit code (0) and obviously incorrect inputs produce the right exit code (1 or 2, depending) can give us confidence in working out the finer details in a next iteration.
3. Normalize the creation of placeholders and scaffolding, code that will be replaced after it has served its purpose. It's okay to use rough methods to make progress early in the development of a feature.

planet/cli/subscriptions.py
```diff
@@ -0,0 +1,126 @@
...
+from planet.exceptions import PlanetError
+
+# A collection of fake subscriptions for testing. Tests will
+# monkeypatch this attribute.
+_fake_subs = None
+
+
+# For test use.
+def _count_fake_subs():
+    return len(_fake_subs)
```
sgillies, line 20: The `_fake_subs` attribute and `_count_fake_subs` function, at this stage of our work, give us a very rough fake Subscriptions API.

planet/cli/subscriptions.py
```diff
@@ -0,0 +1,126 @@
...
+
+    # Update the request with an id.
+    sub = dict(**request, id='42')
+    _fake_subs.append(sub)
+
+    # End fake subscriptions service. After we refactor we will get
+    # the "sub" from a method in planet.clients.subscriptions (which
+    # doesn't exist yet).
+
+    echo_json(sub, pretty)
```
sgillies, line 126: These functions are copied straight from the bodies of the tests in which they were born with almost no change.

tests/integration/test_subscriptions_cli.py
```diff
@@ -11,32 +11,14 @@
...
-def all_subs():
-    """A collection of fake subscriptions for testing.
-
-    We're not testing the actual API workflow. All subscription states
-    are equal as far as the initial tests are concerned. We have 101
-    fake subscriptions so that we can test the command's default limit
-    of 100.
-
-    """
-    return [{'id': str(i), 'status': 'created'} for i in range(101)]
```
sgillies: This fixture is gone, it's been changed into planet.cli.subscription's `_fake_subs` attribute.

sgillies:
Commit 660f426 moves the skeleton of the Subscriptions CLI from the bodies of the tests, where they began, into a proper planet.cli.subscriptions module. Other than this, almost nothing changes. The tests continue to pass, assuring us that we moved the code correctly.

sgillies:
Commit eb6ec08 adds a test of subscription creation on the command line, still using the fake API contained in planet.subscriptions.cli. We can continue to rough in the CLI and its tests just like this, without using the final Python subscriptions client (which is still a work in progress).

tests/integration/test_subscriptions_cli.py
```diff
@@ -78,3 +78,30 @@ def test_subscriptions_create_failure(monkeypatch):
...
+        input=json.dumps(sub),
+        # Note: catch_exceptions=True (the default) is required if we want
+        # to exercise the "translate_exceptions" decorator and test for
+        # failure.
+        catch_exceptions=True)
+
+    assert result.exit_code == 0  # success.
+    assert "Request lacks required members" not in result.output
+    assert json.loads(result.output)['id'] == '42'
+    assert _count_fake_subs() == 1
```
sgillies, line 111: Note that lines 106 and 107 depend on implementation details of our fake API.

tests/integration/test_subscriptions_cli.py
```diff
@@ -80,22 +80,26 @@ def test_subscriptions_create_failure(monkeypatch):
...
+# This subscription request has the members required by our fake API.
+# It must be updated when we begin to test against a more strict
+# imitation of the Planet Subscriptions API.
+GOOD_SUB_REQUEST = {'name': 'lol', 'delivery': True, 'source': 'wut'}
+
+
+@pytest.mark.parametrize('cmd_arg, runner_input',
+                         [('-', json.dumps(GOOD_SUB_REQUEST)),
+                          (json.dumps(GOOD_SUB_REQUEST), None)])
+def test_subscriptions_create_success(cmd_arg, runner_input, monkeypatch):
```
sgillies, line 92: We intend that planet-subscriptions-create can take a JSON subscription request from a file (including stdin) or string. The new tests parametrize the command's argument and stdin input to cover both these cases.

sgillies:
Commit d3920a6 adds tests of two different sources of good and bad subscription requests. Still using the fake API. We can continue in this way for all the other subscriptions CLI commands using only the fake API, making necessary modifications as we go.

planet/cli/subscriptions.py
```diff
@@ -20,6 +20,10 @@ def _count_fake_subs():
     return len(_fake_subs)
 
 
+def _cancel_fake_sub(sub_id):
+    return _fake_subs.pop(sub_id)
```
sgillies, line 24: For our fake API, cancellation means pop an item out of the dictionary that represents the state of the API.

tests/integration/test_subscriptions_cli.py
```diff
@@ -133,3 +133,114 @@ def test_subscriptions_bad_request(cmd_arg, runner_input):
 
     assert result.exit_code == 2  # bad parameter.
     assert "Request does not contain valid json" in result.output
+
+
+def test_subscriptions_cancel_failure(monkeypatch):
+    """Cancel command exits gracefully from an API error."""
+
+    monkeypatch.setattr(planet.cli.subscriptions, '_fake_subs', {})
```
sgillies, line 141: Tweaking the state of the API for this new test. It's now a dictionary, where keys are subscription ids and values are subscriptions.

tests/integration/test_subscriptions_cli.py
```diff
@@ -133,3 +133,114 @@ def test_subscriptions_bad_request(cmd_arg, runner_input):
...
+        of mocking calls to the Subscriptions API, we'll use a
+        collection of fake subscriptions (the all_subs object).
+        After we refactor we will change to mocking the API.
+
+        """
+        # Begin fake subscriptions service. Note that the Subscriptions
+        # API will report missing keys differently, but the Python API
+        # *will* raise PlanetError like this.
+        try:
+            sub = _cancel_fake_sub(subscription_id)
```
sgillies, line 170: Why not import `_fake_subs` from planet.cli.subscriptions and pop the subscription from that? Because we could miss the monkeypatching.

tests/integration/test_subscriptions_cli.py
```diff
@@ -133,3 +133,114 @@ def test_subscriptions_bad_request(cmd_arg, runner_input):
...
+        except KeyError:
+            raise PlanetError(f"No such subscription: {subscription_id!r}")
```
sgillies, line 172: Working backwards, we now have a great idea of the basic shape of the production subscriptions client's `cancel` method. It will return the JSON description of a subscription on success, and raise PlanetError on failure.

(To be completely honest, we already worked that out when developing the orders client, but we could also have derived it by working backwards as shown in this PR.)

tests/integration/test_subscriptions_cli.py
```diff
@@ -133,3 +133,114 @@ def test_subscriptions_bad_request(cmd_arg, runner_input):
...
+    assert result.exit_code == 1  # failure.
+    assert "No such subscription" in result.output
+
+
+def test_subscriptions_cancel_success(monkeypatch):
+    """Cancel command exits gracefully from an API error."""
+
+    monkeypatch.setattr(planet.cli.subscriptions,
+                        '_fake_subs',
+                        {'42': dict(**GOOD_SUB_REQUEST, id='42')})
```
sgillies, line 197: We need a subscription in the API to make a success cancellation.

sgillies:
I stuck with the methodology all the way to the end :smile:

Issue #561, opened 2022-05-24
Add test for --limit to data search_quick CLI


mkshah605:
Realized while working on search_create that I forgot to add this test in.

PR #560, opened 2022-05-24
Implement search create #487

mkshah605:

PR #568, opened 2022-05-26
Search quick limit test #561

mkshah605:


mkshah605:
Note: this PR depends on https://github.com/planetlabs/planet-client-python/pull/560. Merge that one first before merging this one.

planet/clients/data.py
```diff
@@ -85,12 +85,14 @@ async def _do_request(self, request: Request) -> Response:
...
-                           name: str = None,
-                           sort: str = None,
-                           limit: int = None) -> typing.AsyncIterator[dict]:
+    async def quick_search(
+            self,
+            item_types: typing.List[str],
+            search_filter: dict,
+            name: str = None,
+            sort: str = None,
+            limit: typing.Union[int,
```
jreiberkyle, line 94: I'm curious where the need for supporting `limit=None` comes from?

mkshah605: so that even if limit is not specified (or set to None), the default of 100 items is still supported. We do a similar thing in `planet/clients/orders.py::list_orders` for the limit functionality there.

tests/integration/test_data_cli.py
```diff
@@ -81,6 +81,52 @@ def test_data_search_quick_filter_success(invoke, item_types):
     assert len(result.output.strip().split('\n')) == 1  # we have 1 feature
 
 
+@respx.mock
+@pytest.mark.parametrize("limit,limited_list_length", [(None, 100), (0, 102),
```
jreiberkyle, line 85: This is such a great test. Nice work on the parametrization.

planet/clients/data.py
```diff
@@ -85,12 +85,14 @@ async def _do_request(self, request: Request) -> Response:
...
+                                None] = 100) -> typing.AsyncIterator[dict]:
```
jreiberkyle, line 95: can you add documentation of the behavior if limit=None is specified?

jreiberkyle (approved):
looks great! Just a few documentation suggestions. feel free to merge when ready

tests/integration/test_data_cli.py
mkshah605: I also adapted this straight from another test you've written haha thank you for setting a great example!

planet/clients/data.py
mkshah605: of course!!

PR #570, opened 2022-05-31
match folder structure in order download to manifest

### LINKED ISSUE #322: RFC: Match folder structure of Orders API and manifest upon order download

jreiberkyle:
This changes order download from creating a flat folder structure to creating a folder structure that matches the entries in the manifest file. It also changes the checksum validation to look for files in the new structure.

Side effects:

- To minimize casting between Path and str, OrderClient download_asset() and download_order() parameters and return types changed from str to Path
- download_order() checksum validation now goes through entries in the manifest and fails if one of the entries is not present (previously would only match checksums in cases where both the file existed and the manifest entry was present)
- Moved `_validate_checksum()` to below `download_orders()` to match order of helper function definitions following primary function definition. In retrospect, I would have kept it where it was because the move masks the actual logic changes.

Closes #322

### DIFF
--- a/tests/integration/test_orders_api.py
+++ b/tests/integration/test_orders_api.py
@@ -551,11 +551,11 @@ async def test_download_asset_md(tmpdir, session):
 @respx.mock
 @pytest.mark.asyncio
 @pytest.mark.parametrize("checksum", [("MD5"), ("SHA256")])
-async def test_checksum_success(tmpdir,
-                                order_description,
-                                oid,
-                                session,
-                                checksum):
+async def test_download_order_checksum_success(tmpdir,
+                                               order_description,
+                                               oid,
+                                               session,
+                                               checksum):
     # Mock an HTTP response for download
     order_description['state'] = 'success'
     dl_url1 = TEST_DOWNLOAD_URL + '/asset1'
@@ -631,11 +631,11 @@ async def test_checksum_success(tmpdir,
 @respx.mock
 @pytest.mark.asyncio
 @pytest.mark.parametrize("checksum", [("MD5"), ("SHA256")])
-async def test_checksum_failure(tmpdir,
-                                order_description,
-                                oid,
-                                session,
-                                checksum):
+async def test_download_order_checksum_failure(tmpdir,
+                                               order_description,
+                                               oid,
+                                               session,
+                                               checksum):
     # Note: the hashkeys in the mock manifest below were changed
     # from the correct keys to temporary keys
     # This should cause the checksum to fail.
@@ -710,7 +710,7 @@ async def test_checksum_failure(tmpdir,
     # Test Checksum
     cl = OrdersClient(session, base_url=TEST_URL)
     with pytest.raises(exceptions.ClientError):
-        await cl.download_order(oid, directory=str(tmpdir), checksum=checksum)
+        await cl.download_order(oid, directory=Path(tmpdir), checksum=checksum)
 
 
 @respx.mock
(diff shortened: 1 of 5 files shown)


tests/integration/test_orders_api.py
```diff
@@ -822,37 +838,6 @@ async def test_download_order_state(tmpdir, order_description, oid, session):
         await cl.download_order(oid, directory=str(tmpdir))
 
 
-@respx.mock
-@pytest.mark.asyncio
-async def test_download_order_without_order_links(tmpdir,
```
jreiberkyle, line 827: taken care of in the parameterization of `test_download_order_success`

PR #571, opened 2022-05-31
Begin to extract the skeleton of a SubscriptionsClient from CLI

sgillies:
This PR extracts the outline of a production Subscriptions API client from the bodies of the CLI commands developed in #430. The names and signatures of the placeholder Subscriptions client are close to what they will be in production (in planet.clients.subscriptions) but the internals are not. The placeholder remains within the planet.cli.subscriptions module (its only application) for now. When development of the Subscriptions client begins, we can copy the placeholder to planet.clients.subscriptions and begin replacing the placeholder method implementations.

### DIFF
--- a/planet/cli/subscriptions.py
+++ b/planet/cli/subscriptions.py
@@ -46,10 +46,44 @@ class PlaceholderSubscriptionsClient:
         # Note: placeholder doesn't use the session (yet).
         self.session = session
 
-    async def list_results(self,
-                           subscription_id: str,
-                           status: Set[str] = None,
-                           limit: int = 100) -> AsyncIterator[dict]:
+    async def get_subscription(self, subscription_id: str) -> dict:
+        """Get a description of a Subscription.
+
+        Parameters:
+            subscription_id
+
+        Returns:
+            dict
+
+        Raises:
+
+        """
+        try:
+            sub = _fake_subs[subscription_id].copy()
+        except KeyError:
+            raise PlanetError(f"No such subscription: {subscription_id!r}")
+
+        sub.update(id=subscription_id)
+        return sub
+
+    async def list_subscription_results(
+            self,
+            subscription_id: str,
+            status: Set[str] = None,
+            limit: int = 100) -> AsyncIterator[dict]:
+        """Get Results of a Subscription.
+
+        Parameters:
+            subscription_id
+            status
+            limit
+
+        Returns:
+            dict
+
+        Raises:
+
+        """
         try:
             if status:
                 select_results = (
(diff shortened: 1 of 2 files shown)


planet/cli/subscriptions.py
```diff
@@ -262,34 +298,10 @@ async def list_subscription_results(ctx,
...
-
-    for result in filtered_results:
-        echo_json(result, pretty)
+    """Gets results of a subscription and prints the API response."""
+    async with subscriptions_client(ctx) as client:
+        filtered_results = client.list_results(subscription_id,
+                                               status=status,
+                                               limit=limit)
+        async for result in filtered_results:
+            echo_json(result, pretty)
```
sgillies, line 307: The details of the placeholder subscriptions client used to live in the CLI command but have now been moved to a PlaceholderSubscriptionsClient in the same module.

The body of the command is now quite close to its final state.

planet/cli/subscriptions.py
```diff
@@ -254,19 +288,9 @@ async def describe_subscription(ctx, subscription_id, pretty):
...
-        raise PlanetError(f"No such subscription: {subscription_id!r}")
-
-    # End fake subscriptions service. After we refactor we will get
-    # the "sub" from a method in planet.clients.subscriptions (which
-    # doesn't exist yet).
-
-    echo_json(sub, pretty)
+    async with subscriptions_client(ctx) as client:
+        sub = await client.get_subscription(subscription_id)
+        echo_json(sub, pretty)
```
sgillies, line 293: Body of this command is now also close to its final state.

planet/cli/subscriptions.py
```diff
@@ -19,20 +20,66 @@
...
 
-# The following 5 functions are for use in CLI commands until there
-# is subscriptions Python client to call on.
-def _count_fake_subs():
-    return len(_fake_subs)
+class PlaceholderSubscriptionsClient:
+    """A placeholder client.
 
+    This class and its methods are derived from tests of a skeleton
+    Subscriptions CLI.
```
sgillies, line 27: Important note: the only thing that's roughly right about this placeholder client is the method names and signatures. All of the code inside the methods will be very different in the future production client.

tests/integration/test_subscriptions_cli.py
```diff
@@ -13,14 +13,33 @@
...
+
+        async def count_em():
+            return len([
+                sub async for sub in
+                PlaceholderSubscriptionsClient().list_subscriptions()
+            ])
+
+        return asyncio.run(count_em())
+
+    return func
```
sgillies, line 42: Being able to check the count of fake subscriptions in the system provides a sanity check in a few tests. I wrote a fixture that uses the placeholder client within an event loop to do so. This won't work for the real Subscriptions API, so we'll eventually delete this fixture and all usage.

Issue #504, opened 2022-05-05
Implement CLI command: planet data search-get


cholmes:
As specified in https://github.com/planetlabs/planet-client-python/blob/main/design-docs/CLI-Data.md#search-get

cholmes:
Depends on #509

PR #573, opened 2022-06-02
Implement search get #504

mkshah605:


planet/cli/data.py
```diff
@@ -135,6 +135,22 @@ async def search_create(ctx, name, item_types, filter, daily_email, pretty):
...
+@click.pass_context
+@translate_exceptions
+@coro
+@pretty
+@click.argument('search_id')
+async def search_get(ctx, search_id, pretty):
+    """ This function obtsains an existing saved search, using the search_id.
+    This function outputs a full JSON description of the identified saved
+    search. The output can also be optionally pretty-printed using "--pretty".
+    """
```
jreiberkyle, line 148: I'd recommend updating this to have the first line be a short description and the longer description to follow. `planet orders get` as reference: https://github.com/planetlabs/planet-client-python/blob/a65a117ebba604a9c0e0c5dd2a95b5e519255ad3/planet/cli/orders.py#L84

planet/cli/data.py
```diff
@@ -135,6 +135,22 @@ async def search_create(ctx, name, item_types, filter, daily_email, pretty):
...
+    async with data_client(ctx) as cl:
+        items = await cl.get_search(search_id=search_id)
```
jreiberkyle, line 150: search_id is a required, positional argument. there is no need to specify it as a keyword arg

```suggestion
        items = await cl.get_search(search_id)
```

jreiberkyle, line 150: search_id is a required, positional argument. there is no need to specify it as a keyword arg

```suggestion
        items = await cl.get_search(search_id)
```

jreiberkyle (changes requested):
The code looks great. Added a few comments. Please add full api tests for `get_search`.

planet/cli/data.py
```diff
@@ -142,7 +142,9 @@ async def search_create(ctx, name, item_types, filter, daily_email, pretty):
 @pretty
 @click.argument('search_id')
 async def search_get(ctx, search_id, pretty):
-    """ This function obtsains an existing saved search, using the search_id.
+    """ Get saved search.
```
sgillies, line 145: This is only a nit, but there is no extra whitespace needed before the first word. In fact, I think this could make `search-get --help` look strangely formatted if click didn't strip the leading space.

```suggestion
    """Get saved search.
```

sgillies (commented):
@mkshah605 looks good, I only found one tiny nit.

jreiberkyle (approved):
looks great! good to go

tests/integration/test_data_api.py
```diff
@@ -320,7 +350,6 @@ async def test_delete_search(retcode, expectation, session):
 
     assert route.called
```
jreiberkyle, line 352:
```suggestion


```

PR #579, opened 2022-06-03
 fix checksum validation bug, refactor checksum validation

### LINKED ISSUE #521: Refactor download checksum validation
Before we finalize the Python API for 2.0, I'd like to separate the download validation, where we compare asset checksums to values in their accompanying manifest, from the orders client so that it's more generally available to users. There are use cases that don't need an orders client, such as re-validating the data after copying it to another computer.

jreiberkyle:
Checksum validation was failing on tests with actual orders due to the manifest.json file having a different format than what was created in the tests (I should have double-checked that). This fixes that bug. Testing all cases was difficult with checksum validation embedded in `download_order` function, so validation was moved to it's own separate function. With this change, Python API users must call `validate_checksum` separately. This is added to the API User Guide doc.

Specifics:
- `validate_checksum` as a function separate from `download_order` reduces the complexity of testing (soo many permutations when they were combined)
- `validate_checksum` does not require any state information from `OrdersClient`, but it was kept in `OrdersClient` as a static method because `OrdersClient` is a top-level import and the orders module is a bit harder to get to. To use `validate_checksum`, an `OrdersClient` does not need to be instantiated - it can be called with `OrdersClient.validate_checksum`
- checksum cli test added

Closes #521

### DIFF
--- a/docs/guide.md
+++ b/docs/guide.md
@@ -221,6 +221,27 @@ from planet import reporting
 >>> asyncio.run(create_poll_and_download())
 ```
 
+#### Validating Checksums
+
+Checksum validation provides for verification that the files in an order have
+been downloaded successfully and are not missing, currupted, or changed. This
+functionality is included in the OrderClient, but does not require an instance
+of the class to be used.
+
+
+To perform checksum validation:
+
+```python
+from pathlib import Path
+
+# path includes order id
+order_path = Path('193e5bd1-dedc-4c65-a539-6bc70e55d928')
+OrdersClient.validate_checksum(order_path, 'md5')
+```
+
+
+
+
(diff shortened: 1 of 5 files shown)


jreiberkyle:
@cholmes - this doesn't change CLI interface, but does change Python API interface - feel free to focus on the doc change to give feedback

@sgillies - this is a little different than what you proposed for refactoring checksum validation in that `validate_checksum` is kept as a static method of `OrdersClient`. This decision was made for ease of import by the user. I'm open to thoughts on this approach.

jreiberkyle:
merging to get the fix in. happy to revisit the python api interface and structure in the future.

Issue #488, opened 2022-05-04
implement cli command: planet data stats


jreiberkyle:
implement `planet data stats`

## Interface

planet data stats [OPTIONS] ITEM_TYPES INTERVAL FILTER

Get a bucketed histogram of items matching the filter.

Arguments:
ITEM_TYPES - string. Comma-separated item type identifier(s).
INTERVAL - string. The size of the histogram buckets (<hour, day, week, month, year>)
FILTER - string. A full JSON description filter. Supports file and stdin.

Options:
--utc-offset - string. A "ISO 8601 UTC offset" (e.g. +01:00 or -08:00) that can be used to adjust the buckets to a users time zone.

Output:
A full JSON description of the returned statistics result.

## Usage Examples
User Story: As a CLI user I would like to get the stats of multiple item types with a utc offset using a file to specify the filter.

$ planet data stats --utc-offset +1:00 \
PSScene,PSOrthoTile day filter-desc.json

<unedited API response json>

PR #585, opened 2022-06-07
CLI data stats #488

mkshah605:


mkshah605:
I didn't see any reference to the `utc_offset` in the data client file, so I added that in. @jreiberkyle do you have recommendations on how to actually implement the offset so that it shows the results with the actual offset in the printed timezone?

mkshah605:
Update: @sgillies and I discussed this today and it might be a good idea to clarify some things: how does the filter that we specify interact with the utc_offset? Currently, we are able to specify UTC times in the filter, but does it make sense that we are still able to do an offset on this value, and receive a result back in, still in UTC time? Would love to discuss this whenever you're back @jreiberkyle . 

Also, in the CLI docs I changed the documentation around how to specify the utc_offset to match the current API behavior. Example: "+1h" instead of "+01:00h". We should confirm with the API team that this is how it's intended, and update the API docs accordingly too if that's the case.

jreiberkyle:
it looks like this PR review got stuck on the offset issue. I will put this on the top of my todo list for Tuesday and see if we can get this resolved and the PR closed

planet/cli/data.py
```diff
@@ -37,9 +37,24 @@ def data(ctx, base_url):
 
 def parse_item_types(ctx, param, value: str) -> List[str]:
     """Turn a string of comma-separated names into a list of names."""
-    # Note: we could also normalize case and validate the names against
-    # our schema here.
-    return [part.strip() for part in value.split(",")]
+    # convert user-input strings to match our naming schema
+    update_value = value.lower()
+
+    dict = {
```
jreiberkyle, line 43: Two things here:
1. we want the CLI to be as thin as possible, that means, any validation and adjusting of case is something we want to do in the API, if possible. 
2. We want to avoid hard-coding values that derive from the API schema and instead pull them from the schema itself. See how this is done in the `specs.py` module, which is used by `order_requests.py`

For this effort, I would suggest creating a ticket for validation and normalization of item_types in the API and I would go without this logic in this PR.

planet/cli/data.py
```diff
@@ -135,6 +150,41 @@ async def search_create(ctx, name, item_types, filter, daily_email, pretty):
...
+@click.pass_context
+@translate_exceptions
+@coro
+@click.argument("item_types", callback=parse_item_types)
+@click.argument('interval',
+                default=None,
+                type=click.Choice(['hour', 'day', 'week', 'month', 'year'],
+                                  case_sensitive=False))
+@click.argument("filter", callback=parse_filter)
+@click.option('--utc_offset',
```
jreiberkyle, line 163: let's remove this option for now

planet/clients/data.py
```diff
@@ -327,7 +327,8 @@ async def run_search(
     async def get_stats(self,
                         item_types: typing.List[str],
                         search_filter: dict,
-                        interval: str) -> dict:
+                        interval: str,
+                        utc_offset: str = None) -> dict:
```
jreiberkyle, line 331: let's remove this parameter for now

planet/clients/data.py
```diff
@@ -336,7 +337,7 @@ async def get_stats(self,
             interval: The size of the histogram date buckets.
 
         Returns:
-            Returns a date bucketed histogram of items matching the filter.
+            A full JSON description of the returned statistics result.
```
jreiberkyle, line 340: What is the reasoning behind this change? The prior description seemed more, well, descriptive.

tests/integration/test_data_cli.py
```diff
@@ -218,6 +219,100 @@ def test_search_create_daily_email(invoke, search_result):
     assert json.loads(result.output) == search_result
 
 
+@respx.mock
+@pytest.mark.asyncio
+@pytest.mark.parametrize("filter", ['{1:1}', '{"foo"}'])
```
jreiberkyle, line 224: I am loving seeing these parametrizations

tests/integration/test_data_cli.py
```diff
@@ -218,6 +219,100 @@ def test_search_create_daily_email(invoke, search_result):
...
+@pytest.mark.parametrize(
```
jreiberkyle, line 225: how does changing item_types apply to what is being tested here?

tests/integration/test_data_cli.py
```diff
@@ -218,6 +219,100 @@ def test_search_create_daily_email(invoke, search_result):
...
+        ["stats", item_types, interval, filter, "--utc_offset", utc_offset],
+        runner=runner)
+    assert result.exit_code == 2
+
+
+@respx.mock
+@pytest.mark.parametrize(
+    "item_types", ['PSScene', 'SkySatScene', ('PSScene', 'SkySatScene')])
+@pytest.mark.parametrize("interval", ['hou', 'da', 'wek', 'moth', 'yr'])
+def test_data_stats_invalid_interval(invoke, item_types, interval):
```
jreiberkyle, line 248: you could actually parameterize this so that an interval is invalid, and you expect a return code of 2, and an interval is valid, and you expect a return code of 0. I'm not sure how testing multiple invalid intervals is beneficial vs testing one invalid interval.

tests/integration/test_data_cli.py
```diff
@@ -218,6 +219,100 @@ def test_search_create_daily_email(invoke, search_result):
...
+                    runner=runner)
+
+    assert result.exit_code == 2
+
+
+@respx.mock
+@pytest.mark.parametrize(
+    "item_types", ['PSScene', 'SkySatScene', ('PSScene', 'SkySatScene')])
+@pytest.mark.parametrize("interval", ['hour', 'day', 'week', 'month', 'year'])
+@pytest.mark.parametrize("utc_offset", ['+1h', '-4h', '+10h', '-6h', '-3.5h'])
```
jreiberkyle, line 285: let's get rid of utc offset here

tests/integration/test_data_cli.py
```diff
@@ -218,6 +219,100 @@ def test_search_create_daily_email(invoke, search_result):
...
+        runner=runner)
+    assert result.exit_code == 2
+
+
+@respx.mock
+@pytest.mark.parametrize(
+    "item_types", ['PSScene', 'SkySatScene', ('PSScene', 'SkySatScene')])
+@pytest.mark.parametrize("interval", ['hou', 'da', 'wek', 'moth', 'yr'])
+def test_data_stats_invalid_interval(invoke, item_types, interval):
+    """Test for planet data search_create. Test with multiple item_types.
```
jreiberkyle, line 249: this docstring seems inaccurate

tests/integration/test_data_cli.py
```diff
@@ -218,6 +219,100 @@ def test_search_create_daily_email(invoke, search_result):
...
+        }
+    }
+
+    mock_resp = httpx.Response(HTTPStatus.OK,
+                               json={'features': [{
+                                   "key": "value"
+                               }]})
+    respx.post(TEST_STATS_URL).return_value = mock_resp
+
+    utc_offset = "+1h"
```
jreiberkyle, line 265: we don't need to specify optional options that are unrelated to what is being tested.

jreiberkyle (changes requested):
High-level, it would be great for a PR to have a description that provides what is being done, why, and outlines the changes. Also, thanks in part to this PR and the subsequent discussion, we are removing utc_offset for now. And validation against specs is a bigger effort and can be ticketed and left out of this PR.

planet/clients/data.py
mkshah605: @jreiberkyle I added the word "histogram" to this for context, but this was changed here to clarify that it is returned in JSON form. Generally I understand "histogram" to indicate a graphical/chart representation so this is just for clarification that it is a description of one, not an actual histogram.

tests/integration/test_data_cli.py
mkshah605: you're right, it doesn't! I had just included all of the options just to test them but that's already being tested elsewhere so we don't need it.

tests/integration/test_data_cli.py
mkshah605: indeed it is

mkshah605:
finally getting a chance to get to these updates. Thank you @jreiberkyle !

planet/cli/data.py
jreiberkyle: I am not seeing the resolution of these issues. Could you address these issues by either removing the validation code and creating/including a reference to the ticket or adding an explanation on how these issues have been resolved in another way?

planet/cli/data.py
```diff
@@ -48,6 +49,52 @@ def data(ctx, base_url):
...
+        "landsat8l1g": "Landsat8L1G",
+        "sentinel2l1c": "Sentinel2L1C"
+    }
+    for original, validated in dict.items():
+        update_value = value.replace(original, validated)
+
+    return [part.strip() for part in update_value.split(",")]
+
+
+def parse_filter(ctx, param, value: str) -> dict:
```
jreiberkyle, line 72: this is new since before the PR was created, but now you can remove this logic and just use the `type.JSON` type. See how it is done in [search-create](https://github.com/planetlabs/planet-client-python/blob/f376018877daf845d3666359082475a25baf9f47/planet/cli/data.py#L272) for an example

planet/cli/data.py
```diff
@@ -295,6 +342,32 @@ async def search_create(ctx, name, item_types, filter, daily_email, pretty):
         echo_json(items, pretty)
 
 
+@data.command()
+@click.pass_context
+@translate_exceptions
+@coro
+@click.argument("item_types", callback=parse_item_types)
```
jreiberkyle, line 349: you can remove `parse_item_types` and just use `types.CommaSeparatedString`, see [search-create](https://github.com/planetlabs/planet-client-python/blob/f376018877daf845d3666359082475a25baf9f47/planet/cli/data.py#L271) for an example.

planet/cli/data.py
```diff
@@ -295,6 +342,32 @@ async def search_create(ctx, name, item_types, filter, daily_email, pretty):
...
+@click.argument('interval',
+                default=None,
+                type=click.Choice(['hour', 'day', 'week', 'month', 'year'],
```
jreiberkyle, line 352: instead of defining this list again, import and use [`STATS_INTERVAL`](https://github.com/planetlabs/planet-client-python/blob/f376018877daf845d3666359082475a25baf9f47/planet/clients/data.py#L35)

planet/cli/data.py
```diff
@@ -295,6 +342,32 @@ async def search_create(ctx, name, item_types, filter, daily_email, pretty):
         echo_json(items, pretty)
 
 
+@data.command()
+@click.pass_context
+@translate_exceptions
+@coro
+@click.argument("item_types", callback=parse_item_types)
+@click.argument('interval',
+                default=None,
```
jreiberkyle, line 351: I believe cl.get_stats will fail if stats_interval is None. I recommend removing the default value. Is there a test for the CLI that would tell us if not specifying interval would result in an error or not?

jreiberkyle (changes requested):
It is great to see this getting some traction again! I have included some requested changes. Also, would you be willing to include in this PR an update to design_docs/CLI_data.md that removes the UTC option? Thanks!

planet/cli/data.py
```diff
@@ -295,6 +296,29 @@ async def search_create(ctx, name, item_types, filter, daily_email, pretty):
...
+@click.argument("item_types", type=types.CommaSeparatedString())
+@click.argument('interval', type=click.Choice(STATS_INTERVAL))
+@click.argument("filter", type=types.JSON(), default="-", required=False)
+async def stats(ctx, item_types, interval, filter):
+    """Get a bucketed histogram of items matching the filter.
+
+    This function returns a bucketed histogram of results based on the
+    item_types, interval, and json filter specified (using file or stdin).
+    This function outputs a full JSON description of the returned statistics
+    result.
```
jreiberkyle, line 312: This seems to read like two different descriptions for the function response (aka "returns" and "outputs"). Maybe delete one of them?

tests/integration/test_data_cli.py
```diff
@@ -582,6 +583,80 @@ def test_search_create_daily_email(invoke, search_result):
...
+    item_type = 'PSScene'
+    runner = CliRunner()
+    result = invoke(["stats", item_type, interval, filter], runner=runner)
+    assert result.exit_code == 2
+
+
+@respx.mock
+@pytest.mark.parametrize(
+    "item_types", ['PSScene', 'SkySatScene', ('PSScene', 'SkySatScene')])
+@pytest.mark.parametrize("interval, exit_code", [(None, 1), ('hou', 2),
```
jreiberkyle, line 607: beautiful parametrization scheme

jreiberkyle (approved):
This looks excellent @mkshah605 ! Well done! I only found one little opportunity to clarify docs. Great work!

PR #589, opened 2022-06-13
Environment variable credentials above secret file credentials

### LINKED ISSUE #574: Allow SDK to automatically generate Auth from environment variable
In the V1 SDK, if the environment variable `PL_API_KEY` contained a valid API key, it was automatically used by the SDK (and CLI) for authentication.

It would be nice if this behavior was retained in V2, such that I could instantiate a `Session()` without needing to first initialize and `write()` an `Auth` object if this environment variable was already defined

sgillies:
Like, for example, AWS CLI and v1 of the Python client.

Resolves #574

### DIFF
--- a/planet/cli/data.py
+++ b/planet/cli/data.py
@@ -5,7 +5,6 @@ from typing import List
 from contextlib import asynccontextmanager
 
 import click
-import planet
 from planet import DataClient, Session
 
 from .cmds import coro, translate_exceptions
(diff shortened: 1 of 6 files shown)


tbarsballe (commented):
Looks good to me.
I pulled this branch into the project I was testing SDK v2 with previously, and was able to remove the `planet.Auth.from_env("PL_API_KEY").write()` without issue, seems to be picking up API key auth as expected

PR #607, opened 2022-06-27
Subscriptions API client

### LINKED ISSUE #339: Change v2 X-Planet-App header to reflect SDK
Related: #336 

X-Planet-App is used internally by Planet; the list of allowed values is limited. Work with the team to determine new value for SDK requests (and, potentially, for CLI vs SDK requests)

### LINKED ISSUE #512: implement subscriptions api & CLI create/update functions
implement the following subscriptions api & CLI functions:
* create_subscription
* update_subscription

CLI specified in https://github.com/planetlabs/planet-client-python/blob/main/design-docs/CLI-Subscriptions.md#subscriptions-command-line-interface-specification

### LINKED ISSUE #513: implement subscriptions api & CLI results function
implement the following subscriptions api function:
* get_results

CLI specified in https://github.com/planetlabs/planet-client-python/blob/main/design-docs/CLI-Subscriptions.md#subscriptions-command-line-interface-specification

### LINKED ISSUE #514: implement subscriptions api & CLI cancel/get/list functions
implement the following subscriptions api functions:
* get_subscription
* cancel_subscription
* list_subscriptions

CLI specified in https://github.com/planetlabs/planet-client-python/blob/main/design-docs/CLI-Subscriptions.md#subscriptions-command-line-interface-specification

sgillies:
Resolves #512
Resolves #513
Resolves #514
Resolves #339

### DIFF
--- a/planet/cli/session.py
+++ b/planet/cli/session.py
@@ -1,0 +1,11 @@
+"""CLI HTTP/auth sessions."""
+
+from planet.auth import AuthType
+from planet.http import Session
+
+
+class CliSession(Session):
+
+    def __init__(self, auth: AuthType = None):
+        super().__init__(auth)
+        self._client.headers.update({'X-Planet-App': 'python-cli'})
--- a/planet/cli/subscriptions.py
+++ b/planet/cli/subscriptions.py
@@ -35,7 +35,7 @@ def subscriptions(ctx):
               default=100,
               help='Maximum number of results to return. Defaults to 100.')
 @click.pass_context
-# @translate_exceptions
+@translate_exceptions
 @coro
 async def list_subscriptions_cmd(ctx, status, limit, pretty):
     """Prints a sequence of JSON-encoded Subscription descriptions."""
--- a/planet/cli/subscriptions.py
+++ b/planet/cli/subscriptions.py
@@ -4,9 +4,9 @@ import json
 
 import click
 
-from planet import Session
 from planet.cli.cmds import coro, translate_exceptions
 from planet.cli.io import echo_json
+from planet.cli.session import CliSession
 from planet.clients.subscriptions import SubscriptionsClient
 
 
@@ -39,7 +39,7 @@ def subscriptions(ctx):
 @coro
 async def list_subscriptions_cmd(ctx, status, limit, pretty):
     """Prints a sequence of JSON-encoded Subscription descriptions."""
-    async with Session(auth=ctx.obj['AUTH']) as session:
+    async with CliSession(auth=ctx.obj['AUTH']) as session:
         client = SubscriptionsClient(session)
         filtered_subs = client.list_subscriptions(status=status, limit=limit)
         async for sub in filtered_subs:
@@ -75,7 +75,7 @@ def parse_request(ctx, param, value: str) -> dict:
 @coro
 async def create_subscription_cmd(ctx, request, pretty):
     """Submits a subscription request and prints the API response."""
-    async with Session(auth=ctx.obj['AUTH']) as session:
+    async with CliSession(auth=ctx.obj['AUTH']) as session:
         client = SubscriptionsClient(session)
         sub = await client.create_subscription(request)
         echo_json(sub, pretty)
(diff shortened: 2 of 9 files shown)


planet/cli/subscriptions.py
```diff
@@ -35,7 +35,7 @@ def subscriptions(ctx):
               default=100,
               help='Maximum number of results to return. Defaults to 100.')
 @click.pass_context
-@translate_exceptions
+# @translate_exceptions
```
sgillies, line 38: TODO: restore.

tests/integration/test_subscriptions_api.py
```diff
@@ -1,59 +1,143 @@
...
+sub_api_mock.route(
+    M(url__startswith='https://api.planet.com/subscriptions/v1'),
+    M(method__in=['GET'])).mock(
+        return_value=Response(200,
+                              json={
+                                  'id': '42',
+                                  'name': 'test',
+                                  'delivery': 'yes, please',
+                                  'source': 'test'
+                              }))
```
sgillies, line 98: This is most of the work: creating the mock router configurations that allow us to test features.

planet/cli/session.py
```diff
@@ -0,0 +1,11 @@
...
+
+from planet.auth import AuthType
+from planet.http import Session
+
+
+class CliSession(Session):
+
+    def __init__(self, auth: AuthType = None):
+        super().__init__(auth)
+        self._client.headers.update({'X-Planet-App': 'python-cli'})
```
sgillies, line 11: We might consider making this header a class attribute @jreiberkyle ? Then we wouldn't need to override the constructor like I'm doing here. Not a big deal, though.

planet/clients/subscriptions.py
```diff
@@ -1,87 +1,89 @@
...
 
+    # This constructor overrides Paged's and takes an httpx
+    # async client as the first argument.
+    def __init__(self, client, request, limit=None):
+        self.request = request
+        self.client = client
+        self._pages = None
+        self._items = []
+        self.i = 0
+        self.limit = limit
```
sgillies, line 26: @jreiberkyle I subclassed Paged while developing this feature to reduce layers of abstraction and make things easier to debug. I do not mind at all if this extra class is deleted on the way to enabling automatic retries (if we need that).

I *do* think the Paged API would be cleaner if the constructor took a client (as I did) or session object as an argument instead of a "request doer". So maybe that could be something that comes out of this PR into other modules?

planet/clients/subscriptions.py
```diff
@@ -97,18 +99,27 @@ async def list_subscriptions(self,
...
+        url = URL('https://api.planet.com/subscriptions/v1', params=params)
+        req = self.session._client.build_request('GET', url)
+
         try:
-            # TODO: replace with httpx request.
-            async for sub in _server_subscriptions_get(status=status,
-                                                       limit=limit):
+            async for sub in _SubscriptionsPager(self.session._client,
+                                                 req,
+                                                 limit=limit):
```
sgillies, line 115: @jreiberkyle I've made the subscriptions pager an implementation detail here. I can't think of a reason why Paged needs to be part of this library's public API. Other than the async iterator magic methods its methods are all private. We could make the class itself private and reduce our API surface area and documentation needs. What do you think?

planet/clients/subscriptions.py
```diff
@@ -174,17 +191,20 @@ async def update_subscription(self, subscription_id: str,
...
-            raise APIError("Subscription failure.") from server_error
+            resp = await self.session._client.send(req)
+        # Forward APIError. We don't strictly need this clause, but it
+        # makes our intent clear.
+        except APIError:
+            raise
+        except ClientError:  # pragma: no cover
+            raise
         else:
+            sub = resp.json()
```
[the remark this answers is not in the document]
sgillies: The retry logic in http.Session is something I would want to use for sure, but I would really like to use it differently. As a standalone utility. Think of it as an inline decorator.

```
url = URL(f'https://api.planet.com/subscriptions/v1/{subscription_id}')
req = self.session._client.build_request('PUT', url, json=request)
resp = await retry(retry_config, self.session._client.send, req)
return resp.json()
```

PR #622, opened 2022-06-30
limit option: remove None as a possibility, value of zero means no limit

jreiberkyle:


planet/clients/subscriptions.py
```diff
@@ -101,17 +59,17 @@ async def list_subscriptions(self,
             ClientError: on a client error.
         """
 
-        class _SubscriptionsPager(_Pager):
+        class _SubscriptionsPager(Paged):
```
sgillies, line 62: @jreiberkyle I accidentally pushed to the the wrong branch. I will undo this last commit now.

jreiberkyle:
thanks @sgillies for the quick fix of the blocker

PR #625, opened 2022-06-30
Fixed code block bug.

kevinlacaille:
Fixed a code block bug, which added the `Installing from source` section in a code block.


cholmes (approved):
Thanks!

PR #627, opened 2022-06-30
reuse code, break tight coupling

jreiberkyle:


planet/clients/subscriptions.py
```diff
@@ -73,7 +43,22 @@ class SubscriptionsClient:
...
+    def _request(self, url, method, data=None, params=None, json=None):
+        return Request(url, method=method, data=data, params=params, json=json)
+
+    async def _do_request(self, request: Request) -> Response:
+        """Submit a request and get response.
+
+        Parameters:
+            request: request to submit
+        """
+        return await self._session.request(request)
```
sgillies, line 61: @jreiberkyle in #629 I found that _request and _do_request aren't needed. We can use the bodies of the functions instead.

sgillies:
@jreiberkyle @sarasafavi @cholmes I have merged #629, which unblocks #622.

Issue #580, opened 2022-06-03
Persistent failure when submitting many API requests concurrently


tbarsballe:
When submitting many (1000+) quick-search requests concurrently using `asyncio.gather`, they consistently fail with a `planet.exceptions.TooManyRequests: max rate reached: retry-in 200ms` error. 

I have also occasionally seen a `planet.exceptions.ServerError: Internal Server Error` instead, but only when using a shared session among all requests; using a unique session for each request seems to only produce the TooManyRequests error.

This failure is consistent, and happens every time I attempt this. 

Stacktrace of the TooManyRequests error is:

```
unhandled exception during asyncio.run() shutdown
task: <Task finished name='Task-111' coro=<do_searches.<locals>.do_search_geom() done, defined at /Users/torben.barsballe/repos/Planet/order-util/order_aois.py:108> exception=CloseError(SSLError(1, '[SSL: APPLICATION_DATA_AFTER_CLOSE_NOTIFY] application data after close notify (_ssl.c:2756)'))>
Traceback (most recent call last):
  File "/Users/torben.barsballe/repos/Planet/order-util/order_aois.py", line 113, in do_search_geom
    itms = await do_search(start, end, geom)
  File "/Users/torben.barsballe/repos/Planet/order-util/order_aois.py", line 97, in do_search
    items.append(item)
  File "/Users/torben.barsballe/repos/Planet/order-util/venv-v2/lib/python3.9/site-packages/httpcore/_async/connection.py", line 183, in aclose
    await self.connection.aclose()
  File "/Users/torben.barsballe/repos/Planet/order-util/venv-v2/lib/python3.9/site-packages/httpcore/_async/http11.py", line 202, in aclose
    await self.socket.aclose()
  File "/Users/torben.barsballe/repos/Planet/order-util/venv-v2/lib/python3.9/site-packages/httpcore/_backends/asyncio.py", line 198, in aclose
    await self.stream_writer.wait_closed()  # type: ignore
  File "/Library/Frameworks/Python.framework/Versions/3.9/lib/python3.9/contextlib.py", line 137, in __exit__
    self.gen.throw(typ, value, traceback)
  File "/Users/torben.barsballe/repos/Planet/order-util/venv-v2/lib/python3.9/site-packages/httpcore/_exceptions.py", line 12, in map_exceptions
    raise to_exc(exc) from None
[...]

Issue #578, opened 2022-06-03
Occassional ReadTimeout when submitting multiple quick-search requests in sequence


tbarsballe:
When submitting a large number (1000+) of quick-search requests in sequence, one will occasionally fail with a `httpx.ReadTimeout`. This only happens sometimes, and is not readily reproducible. It can be resolved by wrapping in a try-catch, waiting, and retrying.

Full stacktrace follows:

```
Traceback (most recent call last):
  File "/Users/torben.barsballe/repos/Planet/order-util/venv-v2/lib/python3.9/site-packages/httpx/_exceptions.py", line 326, in map_exceptions
    yield
  File "/Users/torben.barsballe/repos/Planet/order-util/venv-v2/lib/python3.9/site-packages/httpx/_client.py", line 1502, in _send_single_request
    (status_code, headers, stream, ext,) = await transport.arequest(
  File "/Users/torben.barsballe/repos/Planet/order-util/venv-v2/lib/python3.9/site-packages/httpcore/_async/connection_pool.py", line 218, in arequest
    response = await connection.arequest(
  File "/Users/torben.barsballe/repos/Planet/order-util/venv-v2/lib/python3.9/site-packages/httpcore/_async/connection.py", line 106, in arequest
    return await self.connection.arequest(method, url, headers, stream, ext)
  File "/Users/torben.barsballe/repos/Planet/order-util/venv-v2/lib/python3.9/site-packages/httpcore/_async/http11.py", line 72, in arequest
    ) = await self._receive_response(timeout)
  File "/Users/torben.barsballe/repos/Planet/order-util/venv-v2/lib/python3.9/site-packages/httpcore/_async/http11.py", line 133, in _receive_response
    event = await self._receive_event(timeout)
  File "/Users/torben.barsballe/repos/Planet/order-util/venv-v2/lib/python3.9/site-packages/httpcore/_async/http11.py", line 172, in _receive_event
    data = await self.socket.read(self.READ_NUM_BYTES, timeout)
  File "/Users/torben.barsballe/repos/Planet/order-util/venv-v2/lib/python3.9/site-packages/httpcore/_backends/asyncio.py", line 157, in read
    raise
  File "/Library/Frameworks/Python.framework/Versions/3.9/lib/python3.9/contextlib.py", line 137, in __exit__
    self.gen.throw(typ, value, traceback)
[...]

PR #640, opened 2022-07-18
Increase reliability of communication with Planet servers and avoid clobbering them

### LINKED ISSUE #551: Handle additional rate-limit failure modes with exponential backoff.
#279 addressed handling HTTP 429 TOO_MANY_REQUESTS failures through use of retry with exponential backoff.

Use of the Data and Orders APIs have shown additional failure modes in rate-limiting situations beyond just the HTTP 429 response, that should be handled in the same way. These include timeouts and partial responses, which may occur when many (1000+) API requests are submitted all at once, and the scaling and rate-limiting mechanisms of the APIs fail to catch them all, and may manifest as the following exceptions (assuming use of the requests library):

- requests.exceptions.ChunkedEncodingError
- requests.exceptions.ReadTimeout

In addition to these failures, occasionally 4xx or 5xx responses are returned with a blank or corrupt body, such that attempting to parse the response body results in some form of decode error (e.g. `json.decoder.JSONDecodeError` when calling `response.json()`) - this appears to be another manifestation of the above failure.

I have also observed intermittent failures specifically when submitting orders, where I receive a 5xx status with the message "Problem confirming products". This failure may also be related to rate-limiting, as a resubmission of the same order always appears to succeed. I am less certain as to whether this particular failure should be handled through the same method, as there may be cases where it is actually a relevant response.

jreiberkyle:
This introduces three major changes to communication with the Planet services to increase reliability of the communication and avoid clobbering the services.

1. Every request sent by a Session is now limited according to maximum workers and rate_limit, imposed by the new `_Limiter` class.
2. Increases the set of exceptions that trigger a (exponential-backoff) retry.
3. Increases read timeout.

Additionally, this adds the script that was used to identify, explore, and test the configuration and populate the exception list. 

Results from testing and findings that led to configuration settings and retry exception identification can be found in #580. 

This also closes #551 and #578.

TODO checklist:

- [x] retest final settings with script
- [x] document in http how settings were obtained and how to rerun script to tweak them if they aren't working for another endpoint

### DIFF
--- a/planet/http.py
+++ b/planet/http.py
@@ -21,6 +21,7 @@ import logging
 import random
 import time
 
+import httpcore
 import httpx
 
 from .auth import Auth, AuthType
@@ -28,6 +29,7 @@ from . import exceptions, models
 from .__version__ import __version__
 
 RETRY_EXCEPTIONS = [
+    httpcore.CloseError,  # this is actually a httpx bug, will go away with a httpx upgrade - see https://github.com/encode/httpcore/pull/310
     httpx.ConnectError,
     httpx.ReadError,
     httpx.ReadTimeout,
--- a/planet/http.py
+++ b/planet/http.py
@@ -135,9 +135,9 @@ class _Limiter:
 
         self.retry_interval = 0.01
 
-        # State
+        # track state
         self._running = 0
-        self._last_call = self._get_now() - self.cadence
+        self._last_call = None
 
     @staticmethod
     def _get_now():
@@ -147,9 +147,13 @@ class _Limiter:
         if self.cadence:
             while True:
                 now = self._get_now()
-                if now - self._last_call > self.cadence:
+                if self._last_call is None:
+                    # first call, no need to throttle
+                    self._last_call = now
+                    break
+                elif now - self._last_call >= self.cadence:
                     LOGGER.debug(
-                        f'Throught throttle, delta: {now-self._last_call}')
+                        f'Throught throttle, delta: {now - self._last_call}')
                     self._last_call = now
                     break
                 await asyncio.sleep(self.retry_interval)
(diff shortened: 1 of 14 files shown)


tests/unit/test_http.py
```diff
@@ -71,6 +71,30 @@ def test_basesession__raise_for_status(mock_response):
         http.BaseSession._raise_for_status(
             mock_response(HTTPStatus.METHOD_NOT_ALLOWED, json={}))
 
+import asyncio
+import time
+@pytest.mark.asyncio
+async def test__Limiter_rate_limit():
```
jreiberkyle, line 77: @sgillies any ideas for how we can test the rate/worker limiting? Its a tricky async/timing problem

planet/http.py
```diff
@@ -15,19 +15,39 @@
...
+import httpcore
 import httpx
 
 from .auth import Auth, AuthType
 from . import exceptions, models
 from .__version__ import __version__
 
+# For how this list was determined, see
+# https://github.com/planetlabs/planet-client-python/issues/580
+RETRY_EXCEPTIONS = [
```
jreiberkyle, line 33: @cholmes see extra exceptions caught here

planet/http.py
```diff
@@ -99,10 +99,27 @@ def _raise_for_status(cls, response):
...
+    close to eachother (even though max rate limit is 5 calls per second)[2].
+
+    In investigating options, aiometer[3] was also looked at but it seems to
+    have odd behavior with httpx [4].
+
+    References:
+    [1] https://github.com/mjpieters/aiolimiter
+    [2] https://github.com/planetlabs/planet-client-python/issues/580#issuecomment-1182752851
+    [3] https://github.com/florimondmanca/aiometer
+    [4] https://github.com/florimondmanca/aiometer/issues/24
```
sgillies, line 120: References and design notes are great! This is a good example for projects to follow.

tests/unit/test_http.py
sgillies: You mean testing the cadence? How about wrapping the limiter around calls to time.monotonic, collecting N number of times, and then asserting that the N calls don't exceed the Limiters rate limit and that no calls are closer together than the cadence threshold? The limiter can throttle anything, not only HTTP requests, right? 

```python
async def get_time_limited(limiter):
    async with limiter:
        return time.monotonic()

limiter = Limiter(...)
times = [get_time_limited(limiter) async for i in range(100)]
assert all(second - first > threshold for first, second in itertools.pairwise(times))
```

Is that useful?

jreiberkyle: yeah, it can limit anything! This is a cool approach and is helpful.

planet/http.py
jreiberkyle: thanks! I agree wholeheartedly

jreiberkyle:
@sgillies @mkshah605 @kevinlacaille and @tbarsballe FYI.

PR #641, opened 2022-07-18
CLI search-quick -> search, API quick-search -> search

### LINKED ISSUE #634: rename `search-quick` to `search` in CLI
In v1 the quick search command was just called `search`. In v2 we tried out being a bit more true to the API's, but it really is the most used command, and so gets annoying to continue to type out. So let's just have it be `search` in the CLI.

Note I'm fine either way with the python api, to have it remain `search-quick` or to also change it to `search`.

jreiberkyle:
Closes #634

### DIFF
--- a/design-docs/CLI-Data.md
+++ b/design-docs/CLI-Data.md
@@ -204,11 +204,11 @@ response (pretty-printed)
 }
 ```
 
-## search-quick
+## search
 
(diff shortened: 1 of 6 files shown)


cholmes (approved):
Looks good to me. Thanks!

PR #650, opened 2022-07-25
Better cli info item types 572

kevinlacaille:
Created a working solution, but debating whether it should exist in `planet.cli.data` (current home), in `planet.clients.data.DataClient.get_item_type()`

### DIFF
--- a/planet/cli/data.py
+++ b/planet/cli/data.py
@@ -19,7 +19,7 @@ import click
 
 from planet import data_filter, DataClient
 from planet.clients.data import SEARCH_SORT, SEARCH_SORT_DEFAULT, STATS_INTERVAL
-from planet.specs import get_item_types
+from planet.specs import get_item_types, get_product_bundles
 
 from . import types
 from .cmds import coro, translate_exceptions
@@ -27,7 +27,11 @@ from .io import echo_json
 from .options import limit, pretty
 from .session import CliSession
 
-ITEM_TYPES = get_item_types('analytic')
+BUNDLES = get_product_bundles()
+all_item_types = []
+for bundle in BUNDLES:
+    all_item_types += [*get_item_types(bundle)]
+ITEM_TYPES = set(all_item_types)
 
 
 @asynccontextmanager
@@ -233,7 +237,7 @@ def filter(ctx,
 @click.pass_context
 @translate_exceptions
 @coro
-@click.argument("item_types", type=types.CommaSeparatedString())
+@click.argument('item_types', type=click.Choice(ITEM_TYPES))
 @click.argument("filter", type=types.JSON(), default="-", required=False)
 @limit
 @click.option('--name', type=str, help='Name of the saved search.')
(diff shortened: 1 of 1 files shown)


kevinlacaille:
Almost done, except:
-  `planet data search item-type <some invalid item type` doesn't give me the error message I'd like
**Expected:** Raised `NoMatchException` from `planet.specs.get_match()`
**Returned:** An error message from `click.ClickException()`, which routes to `click.exceptions.ClickException.show()`, which gets the message from `click.exceptions.BadParameter.format_message()`
- Need to write tests for:
    - `get_match()` for new exception message
    - `get_item_types()` for the new case where `product_bundle` is empty 
    - `planet.cli.data.search(), stats(), search_create()` for new epilog message

kevinlacaille:
**planet.data.cli.py**
- Added epilog to all commands that use `item_types` (`search()`, `search-create()`, and `stats()`) to tell user which item types are valid inputs.
    - Added a callback function, `check_item_types()`, to raise a `BadParameter` exception when an invalid item type is given, with a helpful message.

**planet.specs.py**
- Made `get_item_types()` more flexible. Namely, allowed it to return all item types supported by Order API when given no product bundle. 
    - **NOTE:** In the future, we should get the item types from the Data API spec (see #663)
- Added helpful a message to a `NoMatchException` exception.

**tests.integration.test_data_cli.py**
- Added a test for the `planet data search` command. Specifically, this tests what the epilog returns.
- Fixed all the erroneous tests that had item types placed in tuples, not comma separated strings (not sure why these tests were passing in the first place).

**tests.unit.test_data_item_type.py**
- Created this new module to test a new callback function in the Data CLI, `check_item_types()`. Huge thanks to @sgillies and rasterio’s tests for the inspiration!

**tests.unit.test_specs.py**
- Changed `test_get_item_types()` to test for a given product bundle.
- Added a test to test `get_item_types()` without a given product bundle.

**FYI:** @jreiberkyle @cholmes

tests/unit/test_data_item_type.py
```diff
@@ -0,0 +1,55 @@
...
+                             'skysatcollect',
+                             'mod09ga',
+                             'psscene3band',
+                             'mod09gq',
+                             'psscene4band',
+                             'psscene'
+                         ])
+def test_item_type_success(item_type):
+    ctx = MockContext()
+    with pytest.raises(click.BadParameter):
```
sgillies, line 48: @kevinlacaille if the check_item_types calls are expected to succeed, then we shouldn't get an exception, right?

tests/unit/test_specs.py
```diff
@@ -90,6 +107,12 @@ def test_get_product_bundles():
     assert TEST_PRODUCT_BUNDLE in bundles
 
 
-def test_get_item_types():
+def test_get_item_types_with_bundle():
     item_types = specs.get_item_types(TEST_PRODUCT_BUNDLE)
```
sgillies, line 111: Let's use the keyword argument form

```suggestion
    item_types = specs.get_item_types(product_bundle=TEST_PRODUCT_BUNDLE)
```

tests/unit/test_data_item_type.py
kevinlacaille: Indeed, you're right. It was actually failing on a BadParameter, as the item types are to be passed as a list. Fixed with the most recent push and changed to an assert.

sgillies (approved):
Good work @kevinlacaille !

PR #652, opened 2022-07-28
Warn on command execution using custom styled format

### LINKED ISSUE #651: Improve notice in the CLI about upcoming changes
In version 1.5.0, a warning is raised on import of planet.api, which is good, but it goes straight to stderr when we use the planet command, which is less good. For example:

```
$ planet
xxx/planet/api/__init__.py:38: ClientV1DeprecationWarning: The planet.api module is deprecated and will be removed in version 2.0.0. For more details please see the discussion at https://github.com/planetlabs/planet-client-python/discussions.
  warnings.warn(
Usage: planet [OPTIONS] COMMAND [ARGS]...

  Planet API Client
...
```

I propose that we intercept the warning using `warnings.filterwarnings` and instead add some boilerplate to the docstring of every CLI command so that a user sees something like this:

```
$ planet
Usage: planet [OPTIONS] COMMAND [ARGS]...

  Planet API Client

  Attention!

  You are using version 1 of the Planet CLI. In version 2, incompatible
  changes will be made to the Planet CLI, mainly to the arguments, options,
  and return values of subcommands. A pre-release of version 2 is available on
  the Python Package Index now. For more information about the new version,
  please see the discussion at https://github.com/planetlabs/planet-client-
  python/discussions.

Options:
  ...
```

sgillies:
Resolves #651.

Here's what the short warning looks like:

Screenshot from 2022-07-28 10-41-27

And here's what we see in `planet --help`:

```
$ planet --help
Usage: planet [OPTIONS] COMMAND [ARGS]...

  Planet API Client

  Attention!

  You are using version 1 of the Planet CLI. A pre-release of version 2 is
  available at https://pypi.org/project/planet/. Version 2 will not be
  backward compatible. Changes are planned for the arguments, options, and
  return values of subcommands. For more information about the new version,
  please join the discussion at https://github.com/planetlabs/planet-client-
  python/discussions.

Options:
```

### DIFF
--- a/planet/api/__init__.py
+++ b/planet/api/__init__.py
@@ -40,4 +40,5 @@ warnings.warn(
     "For more details please see the discussion at "
     "https://github.com/planetlabs/planet-client-python/discussions.",
     ClientV1DeprecationWarning,
+    stacklevel=2
 )
--- a/planet/scripts/cli.py
+++ b/planet/scripts/cli.py
@@ -15,6 +15,14 @@
 import click
 import logging
 import sys
+import textwrap
+import warnings
+
+# Importing planet.api triggers a deprecation warning. In the planet
+# command and its sub-commands we will intercept and ignore the
+# warning and instead show users a detailed warning in the command
+# usage text, and a short warning on command execution.
+warnings.filterwarnings("ignore", module="planet.api")
 
 from planet import api
 from planet.api.__version__ import __version__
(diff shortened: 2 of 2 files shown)


planet/scripts/cli.py
```diff
@@ -64,7 +72,35 @@ def configure_logging(verbosity):
...
+                "`PYTHONWARNINGS=ignore` in your environment.",
+                width=80,
+            )
+        ),
+        fg="yellow",
+    )
+    prev_formatwarning = warnings.formatwarning
+    warnings.formatwarning = lambda msg, *args, **kwargs: f"{msg}\n"
+    warnings.warn(warnmsg, UserWarning)
+    warnings.formatwarning = prev_formatwarning
```
sgillies, line 103: Previously, the warning was sent to the terminal using `click.echo`, but this doesn't provide an easy way to disable the warning. Using Python's warning system, as above, lets us use `PYTHONWARNINGS=ignore`.

planet/api/__init__.py
```diff
@@ -40,4 +40,5 @@ class ClientV1DeprecationWarning(FutureWarning):
     "For more details please see the discussion at "
     "https://github.com/planetlabs/planet-client-python/discussions.",
     ClientV1DeprecationWarning,
+    stacklevel=2
```
sgillies, line 43: Now, the warning will point to the module that imports planet.api instead of to planet.api itself.

PR #653, opened 2022-07-28
Add a draft GHA workflow to release to PyPI

sgillies:
This is a draft of a workflow that builds sdist and wheel distributions, checks them, and uploads to PyPI using twine. It's based upon workflows in [attrs](https://github.com/python-attrs/attrs/blob/main/.github/workflows/main.yml) and [shapely](https://github.com/shapely/shapely/blob/main/.github/workflows/release.yml). A supporting secret upload token has been added to the GitHub settings. I would like to test it by running it using a one-off pre-release tag. If it succeeds, I would yank the resulting pre-release.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,7 +1,5 @@
 on:
   push:
-    branches: ["main"]
-    tags: ["*"]
 
 jobs:
   package:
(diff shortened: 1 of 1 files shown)


sgillies:
I shouldn't overlook using the testing-pypi option: https://github.com/planetlabs/planet-client-python/pull/653/files#diff-87db21a973eed4fef5f32b267aa60fcee5cbdf03c67fafdc2a9b553bb0b15f34R3. I'll try that first.

sgillies:
Done! This action is run manually after selecting an existing tag. It makes a GitHub release (optionally pre-release), then uploads to TestPyPI, then uploads to PyPI. Note that if the TestPyPI upload fails, the PyPI upload will not start. This is a safeguard of sorts.

PR #654, opened 2022-07-29
filter cli command outputs a valid empty filter if all subfilters are removed. 

### LINKED ISSUE #600: Don't require a filter for a quick search.
To solve the issues below, do the following:
1. In quick search, make filter a named option (`--filter` for CLI, `search_filter` for api), if not supplied then no filtering will take place
2. Change `planet data filter` to output an empty filter (no filtering) by default so that `planet data filter | planet search PSScene --filter -` has the same behavior as `planet search PSScene` - this resolves #704 

---

If I try to do a quick search with no filter I get:

```
% planet data search-quick PSScene                                                              

Usage: planet data search-quick [OPTIONS] ITEM_TYPES FILTER

Try 'planet data search-quick --help' for help.

Error: Missing argument 'FILTER'.
```

In v1 doing a search without a filter would just show you the latest data. 

Right now you can do `planet data filter | planet data search-quick SkysatCollect -`, since filter outputs `{"type": "PermissionFilter", "config": ["assets:download"]}` by default.

But then if you didn't want to only search downloadable assets you could do:

`planet data filter --permission false | planet data search-quick SkysatCollect -` and then you get a failure, since the latter doesn't actually output a filter, it emits nothing.

jreiberkyle:
filter cli command outputs a valid empty filter if all subfilters are removed. document how to create this filter with the api.

closes #600

### DIFF
--- a/planet/cli/data.py
+++ b/planet/cli/data.py
@@ -218,12 +218,8 @@ def filter(ctx,
             else:
                 filters.append(f)
 
-    if filters:
-        if len(filters) > 1:
-            filt = data_filter.and_filter(filters)
-        else:
-            filt = filters[0]
-        echo_json(filt, pretty)
+    filt = data_filter.and_filter(filters)
+    echo_json(filt, pretty)
 
 
 @data.command()
--- a/planet/clients/data.py
+++ b/planet/clients/data.py
@@ -109,6 +109,21 @@ class DataClient:
         `name` parameter of the search defaults to the id of the generated
         search id if `name` is not specified.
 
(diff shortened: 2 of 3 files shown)


jreiberkyle:
@cholmes if this is a reasonable fix for 600, go ahead and approve and merge this. If not, comment with your thoughts and I'll address them when I get back.

cholmes (approved):

cholmes:
Hey, apologies for not getting to this earlier. I approved the request, as I think it's a fine improvement, but don't see it as fully fixiing #600

PR #668, opened 2022-08-04
create noxfile session and github action for publishing to testpypi

jreiberkyle:
Use nox for all of our build/publish-to-testpypi logic, which allows for testing and performing these steps outside of Github. Use GH Action to trigger the nox build/publish logic when a new tag is created.


jreiberkyle:
@sarasafavi and @sgillies FYI

Issue #597, opened 2022-06-21
Add STAC output as option for `planet orders request`


cholmes:
Right now there's no way in the CLI to request stac output, unless you hand craft your entire request. We should add a `--metadata stac` option to be able to include it in `planet orders request`.

cholmes:
I'm leaning towards just making stac the default, with a `--stac false` for people who explicitly don't want STAC. But afaik it's very little overhead, and it's a small file. But enables a useful set of tools.

PR #676, opened 2022-08-11
Add STAC output orders request

kevinlacaille:
As the title reads, I've added STAC output to the Orders CLIs and APIs `request` function.

`planet.cli.orders.py`
- Added `--stac` as a `click.option` as a flag (i.e., boolean)
    - @cholmes, I've set the default to be **_false_**, but I could be convinced that it should be set to true

`planet.orders_request.py`
- Added `stac` as an input to `build_request()`

`tests.integration.test_orders_cli.py`
- Added an integration test for the Orders CLI for a STAC output

`tests.unit.test_orders_request.py`
- Added a unit test for the Orders API (`build_request()`) for a STAC output

Below is an example of the STAC metadata at work:
```
> planet orders request \                          
--name Bangladesh_2018_visual \
--id 20180418_043113_1053,20180418_043112_1053 \
--item-type PSScene \
--bundle visual \
--stac

{"name": "Bangladesh_2018_visual", "products": [{"item_ids": ["20180418_043113_1053", "20180418_043112_1053"], "item_type": "PSScene", "product_bundle": "visual"}], "metadata": {"stac": {}}}
```

**FYI:** @cholmes @sgillies @jreiberkyle @mkshah605 

Closing #597

### DIFF
--- a/planet/cli/orders.py
+++ b/planet/cli/orders.py
@@ -263,6 +263,12 @@ async def create(ctx, request: str, pretty):
     type=types.JSON(),
     help="""Credentials for cloud storage provider to enable cloud delivery of
     data. Can be a json string, filename, or '-' for stdin.""")
+@click.option(
+    '--stac',
+    default=False,
+    is_flag=True,
+    help='Request metadata to be in SpatioTemporal Asset Catalog (STAC) format.'
+)
 @pretty
 async def request(ctx,
                   name,
@@ -273,6 +279,7 @@ async def request(ctx,
                   item_type,
                   email,
                   cloudconfig,
+                  stac,
                   pretty):
     """Generate an order request.
 
@@ -309,6 +316,7 @@ async def request(ctx,
                                                  products=[product],
                                                  delivery=delivery,
                                                  notifications=notifications,
-                                                 tools=tools)
+                                                 tools=tools,
+                                                 stac=stac)
 
     echo_json(request, pretty)
--- a/planet/cli/orders.py
+++ b/planet/cli/orders.py
@@ -264,8 +264,8 @@ async def create(ctx, request: str, pretty):
     help="""Credentials for cloud storage provider to enable cloud delivery of
     data. Can be a json string, filename, or '-' for stdin.""")
 @click.option(
-    '--no-stac',
-    default=False,
+    '--stac/--no-stac',
+    default=True,
     is_flag=True,
     help="""Do not request metadata to be in SpatioTemporal Asset Catalog
     (STAC) format.""")
@@ -279,7 +279,7 @@ async def request(ctx,
                   item_type,
                   email,
                   cloudconfig,
-                  no_stac,
+                  stac,
                   pretty):
     """Generate an order request.
 
@@ -312,10 +312,10 @@ async def request(ctx,
     else:
         delivery = None
 
-    if no_stac:
-        stac_json = {}
-    else:
+    if stac:
         stac_json = {'stac': {}}
+    else:
+        stac_json = {}
 
     request = planet.order_request.build_request(name,
                                                  products=[product],
--- a/planet/cli/orders.py
+++ b/planet/cli/orders.py
@@ -264,11 +264,11 @@ async def create(ctx, request: str, pretty):
     help="""Credentials for cloud storage provider to enable cloud delivery of
     data. Can be a json string, filename, or '-' for stdin.""")
 @click.option(
-    '--stac',
+    '--no-stac',
     default=False,
     is_flag=True,
-    help='Request metadata to be in SpatioTemporal Asset Catalog (STAC) format.'
-)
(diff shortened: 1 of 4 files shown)


kevinlacaille:
What's left?
- [ ] test STAC option in Data CLI
- [ ] test STAC option in `order_request.build_request()`

planet/cli/orders.py
```diff
@@ -263,6 +263,12 @@ async def create(ctx, request: str, pretty):
     type=types.JSON(),
     help="""Credentials for cloud storage provider to enable cloud delivery of
     data. Can be a json string, filename, or '-' for stdin.""")
+@click.option(
+    '--stac',
+    default=False,
```
tschaub, line 268: I think a default of `True` here makes sense (with a `--no-stac` option to disable).  Having a standard metadata format should provide more forward compatibility – someone could upgrade to our future Orders v3 API and ideally still use the same STAC metadata.

planet/order_request.py
```diff
@@ -86,6 +87,9 @@ def build_request(name: str,
     if tools:
         details['tools'] = tools
 
+    if stac:
+        details['metadata'] = {'stac': {}}
```
tschaub, line 91: This is an aside, but I notice that the metadata object is not well documented here: https://developers.planet.com/docs/orders/reference/#tag/Orders/operation/createOrder.  I'll create an upstream ticket looking into why that part of the spec doesn't get documented.

planet/cli/orders.py
cholmes: I'm +1 on making it default, especially since it's just an extra file that won't break anything. The one argument 'against' doing so would be that the STAC output from orders is still 'beta'. But so is the Python SDK, so I say we push to help get the stac output from orders to out of beta by the time we release the Python SDK as 2.0.0.

kevinlacaille: Sounds good, changed to STAC metadata as default, and changed flag accordingly (`--stac` ➡️  `--no-stac`)

planet/cli/orders.py
```diff
@@ -263,6 +263,12 @@ async def create(ctx, request: str, pretty):
     type=types.JSON(),
     help="""Credentials for cloud storage provider to enable cloud delivery of
     data. Can be a json string, filename, or '-' for stdin.""")
+@click.option(
+    '--no-stac',
```
tschaub, line 267: I don't use click a lot, so others might have a more informed comment on this (cc @sgillies), but it could be nice to have `--stac/--no-stac` be the pair of options here (with the default being `False` (e.g. https://click-docs-cn.readthedocs.io/zh_CN/latest/options.html#boolean-flags).  That would let everybody be explicit if they wanted to, and both `--stac` and `--no-stac` options would continue to work if we ever change the default.

tests/integration/test_orders_cli.py
```diff
@@ -649,6 +666,33 @@ def test_cli_orders_request_tools(invoke, geom_geojson):
...
+    order_request = {
+        "name":
+        "test",
+        "products": [{
+            "item_ids": ["4500474_2133707_2021-05-20_2419"],
+            "item_type": "PSOrthoTile",
+            "product_bundle": "analytic",
+        }],
+        "metadata":
+        stac_json
```
tschaub, line 696: I'm assuming the API treats `"metadata": {}` the same as omitting the `metadata` member altogether - but don't know this for sure.

tschaub (approved):
Looks good, @kevinlacaille.  You can take or leave my comment as you like.

planet/cli/orders.py
kevinlacaille: @tschaub @sgillies I'm not sure I see the purpose of `--stac/--no-stac`, if we're going to provide STAC by default - i.e., what's the purpose of having a `--stac` flag?

tests/integration/test_orders_cli.py
kevinlacaille: In these tests we're only testing for when `--no-stac` is set to false, i.e., we are requesting STAC metadata. I'm not entirely sure what would happen if we requested `"metadata": {}`. I suppose this should be something we test for in `tests.integration.test_order_request.test_build_request()`. Nice catch.

kevinlacaille: Indeed, by adding a `--no-stac` flag, the output from the Orders API does not have `"metadata": {}`. See: https://github.com/planetlabs/planet-client-python/pull/676/commits/84d2e5423afa563005678e770d9f5aba2a7d2adb

planet/cli/orders.py
kevinlacaille: After our discussion, I the `--stac/--no-stac` boolean pair flags are helpful for being explicit. See: https://github.com/planetlabs/planet-client-python/pull/676/commits/affba48d659055fe102fe088278246479ac1b6c6

tschaub (approved):

PR #677, opened 2022-08-18
Compatible bundle with item type error

### LINKED ISSUE #618: planet order request: be more helpful on what bundles are available for a given item_type
Right now `planet order request` fails if item_type and bundle are incompatible by saying what item-types are compatible with the specified bundle. It would be more helpful to switch this so that it indicates what bundles are compatible with an item-type.

This has been hitting me as I've changed over from `PSScene4Band` to `PSScene` item-type, which is not compatible with the `analytic` bundle.

kevinlacaille:
Validating `item_type` before `bundle`, so that the item type compatibility exception will propagate before the bundle validation exception.

Closes #618

### DIFF
--- a/planet/specs.py
+++ b/planet/specs.py
@@ -37,9 +37,22 @@ SUPPORTED_FILE_FORMATS = ['COG', 'PL_NITF']
 LOGGER = logging.getLogger(__name__)
 
 
-class SpecificationException(Exception):
+class NoMatchException(Exception):
     '''No match was found'''
     pass
+
+
+class SpecificationException(Exception):
+    '''No match was found'''
+
+    def __init__(self, value, supported, field_name):
+        self.value = value
+        self.supported = supported
+        self.field_name = field_name
+        self.opts = ', '.join(["'" + s + "'" for s in supported])
+
+    def __str__(self):
+        return f'{self.field_name} - \'{self.value}\' is not one of {self.opts}.'
 
 
 def validate_bundle(bundle):
@@ -48,7 +61,8 @@ def validate_bundle(bundle):
 
 
 def validate_item_type(item_type, bundle):
-    supported = get_item_types(bundle)
+    validated_bundle = validate_bundle(bundle)
+    supported = get_item_types(validated_bundle)
     return _validate_field(item_type, supported, 'item_type')
 
 
@@ -72,20 +86,13 @@ def validate_file_format(file_format):
 
 def _validate_field(value, supported, field_name):
     try:
-        value = get_match(value, supported)
+        value = get_match(value, supported, field_name)
     except (NoMatchException):
-        opts = ', '.join(["'" + s + "'" for s in supported])
-        msg = f'{field_name} - \'{value}\' is not one of {opts}.'
-        raise SpecificationException(msg)
+        raise SpecificationException(value, supported, field_name)
     return value
 
 
-class NoMatchException(Exception):
-    '''No match was found'''
-    pass
-
-
-def get_match(test_entry, spec_entries):
+def get_match(test_entry, spec_entries, field_name):
     '''Find and return matching spec entry regardless of capitalization.
 
     This is helpful for working with the API spec, where the capitalization
@@ -95,7 +102,7 @@ def get_match(test_entry, spec_entries):
         match = next(e for e in spec_entries
                      if e.lower() == test_entry.lower())
     except (StopIteration):
-        raise NoMatchException('{test_entry} should be one of {spec_entries}')
+        raise SpecificationException(test_entry, spec_entries, field_name)
 
     return match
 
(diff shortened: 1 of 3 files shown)


mkshah605 (approved):
makes sense and looks good!

planet/specs.py
```diff
@@ -48,8 +48,8 @@ def validate_bundle(bundle):
 
 
 def validate_item_type(item_type, bundle):
-    bundle = validate_bundle(bundle)
     supported = get_item_types(bundle)
+    bundle = validate_bundle(bundle)
```
sgillies, line 52: @kevinlacaille am I missing something or does this repeat the validation on line 117 of order_request.py?

planet/specs.py
```diff
@@ -111,15 +111,18 @@ def get_item_types(product_bundle=None):
...
-    else:
-        product_bundle = get_product_bundles()
-        all_item_types = []
-        for bundle in product_bundle:
-            all_item_types += [*spec['bundles'][bundle]['assets'].keys()]
-        item_types = set(all_item_types)
-    return item_types
+    try:
+        if product_bundle:
+            item_types = spec['bundles'][product_bundle]['assets'].keys()
```
sgillies, line 116: Small thing, but let's not put any empty lines after `try`, `if`, `for`, `while`, etc. I think an empty line *before* those can help readability, but not after.

sgillies: Also it looks like `item_types` is never returned in this case?

sgillies: I would suggest this for line 116: `return set(spec['bundles'][product_bundle]['assets'].keys())`

planet/specs.py
```diff
@@ -111,15 +111,18 @@ def get_item_types(product_bundle=None):
...
+        else:
+            product_bundle = get_product_bundles()
+            all_item_types = []
+            for bundle in product_bundle:
+                all_item_types += [*spec['bundles'][bundle]['assets'].keys()]
+            item_types = set(all_item_types)
+        return item_types
```
sgillies, line 123: An alternative Pythonic approach:

```suggestion
        return set(itertools.chain.from_iterable(spec['bundles'][bundle]['assets'].keys() for bundle in get_product_bundles()))
```

planet/specs.py
```diff
@@ -111,15 +111,18 @@ def get_item_types(product_bundle=None):
...
+    except KeyError:
+        raise SpecificationException()
```
sgillies, line 125: Add a specific message?

sgillies (changes requested):
@kevinlacaille I may have found a issue that needs correction. I also made a couple suggestions to use Python language features for improved performance.

planet/specs.py
kevinlacaille: > Small thing, but let's not put any empty lines after `try`, `if`, `for`, `while`, etc. I think an empty line _before_ those can help readability, but not after.

You got it.



> Also it looks like `item_types` is never returned in this case?

`item_types` is returned on line 123, after the `if`/`else` statement.



> I would suggest this for line 116: `return set(spec['bundles'][product_bundle]['assets'].keys())`

On it!

kevinlacaille: For readability, I'll move line 123 (`return item_types`) to after `except KeyError`.

planet/specs.py
kevinlacaille: I think the `SpecificationException` raised by `_validate_field()`, which `validate_bundle()` calls is perfect for this case. I'm not quite sure if this is fine, but could I just raise `validate_bundle(product_bundle)`, as that'll raise a `SpecificationException` with the proper message?

kevinlacaille: See https://github.com/planetlabs/planet-client-python/pull/677/commits/614ebe5d0d7edd7d34b052ec81e266d757cbdacc#diff-0c5c73e85ab9430a67e6b85ba97b8d4fd186f5240bc8f4c1ddb7ef52fade26feR123

planet/specs.py
kevinlacaille: I wanted to make sure `validate_item_type()` had a validated bundle, but I suppose putting it after `get_item_type()` was useless haha. That being said, now when we call `get_item_type()` that'll raise a `KeyError` and raise `validate_bundle()`, which will raise the proper exception.

planet/specs.py
```diff
@@ -111,14 +110,18 @@ def get_item_types(product_bundle=None):
...
+    try:
+        if product_bundle:
+            item_types = set(spec['bundles'][product_bundle]['assets'].keys())
+        else:
+            item_types = set(
+                itertools.chain.from_iterable(
+                    spec['bundles'][bundle]['assets'].keys()
+                    for bundle in get_product_bundles()))
+    except KeyError:
+        raise validate_bundle(product_bundle)
```
sgillies, line 123: @kevinlacaille I think validating the product bundle should be done before line 116. Then we would probably not see a KeyError ever except if the bundle object had no 'assets' item.

planet/specs.py
sgillies: @kevinlacaille let's not make a one-line function for `spec['bundles'][product_bundle]['assets']`. I believe a Bundle class would be more effective. Imagine usage like `return set(itertools.chain.from_iterable(bundle.assets for bundle in get_product_bundles())`.

The advantage of `itertools.chain.from_iterable`, IMO, is that it tells me clearly (as reader) what the intent is. Whereas I have to deduce the intent of the more verbose code.

planet/specs.py
sgillies: @kevinlacaille it occurs to me that with some small changes to SpecificationException, it could be raised from anywhere and you wouldn't need to call any one particular function to trigger it. For example.

```
class SpecificationException(Exception):
    def __init__(value, supported, field_name):
        self.value = value
        self.supported = supported
        self.field_name = field_name
    def __str__(self):
        return f'{self.field_name} - \'{self.value}\' is not one of {self.opts}.'
```

kevinlacaille: Oh, I really prefer this - great suggestion!

planet/specs.py
kevinlacaille: You're right, that makes more sense. Done!

sgillies (approved):
Good work @kevinlacaille !

PR #616, opened 2022-06-29
propose new request-update command

jreiberkyle:

PR #680, opened 2022-08-24
Get supported bundles for a given item_type

kevinlacaille:
PR #677 did not fully address this issue, however this PR should cover all the bases!

Validating `item_type` before `product_bundle`, so that the item type compatibility exception will propagate before the bundle validation exception.
1. `sepcs.validate_item_type()` now just checks if input `item_type` is a valid input (i.e., if part of all item type)
2. `specs. validate_bundle()`  now checks if the input product bundle is a valid product bundle (i.e., if part of all product bundles), then checks if item type is compatible with the given product bundle
    - Important part! If the item type and product bundle are not compatible, it returns all valid product bundles for given item type with the new function `specs.validate_supported_bundles()`
3. Created tests for new function, `specs.validate_supported_bundles()`
4. Reformatted tests for `specs.validate_bundle()` and `specs.validate_item_type()`
5. As per @jreiberkyle's new PR, #616, `item-type` and `bundle` are now both arguments, instead of options
    - Changed all associated tests
6. @cholmes We could not accomplish adding the feature that presents the user with a subset of valid bundles for a given item type when only providing the item type, as to do this we'd have to add the `is_eagar` flag to `item_type`, which 1. requires `item_type` to be an option and 2. requires an item type to be given when entering `planet orders request --help`.
    - So, now the user is presented with all compatible bundles only when the API is presented with an incompatible bundle. See **New behaviour (1)**.
    
**Previous behaviour (1):** Incompatible item type with given bundle would return all compatible item types with given bundle.
```
❯ planet orders request --item-type PSScene --bundle analytic --id 20220919_183509_93_248f --name hood-stac 

Usage: planet orders request [OPTIONS]
Try 'planet orders request --help' for help.

Error: Invalid value: item_type - 'PSScene' is not one of 'REOrthoTile', 'PSOrthoTile', 'PSScene4Band', 'Sentinel1', 'SkySatScene', 'Landsat8L1G', 'Sentinel2L1C', 'PSScene3Band', 'SkySatCollect'.
```

**New behaviour (1):** Incompatible item type with given bundle now returns all compatible bundles with given item type.
```
❯ planet orders request PSScene analytic --id 20220919_183509_93_248f --name hood-stac

Usage: planet orders request [OPTIONS]
Try 'planet orders request --help' for help.

Error: Invalid value: bundle - 'analytic' is not one of 'analytic_udm2', 'analytic_3b_udm2', 'analytic_8b_udm2', 'visual', 'basic_analytic_udm2', 'basic_analytic_8b_udm2', 'analytic_sr_udm2', 'analytic_8b_sr_udm2'.
```

**Previous behaviour (2):** Only `--name` provided prompted user to enter a bundle.
```
> planet orders request --name "dude"                   
Usage: planet orders request [OPTIONS]
Try 'planet orders request --help' for help.

Error: Missing option '--bundle'. Choose from:
	analytic,
	analytic_udm2,
	analytic_3b_udm2,
	analytic_5b,
	analytic_5b_udm2,
	analytic_8b_udm2,
	visual,
	uncalibrated_dn,
	uncalibrated_dn_udm2,
	basic_analytic,
	basic_analytic_udm2,
	basic_analytic_8b_udm2,
	basic_uncalibrated_dn,
	basic_uncalibrated_dn_udm2,
	analytic_sr,
	analytic_sr_udm2,
	analytic_8b_sr_udm2,
	basic_uncalibrated_dn_nitf,
	basic_uncalibrated_dn_nitf_udm2,
	basic_analytic_nitf,
	basic_analytic_nitf_udm2,
	basic_panchromatic,
	basic_panchromatic_dn,
	panchromatic,
	panchromatic_dn,
	panchromatic_dn_udm2,
	pansharpened,
	pansharpened_udm2,
	basic_l1a_dn
```
**New behaviour (2):** Only `--name` provided now prompts the user to enter an item type **and** now provides all choices for item types.
```
❯ planet orders request --name "dude"        
Usage: planet orders request [OPTIONS] ITEM_TYPE BUNDLE
Try 'planet orders request --help' for help.

Error: Missing argument 'ITEM_TYPE'. Choose from:
        PSScene3Band,
        SkySatScene,
        PSScene,
        MYD09GA,
        PSOrthoTile,
        MYD09GQ,
        MOD09GA,
        Sentinel1,
        REOrthoTile,
        REScene,
        SkySatCollect,
        Sentinel2L1C,
        MOD09GQ,
        Landsat8L1G,
        PSScene4Band
```
Closing #618, #698


tests/unit/test_orders_callbacks.py
```diff
@@ -0,0 +1,54 @@
...
+class Param(object):
+
+    def __init__(self):
+        self.name = 'bundle'
+
+
+def test_stash_item_type_success():
+    ctx = MockContext()
+    stash_item_type(ctx, 'item_type', TEST_ITEM_TYPE)
+    assert ctx.obj == {'item_type': TEST_ITEM_TYPE}
```
sgillies, line 39: @kevinlacaille here's a more limited assertion that doesn't presume anything else about `ctx.obj`.

```suggestion
    assert ctx.obj['item_type'] = TEST_ITEM_TYPE
```

tests/unit/test_orders_callbacks.py
```diff
@@ -0,0 +1,54 @@
...
+
+
+def test_bundle_cb_success():
+    ctx = MockContext()
+    ctx.obj['item_type'] = TEST_ITEM_TYPE
+    result = bundle_cb(ctx, 'bundle', TEST_ITEM_TYPE)
+    assert result == TEST_ITEM_TYPE
+
+
+def test_bundle_cb_fail():
```
sgillies, line 49:
```suggestion
def test_bundle_cb_missing_parameter():
    """If bundle option is missing, print helpful error using item type information."""
```

tests/unit/test_orders_callbacks.py
```diff
@@ -0,0 +1,54 @@
...
+    ctx = MockContext()
+    ctx.obj['item_type'] = TEST_ITEM_TYPE
+    param = Param()
+    with pytest.raises(click.ClickException):
+        bundle_cb(ctx, param, None)
```
sgillies, line 54: @kevinlacaille I feel like this test should assert on the error msg text, demonstrating that we're using the item type information. That's one of the points of this PR.

tests/unit/test_orders_callbacks.py
```diff
@@ -0,0 +1,54 @@
...
+        self.name = 'bundle'
+
+
+def test_stash_item_type_success():
+    ctx = MockContext()
+    stash_item_type(ctx, 'item_type', TEST_ITEM_TYPE)
+    assert ctx.obj == {'item_type': TEST_ITEM_TYPE}
+
+
+def test_bundle_cb_success():
```
sgillies, line 42:
```suggestion
def test_bundle_cb_pass_through():
    """Do nothing if the bundle value is defined."""
```

tests/unit/test_orders_callbacks.py
```diff
@@ -0,0 +1,54 @@
...
+    ctx = MockContext()
+    ctx.obj['item_type'] = TEST_ITEM_TYPE
+    result = bundle_cb(ctx, 'bundle', TEST_ITEM_TYPE)
+    assert result == TEST_ITEM_TYPE
```
sgillies, line 46: Passing an item type in could be a future red herring. Let's not suggest that we use it like that.

```suggestion
    result = bundle_cb(ctx, 'bundle', 'anything')
    assert result == 'anything'
```

tests/unit/test_orders_callbacks.py
kevinlacaille: I agree with you. I initially had it that way, but for some reason changed it. I'll change it back.

tests/unit/test_orders_callbacks.py
sgillies: Or I think it would suffice to simply assert that `bundle_cb(ctx, 'bundle', 'anything')` doesn't raise an exception.

planet/cli/orders.py
```diff
@@ -221,6 +221,35 @@ async def create(ctx, request: str, pretty):
...
+    name = "bundle"
+
+    def __init__(self, case_sensitive: bool = True) -> None:
+        self.choices = []
+        self.case_sensitive = case_sensitive
+
+    def convert(self, value, param, ctx):
+        item_type = ctx.obj['item_type']
+        self.choices = planet.specs.get_product_bundles(item_type)
+        return super().convert(value, param, ctx)
```
sgillies, line 250: @kevinlacaille this class and the functions above it need some docstrings and comments to explain their purpose. Like I said yesterday, this is a bit hacky, and I'm concerned that we might eventually forget why we took this particular approach or how it works.

planet/cli/orders.py
```diff
@@ -221,6 +221,35 @@ async def create(ctx, request: str, pretty):
...
+    return value
+
+
+def bundle_cb(ctx, param, value):
+    item_type = ctx.obj['item_type']
+    choices = planet.specs.get_product_bundles(item_type)
+    if value is None:
+        msg = ("Choose from:\n\t{choices}").format(
+            choices=",\n\t".join(choices))
+        raise click.ClickException(f"Missing option '--{param.name}'. {msg}")
```
sgillies, line 235: @kevinlacaille will you try moving this branch into Bundle.convert? Maybe we could do away with the callback and simplify this.

planet/cli/orders.py
```diff
@@ -221,6 +221,35 @@ async def create(ctx, request: str, pretty):
...
+class Bundle(click.Choice):
+    name = "bundle"
+
+    def __init__(self, case_sensitive: bool = True) -> None:
+        self.choices = []
+        self.case_sensitive = case_sensitive
+
+    def convert(self, value, param, ctx):
+        item_type = ctx.obj['item_type']
+        self.choices = planet.specs.get_product_bundles(item_type)
```
sgillies, line 249: Something like this

```suggestion
        self.choices = planet.specs.get_product_bundles(item_type)
        if value is None:
            msg = ("Choose from:\n\t{choices}").format(
                choices=",\n\t".join(self.choices))
            raise click.ClickException(f"Missing option '--{param.name}'. {msg}")
  
```

jreiberkyle:
When I run `planet orders request --help` I get the following:

```
Usage: planet orders [OPTIONS] COMMAND [ARGS]...
Try 'planet orders --help' for help.

Error: No such command 'requests'.
planet-client-python$> planet orders request --help
Usage: planet orders request [OPTIONS]
Try 'planet orders request --help' for help.

Error: Missing option '--item-type'. Choose from:
	SkySatCollect,
	PSScene,
	MYD09GQ,
	SkySatScene,
	Landsat8L1G,
	MOD09GA,
	MYD09GA,
	Sentinel1,
	PSScene3Band,
	PSScene4Band,
	REOrthoTile,
	REScene,
	MOD09GQ,
	PSOrthoTile,
	Sentinel2L1C
```

This is due to using the `is_eager` flag in the `item-type` option. 

I imagine the `is_eager` flag is set to True to allow filtering bundles by item-type.

What if, instead, we to keep the CLI option logic pretty simple here. That is, the `bundle` option lists all possible bundles, the `item-types` option lists all possible item types, and the determination of whether the bundle and item-type are compatible, and creation of the error message and raising of `BadParameter` with the subset of bundles valid for the specified item-type is done within the command function.

kevinlacaille:
> When I run `planet orders request --help` I get the following:
> 
> ```
> Usage: planet orders [OPTIONS] COMMAND [ARGS]...
> Try 'planet orders --help' for help.
> 
> Error: No such command 'requests'.
> planet-client-python$> planet orders request --help
> Usage: planet orders request [OPTIONS]
> Try 'planet orders request --help' for help.
> 
> Error: Missing option '--item-type'. Choose from:
> 	SkySatCollect,
> 	PSScene,
> 	MYD09GQ,
> 	SkySatScene,
> 	Landsat8L1G,
> 	MOD09GA,
> 	MYD09GA,
> 	Sentinel1,
> 	PSScene3Band,
> 	PSScene4Band,
> 	REOrthoTile,
> 	REScene,
> 	MOD09GQ,
> 	PSOrthoTile,
> 	Sentinel2L1C
> ```
> 
> This is due to using the `is_eager` flag in the `item-type` option.
> 
> I imagine the `is_eager` flag is set to True to allow filtering bundles by item-type.
> 
> What if, instead, we to keep the CLI option logic pretty simple here. That is, the `bundle` option lists all possible bundles, the `item-types` option lists all possible item types, and the determination of whether the bundle and item-type are compatible, and creation of the error message and raising of `BadParameter` with the subset of bundles valid for the specified item-type is done within the command function.

That's a great catch, thanks, @jreiberkyle!

I'd love to find a way around the `is_eager` flag, but I can't see a solution. 

I'm not against presenting all bundle options to the users, then raising a `BadParameter` with the subset of valid bundles, if they enter an incompatible bundle, like you mentioned. This would vastly reduce the function's dependencies, but would come at the cost of not giving the user precisely the information they need, right away. 

Thoughts @sgillies @cholmes ?

jreiberkyle:
Another thing to bring into consideration is that with #616, item-type and bundle will become required arguments instead of options. Not sure how, if at all, this would affect this interaction between item-type and bundle.

jreiberkyle (approved):
This seems to work as I'd expect:
```
> planet orders request bop boop
Usage: planet orders request [OPTIONS] ITEM_TYPE BUNDLE
Try 'planet orders request --help' for help.

Error: Invalid value for 'ITEM_TYPE': 'bop' is not one of 'MYD09GQ', 'PSScene3Band', 'SkySatCollect', 'Landsat8L1G', 'PSScene4Band', 'Sentinel1', 'SkySatScene', 'PSScene', 'REOrthoTile', 'MYD09GA', 'MOD09GA', 'Sentinel2L1C', 'REScene', 'PSOrthoTile', 'MOD09GQ'.
```

```
> planet orders request psscene boop
Usage: planet orders request [OPTIONS] ITEM_TYPE BUNDLE
Try 'planet orders request --help' for help.

Error: Invalid value for 'BUNDLE': 'boop' is not one of 'analytic', 'analytic_udm2', 'analytic_3b_udm2', 'analytic_5b', 'analytic_5b_udm2', 'analytic_8b_udm2', 'visual', 'uncalibrated_dn', 'uncalibrated_dn_udm2', 'basic_analytic', 'basic_analytic_udm2', 'basic_analytic_8b_udm2', 'basic_uncalibrated_dn', 'basic_uncalibrated_dn_udm2', 'analytic_sr', 'analytic_sr_udm2', 'analytic_8b_sr_udm2', 'basic_uncalibrated_dn_nitf', 'basic_uncalibrated_dn_nitf_udm2', 'basic_analytic_nitf', 'basic_analytic_nitf_udm2', 'basic_panchromatic', 'basic_panchromatic_dn', 'panchromatic', 'panchromatic_dn', 'panchromatic_dn_udm2', 'pansharpened', 'pansharpened_udm2', 'basic_l1a_dn'.
```

The one thing I'd add is some helpful text in the docstring that indicates that the valid bundle is item-type specific, and I'd also consider adding that to the error message that comes up for 'BUNDLE', something like 

"Error: Invalid value for 'BUNDLE': 'boop' is not one of **the following valid bundles for PsScene** 'analytic', ..."

kevinlacaille:
> The one thing I'd add is some helpful text in the docstring that indicates that the valid bundle is item-type specific, and I'd also consider adding that to the error message that comes up for 'BUNDLE', something like
> 
> "Error: Invalid value for 'BUNDLE': 'boop' is not one of **the following valid bundles for PsScene** 'analytic', ..."

The error message bubbles up from an exception from `click.Choice()` when building the request (`planet.cli.orders.request()`). Having the item type known by the `@click.option` decorator is impossible without having a callback function in bundle's argument, which is a known issue (see point no.6 https://github.com/planetlabs/planet-client-python/pull/680#issue-1349658111).

I think for now the exception message will have to remain the same.

cholmes:
Tried this out, and overall it looks pretty good. I agree that it'd be ideal if we can just return the actual valid bundles after a user puts the item type in. But the experience is 'good enough' to merge this in - if I select a wrong one it tells me on the next step. As a user I'll wonder why they didn't just tell me, but I'll get to the right spot.

So let's merge this in. I would like to see if there's any way to do a better warning, but can just make a ticket for the future - it's no longer 'must fix this', more in 'that'd be nice'.

cholmes (approved):
Looks great, thanks for all your hard work on this Kevin.

cholmes:
Oh, also - I added in #712, since this brings a material change to how the CLI works, and we need to be sure the docs stay up to date with the code. But I think we can merge this, just need to be sure to get #712 in before the next release.

kevinlacaille:
Thanks for all of your reviews, @sgillies, @jreiberkyle, and @cholmes! 

I agree the UX isn't the best it could be, but we'll need to think hard about a workaround with or without `click` for the solution we are hoping for, in a future ticket.

Issue #678, opened 2022-08-18
Trying to download an order using order ID and failing due to a 302


olepinard:
I am trying to loop through order ID's and am erroring out with the error included below. 

`   async with Session() as sess:
        client = OrdersClient(sess)
        for order in order_ids:
            print(order)
            await client.download_order(order)
`
Is it possible that the 302 is not being handled correctly?

[302_error_output.txt](https://github.com/planetlabs/planet-client-python/files/9376717/302_error_output.txt)

cholmes:
Hey @olepinard - super sorry for the slow response, this slipped through the cracks with vacation, etc. Will try to investigate this soon - thanks for reporting it!

sgillies:
This will be fixed in 2.0a3.

cholmes:
Closing this as #685 merged and a3 was released. @olepinard - feel free to re-open if the new release doesn't handle your issue.

PR #685, opened 2022-09-01
Enable redirect following and avoid treating 302s as errors

sgillies:
Towards resolving #678.

Without these changes, an attempt to download an order with 0.23 produces

```
Error:
```

With these changes, I'm able to download an order.

### DIFF
--- a/planet/http.py
+++ b/planet/http.py
@@ -97,7 +97,6 @@ class BaseSession:
             try:
                 response.raise_for_status()
             except httpx.HTTPStatusError as e:
-                import pdb; pdb.set_trace()
                 e.response.read()
                 cls._convert_and_raise(e)
 
@@ -243,7 +242,9 @@ class Session(BaseSession):
 
         LOGGER.info(f'Session read timeout set to {READ_TIMEOUT}.')
         timeout = httpx.Timeout(10.0, read=READ_TIMEOUT)
-        self._client = httpx.AsyncClient(auth=auth, timeout=timeout, follow_redirects=True)
+        self._client = httpx.AsyncClient(auth=auth,
+                                         timeout=timeout,
+                                         follow_redirects=True)
 
         self._client.headers.update({'User-Agent': self._get_user_agent()})
         self._client.headers.update({'X-Planet-App': 'python-sdk'})
@@ -271,7 +272,6 @@ class Session(BaseSession):
             try:
                 response.raise_for_status()
             except httpx.HTTPStatusError as e:
-                import pdb; pdb.set_trace()
                 await e.response.aread()
                 cls._convert_and_raise(e)
 
--- a/planet/http.py
+++ b/planet/http.py
@@ -93,11 +93,13 @@ class BaseSession:
 
     @classmethod
     def _raise_for_status(cls, response):
-        try:
-            response.raise_for_status()
-        except httpx.HTTPStatusError as e:
-            e.response.read()
-            cls._convert_and_raise(e)
+        if response.status_code >= 400:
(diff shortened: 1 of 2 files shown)


planet/http.py
```diff
@@ -241,7 +243,7 @@ def __init__(self, auth: AuthType = None):
 
         LOGGER.info(f'Session read timeout set to {READ_TIMEOUT}.')
         timeout = httpx.Timeout(10.0, read=READ_TIMEOUT)
-        self._client = httpx.AsyncClient(auth=auth, timeout=timeout)
+        self._client = httpx.AsyncClient(auth=auth, timeout=timeout, follow_redirects=True)
```
sgillies, line 246: Needed since httpx 0.20.

planet/http.py
```diff
@@ -93,11 +93,13 @@ def _convert_and_raise(cls, error):
 
     @classmethod
     def _raise_for_status(cls, response):
-        try:
-            response.raise_for_status()
-        except httpx.HTTPStatusError as e:
-            e.response.read()
-            cls._convert_and_raise(e)
+        if response.status_code >= 400:
```
sgillies, line 96: Unless we do this we will trip on the fact that `httpx.Response(302).raise_for_status()` raises an exception in version 0.23, even when a redirect is followed. I wonder if httpx isn't updating the status code or request after following a redirect? Might be worth checking the project's bug tracker about this.

tests/integration/test_orders_api.py
```diff
@@ -33,6 +33,7 @@
 TEST_URL = 'http://www.MockNotRealURL.com/api/path'
 TEST_BULK_CANCEL_URL = f'{TEST_URL}/bulk/orders/v2/cancel'
 TEST_DOWNLOAD_URL = f'{TEST_URL}/download'
+TEST_DOWNLOAD_ACTUAL_URL = f'{TEST_URL}/download_actual'
```
sgillies, line 36: There are probably better names and URLs for this. In the actual API, it's `link.planet.com`, but we shouldn't depend on anything but the location header.

tests/integration/test_orders_api.py
```diff
@@ -540,8 +541,16 @@ async def test_download_asset_md(tmpdir, session):
         'Content-Type': 'application/json',
         'Content-Disposition': 'attachment; filename="metadata.json"'
     }
+
+    mock_redirect = httpx.Response(HTTPStatus.FOUND,
+                                   headers={
+                                       'Location': TEST_DOWNLOAD_ACTUAL_URL,
+                                       'Content-Length': '0'
```
sgillies, line 548: @jreiberkyle there was an exception in the model module if the redirect response didn't have a content-length header. This feels like something we should fix; our API redirect response has an empty body.

cholmes (approved):

planet/http.py
sgillies: Good news is that `httpx.Response(302).is_error` is `False`.

planet/http.py
```diff
@@ -265,11 +268,12 @@ async def alog_response(*args, **kwargs):
 
     @classmethod
     async def _raise_for_status(cls, response):
-        try:
-            response.raise_for_status()
-        except httpx.HTTPStatusError as e:
-            await e.response.aread()
-            cls._convert_and_raise(e)
+        if response.is_error:
```
sgillies, line 271: See https://github.com/planetlabs/planet-client-python/pull/685#discussion_r961097899. This is probably the same thing as `response.status_code >= 400` but syntactically a little neater, maybe?

PR #689, opened 2022-09-08
Subscription Helper Method spec

cholmes:
This is still very much work-in-progress, but wanted to put something up for discussion.

### DIFF
--- a/design-docs/CLI-Subscriptions.md
+++ b/design-docs/CLI-Subscriptions.md
@@ -175,6 +175,49 @@ In the case of invalid options, print an error message (stderr) and set the exit
 code to 2. In the case of an API error response, print the response JSON and 
 set code to 1.
 
+## Request
+
+### Interface
+
+```
+planet subscriptions request [OPTIONS]
+
+Generate a subscriptions request.
+
+This command provides support for building the subscription request JSON used to create or 
+update a subscription. It outputs the subscription request.
+
+Options:
+  --name TEXT                     Subscription name. Does not need to be unique.
+                                  [required]
+  --asset-types TEXT              One or more comma-separated asset types [required]
+  --item-types TEXT               One or more comma-separated item-types [required]
+  --geometry JSON                 GeoJSON of area to deliver subscription to. Use 
+                                  --clip only deliver pixels that match this area.
+                                  Can be a json string, filename, or '-' for stdin.
+  --clip                          Use the geometry GeoJSON as the clip geometry for 
+                                  the subscription.
+  --start-time DATETIME           Start date and time to begin subscription.
+  --end-time DATETIME             Date and time to end the subscription.
+  --rrule TEXT                    iCalendar recurrance rule to specify recurrances. 
+  --filter JSON                   A search filter can be specified a json string, 
+                                  filename, or '-' for stdin. 
+  --tools JSON                    Toolchain JSON. Can be a json string,
+                                  filename, or '-' for stdin.
+  --cloudconfig JSON              Credentials for cloud storage provider to
+                                  enable cloud delivery of data. Can be a json
+                                  string, filename, or '-' for stdin.
+  --notifications JSON            Notification JSON to specify webhook topics.
+  --pretty                        Format JSON output.
+  --help                          Show this message and exit.
+```
+
+### Usage Examples
+
+```
+planet subscription request --item-type PSScene --asset-types ortho_analytic_8b_sr,ortho_udm2 --geom JSON --clip --start-time 05/01/2022 --cloudconfig delivery.json | planet subscriptions create -
+```
+
 ## Update
 
 Edit a subscription, such as one with a future start date, before it starts running.


design-docs/CLI-Subscriptions.md
```diff
@@ -175,6 +175,49 @@ In the case of invalid options, print an error message (stderr) and set the exit
...
+
+Options:
+  --name TEXT                     Subscription name. Does not need to be unique.
+                                  [required]
+  --asset-types TEXT              One or more comma-separated asset types [required]
+  --item-types TEXT               One or more comma-separated item-types [required]
+  --geometry JSON                 GeoJSON of area to deliver subscription to. Use 
+                                  --clip only deliver pixels that match this area.
+                                  Can be a json string, filename, or '-' for stdin.
+  --clip                          Use the geometry GeoJSON as the clip geometry for
```
cholmes, line 198: This is an attempt to 'reuse' the geometry JSON. It might be too restrictive, since it could only use the exact same geometry.  If we wanted more flexibility could add an optional argument, but if no argument is given then it'd use the geometry. 

This likely gets complicated trying to combine it with 'tools' JSON. In orders we just don't allow clip and tools in the same operation. But I think we have some convenience that lets you use both. I think we should do #801, so perhaps there's some way to make it easy to input / reuse the same geometry. Or more simple could be that you create a tool.json that is 'ready' for a subscription and would just reuse the geometry JSON when it is there as the clip. Like it'd make a special 'clip' section that would get filled out later.

design-docs/CLI-Subscriptions.md
```diff
@@ -175,6 +175,49 @@ In the case of invalid options, print an error message (stderr) and set the exit
...
+
+This command provides support for building the subscription request JSON used to create or 
+update a subscription. It outputs the subscription request.
+
+Options:
+  --name TEXT                     Subscription name. Does not need to be unique.
+                                  [required]
+  --asset-types TEXT              One or more comma-separated asset types [required]
+  --item-types TEXT               One or more comma-separated item-types [required]
+  --geometry JSON                 GeoJSON of area to deliver subscription to. Use
```
cholmes, line 195: This is not required by the API, though I don't really get the use case of a subscription that would not have a geometry....

design-docs/CLI-Subscriptions.md
```diff
@@ -175,6 +175,49 @@ In the case of invalid options, print an error message (stderr) and set the exit
...
+  --item-types TEXT               One or more comma-separated item-types [required]
+  --geometry JSON                 GeoJSON of area to deliver subscription to. Use 
+                                  --clip only deliver pixels that match this area.
+                                  Can be a json string, filename, or '-' for stdin.
+  --clip                          Use the geometry GeoJSON as the clip geometry for 
+                                  the subscription.
+  --start-time DATETIME           Start date and time to begin subscription.
+  --end-time DATETIME             Date and time to end the subscription.
+  --rrule TEXT                    iCalendar recurrance rule to specify recurrances. 
+  --filter JSON                   A search filter can be specified a json string,
```
cholmes, line 203: This ideally would let people reuse the `planet data filter` command, but I'm not sure of the proper behavior if someone tries to use a filter with a geometry or a time range. Should figure out what the subscriptions API does in that case. 

There also might be some additional convenience we could enable by tapping into the geometry and date time filters, but that might get too complicated.

design-docs/CLI-Subscriptions.md
```diff
@@ -175,6 +175,49 @@ In the case of invalid options, print an error message (stderr) and set the exit
...
+  --clip                          Use the geometry GeoJSON as the clip geometry for 
+                                  the subscription.
+  --start-time DATETIME           Start date and time to begin subscription.
+  --end-time DATETIME             Date and time to end the subscription.
+  --rrule TEXT                    iCalendar recurrance rule to specify recurrances. 
+  --filter JSON                   A search filter can be specified a json string, 
+                                  filename, or '-' for stdin. 
+  --tools JSON                    Toolchain JSON. Can be a json string,
+                                  filename, or '-' for stdin.
+  --cloudconfig JSON              Credentials for cloud storage provider to
```
cholmes, line 207: We called this `cloudconfig` in orders, so keeping it consistent here, but the name of the block in both orders and subscriptions is `delivery`, so we could potentially call it that too...

design-docs/CLI-Subscriptions.md
tbarsballe: If someone is only provisioned with SkySat tasking, they might make a geometryless subscription to just deliver all their fulfilled tasks? I'm not certain if that works with how our provisioning actually functions, and is certainly an edge case, but something to think about.

More generally, since most Planet users are already provisioned with a constrained geometry within which they have access to imagery, anyone who a) wants all their imagery and b) doesn't need clipping might not use geometry in their subscriptions.

design-docs/CLI-Subscriptions.md
tbarsballe: From https://developers.planet.com/docs/subscriptions/source/#filter-details
> A subscription can leverage all filter types (with the exception of DateRangeFilter and GeometryFilter).

So the subscriptions API forbids use of filters containing geometry or time ranges

jreiberkyle:
I'm going to merge this so we have a starting point then propose changes to this as needed in another PR.

PR #695, opened 2022-09-16
Human readable invalid item_type report

### LINKED ISSUE #690: Report invalid item-types with more human-readable list
With `planet data search` if I do an improper item-type I get a response like:

```
Error: Invalid value for 'ITEM_TYPES': ['psscend'] should be one of {'PSScene', 'MOD09GA', 'MYD09GA', 'MYD09GQ', 'SkySatCollect', 'REOrthoTile', 'REScene', 'MOD09GQ', 'PSScene4Band', 'Sentinel1', 'Sentinel2L1C', 'SkySatScene', 'Landsat8L1G', 'PSOrthoTile', 'PSScene3Band'}
```

The { and ' characters make it feel like a dump of a program. It'd be better to have it look like a more natural list. 

`planet orders request` does it in a more readable way, so likely can just copy theirs: 
```
Error: Invalid value: item_type - 'skysatcollectd' is not one of 'PSScene', 'Landsat8L1G', 'SkySatScene', 'SkySatCollect', 'PSOrthoTile', 'REOrthoTile', 'PSScene3Band', 'Sentinel2L1C'.
```

kevinlacaille:
Closes #690

### DIFF
--- a/planet/cli/data.py
+++ b/planet/cli/data.py
@@ -19,7 +19,7 @@ import click
 
 from planet import data_filter, DataClient
 from planet.clients.data import SEARCH_SORT, SEARCH_SORT_DEFAULT, STATS_INTERVAL
-from planet.specs import get_item_types
+from planet.specs import get_item_types, SpecificationException
 
 from . import types
 from .cmds import coro, translate_exceptions
@@ -64,11 +64,12 @@ def assets_to_filter(ctx, param, assets: List[str]) -> Optional[dict]:
 
 def check_item_types(ctx, param, item_types) -> Optional[List[dict]]:
     # Set difference between given item types and all item types
-    set_diff = set([item.lower() for item in item_types]) - set(
+    invalid_item_types = set([item.lower() for item in item_types]) - set(
         [a.lower() for a in ALL_ITEM_TYPES])
-    if set_diff:
-        raise click.BadParameter(
-            f'{item_types} should be one of {ALL_ITEM_TYPES}')
+    if invalid_item_types:
+        raise SpecificationException(invalid_item_types,
+                                     ALL_ITEM_TYPES,
+                                     'item_type')
     else:
         return item_types
 
--- a/tests/unit/test_data_item_type.py
+++ b/tests/unit/test_data_item_type.py
@@ -13,7 +13,7 @@
 # limitations under the License.
 import logging
 import pytest
-import click
+from planet.specs import SpecificationException
 from planet.cli.data import check_item_types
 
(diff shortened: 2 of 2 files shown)


mkshah605 (approved):
LGTM!

planet/cli/data.py
```diff
@@ -64,11 +64,12 @@ def assets_to_filter(ctx, param, assets: List[str]) -> Optional[dict]:
 
 def check_item_types(ctx, param, item_types) -> Optional[List[dict]]:
     # Set difference between given item types and all item types
-    set_diff = set([item.lower() for item in item_types]) - set(
+    invalid_item_types = set([item.lower() for item in item_types]) - set(
         [a.lower() for a in ALL_ITEM_TYPES])
-    if set_diff:
-        raise click.BadParameter(
```
jreiberkyle, line 70: I think we want to still raise `click.BadParameter` here. Ultimately, the issue *is* still a bad parameter. And click handles `click.BadParameter` differently from other errors.

planet/cli/data.py
```diff
@@ -64,11 +64,12 @@ def assets_to_filter(ctx, param, assets: List[str]) -> Optional[dict]:
 
 def check_item_types(ctx, param, item_types) -> Optional[List[dict]]:
     # Set difference between given item types and all item types
-    set_diff = set([item.lower() for item in item_types]) - set(
+    invalid_item_types = set([item.lower() for item in item_types]) - set(
```
jreiberkyle, line 67: Can we somehow change the capitalization of the user-specified item-type to match the capitalization in ALL_ITEM_TYPES? I think there is code for that in the specs module. If we don't change it, the item capitalization will stay how the user specified it, but the API will not accept capitalization that doesn't exactly match what is in the spec.

jreiberkyle (commented):
got some suggestions for changes.

planet/cli/data.py
kevinlacaille: Oh yeah, in `specs.validate_item_type()`!

planet/cli/data.py
kevinlacaille: Good call!

kevinlacaille: If we use `specs.validate_item_type()` to validate the item types, as you suggested in [your comment](https://github.com/planetlabs/planet-client-python/pull/695/files/4935b11afab9a5c731831f6930d5bcc2ab5783fd#r977970613), then we'll need to raise the `SpecificationException`, as is raised in the `specs` module.

Issue #463, opened 2022-04-29
Don't require authentication to get CLI help


sgillies:
Let's not require authentication to get CLI help. Currently, we do. It's because its implemented in the `planet` root command, which is executed before a sub-command's help utility, yes?

```
$ planet orders list --help
Traceback (most recent call last):
  File "/home/seangillies/projects/planet-client-python/planet/auth.py", line 63, in from_file
    secrets = _SecretFile(filename).read()
  File "/home/seangillies/projects/planet-client-python/planet/auth.py", line 236, in read
    with open(self.path, 'r') as fp:
FileNotFoundError: [Errno 2] No such file or directory: '/home/seangillies/.planet.json'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/home/seangillies/projects/planet-client-python/venv2/bin/planet", line 11, in <module>
    load_entry_point('planet', 'console_scripts', 'planet')()
  File "/home/seangillies/projects/planet-client-python/venv2/lib/python3.8/site-packages/click/core.py", line 1128, in __call__
    return self.main(*args, **kwargs)
  File "/home/seangillies/projects/planet-client-python/venv2/lib/python3.8/site-packages/click/core.py", line 1053, in main
    rv = self.invoke(ctx)
  File "/home/seangillies/projects/planet-client-python/venv2/lib/python3.8/site-packages/click/core.py", line 1659, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/home/seangillies/projects/planet-client-python/venv2/lib/python3.8/site-packages/click/core.py", line 1656, in invoke
    super().invoke(ctx)
  File "/home/seangillies/projects/planet-client-python/venv2/lib/python3.8/site-packages/click/core.py", line 1395, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/home/seangillies/projects/planet-client-python/venv2/lib/python3.8/site-packages/click/core.py", line 754, in invoke
    return __callback(*args, **kwargs)
  File "/home/seangillies/projects/planet-client-python/venv2/lib/python3.8/site-packages/click/decorators.py", line 26, in new_func
    return f(get_current_context(), *args, **kwargs)
  File "/home/seangillies/projects/planet-client-python/planet/cli/orders.py", line 48, in orders
    ctx.obj['AUTH'] = planet.Auth.from_file()
  File "/home/seangillies/projects/planet-client-python/planet/auth.py", line 66, in from_file
[...]

Issue #643, opened 2022-07-20
Report PL_API_KEY usage in `planet auth value`


cholmes:
When debugging an early user's CLI it took some digging before I realized that PL_API_KEY was set and being used, since it didn't show up in `planet auth value`. It'd be good to give the user some indication what key will actually be used, over the one reported in `value`. My suggested route would be for `planet auth value` to say something like 'API Key set in PL_API_KEY environment variable:` and show that API key instead of the one they logged in with. But open to other ways to solve this.

cholmes:
Should also warn on `planet auth init`

tschaub:
I hit this same issue and agree that the CLI messages could make it easier to figure out what is going on.

When running `planet auth init` or `planet auth value`, if `PL_API_KEY` is set, it would be helpful if the CLI could say that it will be preferred over the `planet auth` configured value.  If `PL_API_KEY` is set to a different value, a "Warning: the PL_API_KEY environment variable is set to a different value than the one configured with `planet auth init`.  The PL_API_KEY environment variable has precedence and will be used instead of the configured key." would be helpful.

If something like this is printed from `planet auth value`, it should go to STDERR so that the value from STDOUT can still be used (e.g. `curl -u $(planet auth value): https://api.planet.com/compute/ops/orders/v2` should still work).

In addition, the message printed when `planet orders list` (or other requests) fails with a 401 could be more helpful.  The API responds with "API key not found" in the case of a 401.  From the perspective of a CLI user, this sounds like the CLI could not find my API key for use with the API.  Instead, a message like "Unable to list orders with the provided credentials" or something similar would be clearer.

sgillies:
I think I found some low hanging fruit. At https://github.com/planetlabs/planet-client-python/blob/main/planet/cli/auth.py#L59 (planet auth value) we could get the active auth object (which is not necessarily from a file, planet.auth by design supports more sources than that) that we have stored in ctx and report based on its properties. That's a good start. To complement this we could add a source property to the APIKeyAuth class https://github.com/planetlabs/planet-client-python/blob/main/planet/auth.py#L198. With the new property, `planet auth value` could print something like `0123456789abcdef0123456789abcdef (from /home/me/.planet.json)`.

It's also worth warning about discrepancy between keys in the secret file and environment :+1:

jreiberkyle:
[...]

PR #699, opened 2022-09-21
Get auth info as early as possible, store in context, use that

sgillies:
A possible solution for one of the issues reported in #643.

In a nutshell, the root planet command is now responsible for finding auth info and `planet auth value` reports on the active auth, whatever the source.

### DIFF
--- a/planet/cli/auth.py
+++ b/planet/cli/auth.py
@@ -17,6 +17,7 @@ import logging
 import click
 
 import planet
+from planet.exceptions import AuthException
 from .cmds import translate_exceptions
 
 LOGGER = logging.getLogger(__name__)
@@ -53,7 +54,11 @@ def init(ctx, email, password):
 
 
 @auth.command()
+@click.pass_context
 @translate_exceptions
-def value():
+def value(ctx):
     '''Print the stored authentication information'''
-    click.echo(planet.Auth.from_file().value)
+    try:
+        click.echo(ctx.obj['AUTH'].value)
+    except (AttributeError, KeyError):
+        raise AuthException('Command context lacks auth information.')
--- a/planet/cli/cli.py
+++ b/planet/cli/cli.py
@@ -19,6 +19,8 @@ import sys
 import click
 
 import planet
+from planet.auth import Auth
+from planet.exceptions import PlanetError
 
 from . import auth, collect, data, orders, subscriptions
 
@@ -44,6 +46,20 @@ def main(ctx, verbosity, quiet):
     # by means other than the `if` block below)
     ctx.ensure_object(dict)
     ctx.obj['QUIET'] = quiet
+
+    # Check authentication sources and store the highest priority in ctx.
+    # 1. Environment.
+    try:
+        auth_env = Auth.from_env()
+    except PlanetError:
+        auth_env = None
+    # 2. Secret file.
+    try:
+        auth_file = Auth.from_file()
+    except PlanetError:
+        auth_file = None
+
+    ctx.obj['AUTH'] = auth_env or auth_file
 
 
 def _configure_logging(verbosity):
--- a/planet/cli/data.py
+++ b/planet/cli/data.py
@@ -48,7 +48,6 @@ async def data_client(ctx):
               help='Assign custom base Orders API URL.')
 def data(ctx, base_url):
     '''Commands for interacting with the Data API'''
-    ctx.obj['AUTH'] = None
     ctx.obj['BASE_URL'] = base_url
 
 
--- a/planet/cli/orders.py
+++ b/planet/cli/orders.py
@@ -46,7 +46,6 @@ async def orders_client(ctx):
               help='Assign custom base Orders API URL.')
 def orders(ctx, base_url):
     '''Commands for interacting with the Orders API'''
-    ctx.obj['AUTH'] = None
     ctx.obj['BASE_URL'] = base_url
 
 
--- a/planet/cli/subscriptions.py
+++ b/planet/cli/subscriptions.py
@@ -14,9 +14,6 @@ from planet.clients.subscriptions import SubscriptionsClient
 @click.pass_context
 def subscriptions(ctx):
     '''Commands for interacting with the Subscriptions API'''
-    # None means that order of precedence is 1) environment variable,
-    # 2) secret file.
-    ctx.obj['AUTH'] = None
 
 
 # We want our command to be known as "list" on the command line but


planet/cli/auth.py
```diff
@@ -53,7 +54,11 @@ def init(ctx, email, password):
...
+@click.pass_context
 @translate_exceptions
-def value():
+def value(ctx):
     '''Print the stored authentication information'''
-    click.echo(planet.Auth.from_file().value)
+    try:
+        click.echo(ctx.obj['AUTH'].value)
+    except (AttributeError, KeyError):
+        raise AuthException('Command context lacks auth information.')
```
sgillies, line 64: This now reports on the active auth object.

planet/cli/cli.py
```diff
@@ -45,6 +47,20 @@ def main(ctx, verbosity, quiet):
...
+        auth_env = Auth.from_env()
+    except PlanetError:
+        auth_env = None
+    # 2. Secret file.
+    try:
+        auth_file = Auth.from_file()
+    except PlanetError:
+        auth_file = None
+
+    ctx.obj['AUTH'] = auth_env or auth_file
```
sgillies, line 62: Now the root "planet" command takes charge of auth. Previously, we deferred until calling the Session constructor.

sgillies: Warnings might be added here, but I think that could also be done in another PR.

planet/cli/data.py
```diff
@@ -48,7 +48,6 @@ async def data_client(ctx):
               help='Assign custom base Orders API URL.')
 def data(ctx, base_url):
     '''Commands for interacting with the Data API'''
-    ctx.obj['AUTH'] = None
```
sgillies, line 51: This is what deferral to the Session constructor looked like.

jreiberkyle:
Just a drive-by question: will this change affect CLI help at all? that is, `planet orders request --help` won't be upset if we don't have auth specified anywhere, will it?

sgillies:
@jreiberkyle oh yeah, that needs to be checked. Here's what I see after moving my secret file and making sure I don't have PL_API_KEY set.

```
$ planet auth value
Error: Auth information does not exist or is corrupted. Initialize with `planet auth init`.
$ planet orders request --help
Usage: planet orders request [OPTIONS]

  Generate an order request.

  This command provides support for building an order description used in
  creating an order. It outputs the order request, optionally pretty- printed.

Options:
  --name TEXT                     Order name. Does not need to be unique.
                                  [required]
  --bundle [analytic|analytic_udm2|analytic_3b_udm2|analytic_5b|analytic_5b_udm2|analytic_8b_udm2|visual|uncalibrated_dn|uncalibrated_dn_udm2|basic_analytic|basic_analytic_udm2|basic_analytic_8b_udm2|basic_uncalibrated_dn|basic_uncalibrated_dn_udm2|analytic_sr|analytic_sr_udm2|analytic_8b_sr_udm2|basic_uncalibrated_dn_nitf|basic_uncalibrated_dn_nitf_udm2|basic_analytic_nitf|basic_analytic_nitf_udm2|basic_panchromatic|basic_panchromatic_dn|panchromatic|panchromatic_dn|panchromatic_dn_udm2|pansharpened|pansharpened_udm2|basic_l1a_dn]
                                  Product bundle.  [required]
  --id TEXT                       One or more comma-separated item IDs.
                                  [required]
  --item-type TEXT                Specify an item type  [required]
  --clip JSON                     Clip feature GeoJSON. Can be a json string,
                                  filename, or '-' for stdin.
  --tools JSON                    Toolchain JSON. Can be a json string,
                                  filename, or '-' for stdin.
  --email                         Send email notification when order is
                                  complete.
  --cloudconfig JSON              Credentials for cloud storage provider to
                                  enable cloud delivery of data. Can be a json
                                  string, filename, or '-' for stdin.
  --stac / --no-stac              Do not request metadata to be in
                                  SpatioTemporal Asset Catalog (STAC) format.
  --pretty                        Format JSON output.
  --help                          Show this message and exit.
```

planet/cli/cli.py
jreiberkyle: Having `planet` take charge of auth is a great idea! To minimize code in the CLI wrapper and make it easier to maintain and understand default auth behavior, what about adding the logic for getting auth to the `auth` module, as something like `Auth.default() -> AuthType` then using that function here and also in the `Session` class.

planet/cli/orders.py
```diff
@@ -46,7 +46,6 @@ async def orders_client(ctx):
               help='Assign custom base Orders API URL.')
 def orders(ctx, base_url):
     '''Commands for interacting with the Orders API'''
-    ctx.obj['AUTH'] = None
```
jreiberkyle, line 49: nice to be getting rid of these per-client calls to auth. I seem to remember those going in to resolve #463 so I am surprised they are no longer needed

jreiberkyle:
In mulling over this, I can't recall why we need to manage auth information in the CLI. The priority of auth sources is the same for the CLI and API. Why not have the `Session` obtain default auth every time it is initialized (which will be only once per call anyway)? The logic for obtaining default auth can be a method of the `auth.Auth` class, and called by both `http.Session` and `planet auth value`. We should provide logging which reports the source of the auth, probably at `INFO` level. Then the user could obtain the auth source through `planet --verbosity DEBUG ...` e.g. `planet --verbosity DEBUG auth value` or `planet --verbosity DEBUG data search ..`

sgillies:
@jreiberkyle I like to think of software as an onion (this is not my original analogy). On the outside is the filesystem and the environment, which is where our auth sources are (a secret file or an environment variable). The CLI is the outermost layer of our onion and necessarily has to interact with the filesystem and environment. This package's http module is a few layers down in our onion and shouldn't (in my ideal onion world) be interacting with the filesystem and environment. It should be getting auth information from the next layer out (a service client), which in turn gets it from the next layer out (the CLI). I've worked on projects with no such boundaries and they always suffer for the lack of them. That's why I changed the code to have the outermost layer of our software be responsible for getting the secret and passing it further down.

jreiberkyle:
@sgillies That logic makes sense. In keeping with this approach, then, would we want to remove the logic for obtaining auth information (from env and file) from within `Session` as well?

sgillies:
@jreiberkyle no, I think we should keep it. In some applications (like a lot of our usage examples and notebooks), a Session can be the outermost layer, yes? Like

```
async with Session() as session:
    ...
```

It's pragmatic to have it. But let's not use it like that in our CLI, that's all I'm saying.

jreiberkyle:
Ahh okay I understand. It's context-dependent. That makes sense. I'm on board with this approach.

jreiberkyle:
@sgillies for the benefit of my understanding: 

here, in "a few layers down in our onion", our onion == CLI, right?

>  This package's http module is a few layers down in our onion and shouldn't (in my ideal onion world) be interacting with the filesystem and environment. It should be getting auth information from the next layer out (a service client), which in turn gets it from the next layer out (the CLI). 

but for the API, the http module is the outermost layer of the onion.

So, for the CLI application, the CLI should be obtaining the auth info, but for the python application, it is ok for the http module to obtain the auth information - because each is the outermost layer of the onion for it's own application.

I can follow this conceptually, though it does seem confusing since the http module is the same in both applications, so we pay for this 'auth obtained at outer layer of the onion' with 'logic for obtaining auth in two places in the logic chain'. So, your experience with how this tradeoff has come out in favor of the 'outer layer' approach is helpful. Would you be willing to elaborate a little on your experience alluded to in the quote below?

> I've worked on projects with no such boundaries and they always suffer for the lack of them. That's why I changed the code to have the outermost layer of our software be responsible for getting the secret and passing it further down.

jreiberkyle:
this has been used as input to #809, see that PR for implementation

Issue #628, opened 2022-06-30
update subscriptions api/cli to match interface of other clients and reuse common code


jreiberkyle:
The subscriptions client interface and the data/orders clients interfaces and implementations differ in a few ways that make it more difficult for users and for maintenance of the clients.

To do:
- In CLI reuse pretty and limit options
- in client, construct urls from a base_url pulled from a global (which in turn is constructed from the planet.base_url constant). override this with the tests. Have all test mocked urls use this url as well.

See #627 for a solid start on these changes

PR #701, opened 2022-10-03
use flake8 defaults

### LINKED ISSUE #702: use flake8 defaults
We have some customizations of flake8 that are mostly unused, except for a few line-length exceptions. These are easy to fix and it's beneficial to impose some line length limit, why not just use the default.

jreiberkyle:
Use flake8 defaults, fix a few line length errors

Closes #702

### DIFF
--- a/planet/clients/subscriptions.py
+++ b/planet/clients/subscriptions.py
@@ -119,7 +119,8 @@ class SubscriptionsClient:
             APIError: on an API server error.
             ClientError: on a client error.
         """
-        url = f'https://api.planet.com/subscriptions/v1/{subscription_id}/cancel'
+        root_url = 'https://api.planet.com/subscriptions/v1'
+        url = f'{root_url}/{subscription_id}/cancel'
         req = Request(url, method='POST')
 
         try:
@@ -222,7 +223,8 @@ class SubscriptionsClient:
             ITEMS_KEY = 'results'
 
         params = {'status': [val for val in status or {}]}
-        url = f'https://api.planet.com/subscriptions/v1/{subscription_id}/results'
+        root_url = 'https://api.planet.com/subscriptions/v1'
+        url = f'{root_url}/{subscription_id}/results'
         req = Request(url, params=params)
 
         try:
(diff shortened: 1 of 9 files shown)


planet/clients/subscriptions.py
```diff
@@ -119,7 +119,8 @@ async def cancel_subscription(self, subscription_id: str) -> None:
             APIError: on an API server error.
             ClientError: on a client error.
         """
-        url = f'https://api.planet.com/subscriptions/v1/{subscription_id}/cancel'
+        root_url = 'https://api.planet.com/subscriptions/v1'
```
kevinlacaille, line 122: Consider having `ROOT_URL` as a global instead of declaring it twice?

kevinlacaille (approved):
All sensible changes - looks good to me!

planet/clients/subscriptions.py
jreiberkyle: Yep. That will be done as a part of #628

sgillies: I'd prefer the root or base url be a SubscriptionsClient class attribute. It would have the advantages of a global (ease of patching for tests) but would also be scoped a little closer to the only place it is used - in client class methods.

kevinlacaille: That's a great call

sgillies: It looks like the intent is to break a 81 character line into two lines less than 80 characters? I don't think it's worth doing, but Python's implicit string concatenation is probably better than adding a new variable. A future reader might mistake the extra variable assignment as being significant, when it's only incidental. Implicit string concatenation would go like this:

```python
url = ('https://api.planet.com/subscriptions/v1'
       f'{subscription_id}/cancel')
```

This would be even cleaner and tidier

```python
url = '{self.base_url}/{subscription_id}/cancel'
```

sgillies (commented):
@jreiberkyle my only concern is about extra variable assignments being mistaken by future readers as being significant instead of being only incidental. So maybe try to avoid them as much as possible and document them where they are used? Just a suggestion, not a requirement.

planet/clients/subscriptions.py
jreiberkyle: these are good points and I fully agree with them. root_url is arbitrary and going to using `self.base_url` is preferred. I created #705 to address this and referenced this discussion.

PR #707, opened 2022-10-03
Implement data api asset download and supporting functions

### LINKED ISSUE #499: implement data api asset functions
implement the following data api asset functions:
* list_assets
* get_asset
* get_asset_type
* activate_asset
* wait_asset
* download_asset

jreiberkyle:
Implements data asset functions, everything needed to download and validate an asset

Closes #499

### DIFF
--- a/docs/guide.md
+++ b/docs/guide.md
@@ -334,6 +334,42 @@ the context of a `Session` with the `DataClient`:
 >>> asyncio.run(main())
 ```
 
+#### Downloading an Asset
+
+Downloading an asset is a multi-step process involving: activating the asset,
+waiting for the asset to be active, downloading the asset, and, optionally,
+validating the downloaded file.
+
+
+With wait and download, it is often desired to track progress as these
+processes can take a long time. Therefore, in this example, we use a simple 
+print command to report wait status. `download_asset` has reporting built in.
+
+```python
+>>> async def download_and_validate():
+...     async with Session() as sess:
+...         cl = DataClient(sess)
+... 
+...         # get asset description
+...         item_type_id = 'PSScene'
+...         item_id = '20221003_002705_38_2461'
+...         asset_type_id = 'ortho_analytic_4b'
+...         asset = await cl.get_asset(item_type_id, item_id, asset_type_id)
+...         
+...         # activate asset
+...         await cl.activate_asset(asset)
+... 
+...         # wait for asset to become active
+...         asset = await cl.wait_asset(asset, callback=print)
+... 
+...         # download asset
+...         path = await cl.download_asset(asset)
+... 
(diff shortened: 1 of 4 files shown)


tests/integration/test_orders_api.py
```diff
@@ -559,6 +559,41 @@ async def test_download_asset_md(tmpdir, session):
...
+                               request='donotcloneme')
+    respx.get(dl_url).return_value = mock_resp
+
+    cl = OrdersClient(session, base_url=TEST_URL)
+    filename = await cl.download_asset(dl_url, directory=str(tmpdir))
+
+    assert Path(filename).name == 'img.tif'
+    assert os.path.isfile(filename)
+
+
```
jreiberkyle, line 596: this is just a move closer to it's sibling test

cholmes (approved):
This looks good to me, but it looks like it's not the CLI yet (which is totally fine, python api is the target), which just means I can't really do any testing of this on my own.

docs/guide.md
```diff
@@ -334,6 +334,42 @@ the context of a `Session` with the `DataClient`:
...
+waiting for the asset to be active, downloading the asset, and, optionally,
+validating the downloaded file.
+
+
+With wait and download, it is often desired to track progress as these
+processes can take a long time. Therefore, in this example, we use a simple 
+print command to report wait status. `download_asset` has reporting built in.
+
+```python
+>>> async def download_and_validate():
```
kevinlacaille, line 349: I am against having the added "`>>>`"s and "`...`"s in example code because it makes it difficult for users to copy and paste and immediately use. We're also inconsistent with its usage. We add it to many of our examples, but don't have it in the checksum example (line 262 in this MD).

jreiberkyle: I'm against it too. It is included here for consistency and we need to do a sweep of the docs to remove these. I'm not sure why it isn't in the checksum example, it probably slipped through a PR. But that isn't a part of this PR.

jreiberkyle:
@kevinlacaille anything else you'd like to bring up in the review or do you feel ok with approving?

docs/guide.md
sgillies: The tricky thing is that `>>>` is kinda normalized in Python and numpy. See https://docs.python.org/3/tutorial/introduction.html#numbers and https://numpydoc.readthedocs.io/en/latest/format.html#examples. So without some doc linting, this usage might continue to creep in. For what it's worth, the Python doc source includes the `>>>` prompt (see https://github.com/python/cpython/blob/3.10/Doc/tutorial/introduction.rst#numbers) and the website has a js widget to hide it if you want to copy it without.

kevinlacaille (approved):
LGTM!

PR #722, opened 2022-10-20
Downloading multiple assets in parallel example

### LINKED ISSUE #719: Provide example with async, parallelized downloads, with asset activation & download
We recently added asset activation and download, let's create an example of using async to parallelize many downloads at once.

This is basically combining [orders_create_and_download_multiple_orders.py](https://github.com/planetlabs/planet-client-python/blob/main/examples/orders_create_and_download_multiple_orders.py) with the [downloading an asset example](https://github.com/planetlabs/planet-client-python/blob/main/docs/guide.md#downloading-an-asset) in the docs user guide to create a new script in the examples folder.

kevinlacaille:
Created an example of downloading multiple assets in parallel. This example is an expansion on the example given in [the guide](https://github.com/planetlabs/planet-client-python/blob/main/docs/guide.md#downloading-an-asset).

The script takes in two distinct PSScene item IDs with different asset types, gets their asset descriptions, activating them, waiting for them to become active, downloads them, then validating the checksums of downloaded files, in parallel.

Closes #719

### DIFF
--- a/examples/download_multiple_assets.py
+++ b/examples/download_multiple_assets.py
@@ -11,7 +11,7 @@
 # WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
 # License for the specific language governing permissions and limitations under
 # the License.
-"""Example of downloading multiple assets in paralel
+"""Example of downloading multiple assets in parallel
 
 This is an example of getting two assets, activating them, waiting for them to
 become active, downloading them, then validating the checksums of downloaded
--- a/examples/download_multiple_assets.py
+++ b/examples/download_multiple_assets.py
@@ -1,0 +1,59 @@
+# Copyright 2022 Planet Labs PBC.
+#
+# Licensed under the Apache License, Version 2.0 (the "License"); you may not
+# use this file except in compliance with the License. You may obtain a copy of
+# the License at
+#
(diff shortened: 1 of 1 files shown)


kevinlacaille:
A test for a randomly generated number fails, see #723.

examples/download_multiple_assets.py
```diff
@@ -0,0 +1,67 @@
...
+# WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+# License for the specific language governing permissions and limitations under
+# the License.
+"""Example of downloading multiple assets in parallel
+
+This is an example of getting two assets, activating them, waiting for them to
+become active, downloading them, then validating the checksums of downloaded
+files.
+
+[Planet Explorer](https://www.planet.com/explorer/) was used to define
```
jreiberkyle, line 20: Is this true for this script and where are AOIs used here?

jreiberkyle (approved):
Looks great runs great! My one ask is that the PR title be phrased for a Changelog entry

examples/download_multiple_assets.py
kevinlacaille: Certainly was for the wildfire AOI, unsure about the river AOI, I took that from the guide, but I assume so.

PR #740, opened 2022-11-03
add bug report issue template

jreiberkyle:


cholmes (approved):
So this is very 'bug'-centric, and it seems less clear what to do if you just have like a feature request. I do think it's a very nice bug report template, nudging people to provide all the info to make a really good bug report. But it'd be good to have at least some advice for just filing some idea for an improvement, or a new feature, or a documentation fix, etc. I feel like few of our issues to date are really 'bugs', so most wouldn't fit in here. Could try to make it more general, or could just give some instructions for what to do if it's not a bug. 

I'll approve this, since we have nothing now. But it'd be nice to at least add some instructions for non-bugs.

jreiberkyle:
yeah this is bug-specific. feature request / question forms are coming. I am using the GH ui to create the template, which just allows one template per workflow.

jreiberkyle:
closing in favor of #742

Issue #752, opened 2022-11-10
add support for source_type parameter in Orders client list orders


jreiberkyle:
The list orders endpoint provides a `source_type` parameter to include source-types in the the order list. By default this is 'scenes', which only includes scene orders, but with the source_type parameter this can be changed to 'basemaps' or 'scenes,basemaps'. Add support for this in the Orders client list_orders function.

ischneider:
This appears done via 71562d358d

PR #768, opened 2022-11-14
List orders now lists both scenes and basemaps order types

### LINKED ISSUE #767: Include basemaps orders in `planet order list` output

Planet now supports [basemaps in orders](https://developers.planet.com/apis/orders/basemaps/). This works great with all the core orders commands, except for `list`. You can `get` a basemaps order, `wait`, and `download` it. 
 
To have `planet orders list` return basemaps and scene orders.

The core API team wanted to make everything backwards compatible, so they kept the GET at https://api.planet.com/compute/ops/orders/v2/ to just the scene orders. They introduced a parameter called `source_type`, that can take values like `all`, `basemaps`, `scenes`, `basemaps,scenes`. 

I propose we just return `all`, since the SDK and CLI just output the JSON directly, so returning an extra JSON attribute doesn't effect anything. Our `list` should aim to show all the options at the end point.

We could consider supporting the `source_type` parameter as an option, with a default to `all`. But that could also be ticketed as its own issue.

**Additional context**

cholmes:
**Related Issue(s):**

Closes #767 

**Proposed Changes:**

1. List orders now lists both scenes and basemaps order types.

Not intended for changelog:

1. 

**Diff of User Interface**

Old behavior: `list_orders` python call and `planet orders list` would only return orders for scenes

New behavior: `list_orders` python call and `planet orders list` would only return orders for scenes and basemaps

**PR Checklist:**

### DIFF
--- a/planet/clients/orders.py
+++ b/planet/clients/orders.py
@@ -481,15 +481,14 @@ class OrdersClient:
             planet.exceptions.ClientError: If state is not valid.
         """
         url = self._orders_url()
+        params = {"source_type": "all"}
 
         if state:
             if state not in ORDER_STATE_SEQUENCE:
                 raise exceptions.ClientError(
                     f'Order state ({state}) is not a valid state. '
                     f'Valid states are {ORDER_STATE_SEQUENCE}')
-            params = {"state": state}
-        else:
-            params = None
+            params['state'] = state
 
         response = await self._session.request(method='GET',
                                                url=url,
--- a/tests/integration/test_orders_api.py
+++ b/tests/integration/test_orders_api.py
@@ -123,7 +123,7 @@ async def test_list_orders_basic(order_descriptions, session):
 @respx.mock
 @pytest.mark.asyncio
 async def test_list_orders_state(order_descriptions, session):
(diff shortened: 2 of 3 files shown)


jreiberkyle:
Also closes #752 (looks like that got duped - is that how one shortens 'duplicated'?) 😆

jreiberkyle:
oh wait, this is a little different. This PR makes the default behavior be to list all orders. #752 adds a new argument to list which allows filtering. It makes sense to have listing all be the default and we'll keep #752 open for later on.

jreiberkyle (approved):
Your PR passed CI. Your code is clean. You updated the tests so this new functionality is tested. Excellent work @cholmes! 🚀 

Now that you demonstrated you can fold the laundry, you may get called in to take on a few socks or shirts now and then. 😆

cholmes:
> It makes sense to have listing all be the default and we'll keep https://github.com/planetlabs/planet-client-python/issues/752 open for later on.

Sounds great, and in line with what I was thinking - get this one in now, and then later make source_type an explicit parameter. It doesn't feel super pressing to make it a parameter since it's pretty easy to filter client side. Didn't realize you had a ticket for it, I'll schedule it on a milestone.

Issue #779, opened 2022-11-17
provide immediate feedback on success of paged request without returning private classes


jreiberkyle:
When paged requests are made, the user would expect immediate feedback on success or failure of the request and would expect to receive an iterator over the paged results (aka, 'search' for get_searches, 'subscription' for get_subscriptions, 'item' for run_search, etc). 

Currently, the user gets immediate feedback on success or failure but they receive subclasses of `Paged` (which is an AsyncIterator). This should be changed so they just get an AsyncIterator.

This can be achieved by changing how paging is achieved in the respective functions. That is, instead of
```python
        response = await self._session.request(method='GET',
                                               url=url,
                                               params=params)
        return Orders(response, self._session.request, limit=limit)
```
The code would be:
```python
        response = await self._session.request(method='GET',
                                               url=url,
                                               params=params)
        return _session.page(response, limit=limit, next_path='_links.next', results_path='items')
```

```python
def page(response, limit=0, next_path='_links.next', results_path='items') -> typing.AsyncGenerator[dict, None]
```

The benefit of this approach over #706 is that it provides immediate feedback on success or failure of a request (due to the response of the first page being received before creating the iterator).

Issue #476, opened 2022-05-03
Have `Paged` submit first request on initialization


jreiberkyle:
To avoid a delayed request error that wouldn't be noticed until the generator is actually exercised, have `Paged` submit the first page request upon initialization and store the response for use in the generator. ([ref](https://github.com/planetlabs/planet-client-python/pull/462#discussion_r861900299_))

jreiberkyle:
One solution - have Paged take in a Response instead of a Request:

```python
class Paged:
    """Asynchronous iterator over results in a paged resource.

    Each returned result is a json dict.

    Parameters:
        response: Response from submitting a request for the first page.
        do_request_fcn: Function for submitting a request. Takes as input
            a planet.api.models.Request and returns planet.api.models.Response.
        limit: Maximum number of results to obtain and return. Defaults to
            all results.
    """
    LINKS_KEY = '_links'
    NEXT_KEY = 'next'
    ITEMS_KEY = 'items'

    def __init__(self,
                 response: Response,
                 do_request_fcn: typing.Callable,
                 limit: int = None):

        # NOTE: to avoid delayed raising of API errors from the initial API
        # request, the first page should be retrieved before the iterator is
        # processed. However, we cannot retrieve the first page in the __init__
        # fcn because __init__ must be sync while retrieval is async. Therefore
        # require the initial page to already be retrieved and take in the
        # response instead of taking in the request.
        self.first_page = response.json()
        self._items = self.first_page[self.ITEMS_KEY]

        self._do_request = do_request_fcn
        self._pages = None

        self.i = 0
        self.limit = limit

    def __aiter__(self):
        return self

    async def __anext__(self) -> dict:
        """Get next item."""
        # This was implemented because traversing _get_pages()
        # in an async generator was resulting in retrieving all the
        # pages, when the goal is to stop retrieval when the limit
        # is reached
        if self.limit is not None and self.i >= self.limit:
            raise StopAsyncIteration

        try:
            item = self._items.pop(0)
            self.i += 1
        except IndexError:
            self._pages = self._pages or self._get_next_pages()
            page = await self._pages.__anext__()
            self._items = page[self.ITEMS_KEY]
            try:
                item = self._items.pop(0)
                self.i += 1
            except IndexError:
                raise StopAsyncIteration

        return item

    async def _get_next_pages(self):
        next_url = self._get_next_url(self.first_page)

        while (next_url):
            LOGGER.debug('getting next page')
            request = Request(next_url, 'GET')
[...]

PR #785, opened 2022-11-21
Move to async generator functions for paged results

### LINKED ISSUE #706: Use PEP 525 async generator to improve data and orders client design
[PEP525](https://peps.python.org/pep-0525/) added async generators to Python 3.6. An async generator is an `async def` function or method that uses `yield` instead of `return`. Using it for SDK methods that have multiple results gives us a cleaner design that is in harmony with the designs of async Python software like aiohttp and httpx.

Here is a short bit of code that uses the data client and orders client together. We search the catalog for items using DataClient and then use a list of item ids in creating an order with OrdersClient.

```python
async def main():
    """Use DataClient and OrdersClient to search for and order scenes.
    """
    async with Session() as sess:
        data_client = DataClient(sess)

        # A reader might be led to believe that since we're awaiting the results
        # of search(), the return value has all the results materialized in it.
        search_result_iterator = await data_client.search(["PSScene"], filter_config)

        # But it does not. We still need to asynchronously iterate.
        item_ids = [item["id"] async for item in search_result_iterator]

        client = OrdersClient(sess)
        order_request = {
            "name": "my_order",
            "products": [
                {
                    "item_ids": item_ids,
                    "item_type": "PSScene",
                    "product_bundle": "analytic_udm2",
                }
            ]
        }
        order = await client.create_order(order_request)

        # Wait for order to be completed.
        order_id = order["id"]
        state = await client.wait(order_id, state="success")
        ...
```

To reiterate what I wrote in a comment, a reader of the code might be led to believe that since we're awaiting the results of search(), the return value has all the results already materialized in it. And requiring further iteration then looks confusing. I think it looks odd to await a method call and then asynchronously iterate over the return value. And Python doesn't require it. Neither httpx's [aiter_bytes](https://www.python-httpx.org/async/#streaming-responses) nor aiohttp's [iter_chunked](https://docs.aiohttp.org/en/stable/client_quickstart.html#streaming-response-content) return chunkers. They are async generators.

As an improvement, I propose to make DataClient.search an async generator and allow more concise usage like this:

```python
async def main():
    """Use DataClient and OrderClient to search for and order scenes.
    """
    async with Session() as sess:
        data_client = DataClient(sess)
        item_ids = [item["id"] async for item in data_client.search(["PSScene"], filter_config)]
        ...
```

The new implementation would be almost trivial. It would mean, for DataClient, changing the code at 
 https://github.com/planetlabs/planet-client-python/blob/main/planet/clients/data.py#L185 from

```python
        return Items(request, self._do_request, limit=limit)
```

to

```python
        async for item in Items(request, self._do_request, limit=limit):
            yield item
```

This is not unlike what httpx does here: https://github.com/encode/httpx/blob/8152c4facd0f71e0f376287e41a0810a60fec9c6/httpx/_models.py#L921. It *does* have a chunker, but it's used internally, not by httpx users.

What makes the implementation not trivial is updating tests. That's where most of the work is. I think it's worth doing. 

To support less common situations where one really needs a reference to the search results generator, say to interleave calls to `next(search_results)` with other function calls, one can assign `search_results = client.search()` (without `await`) and then use the awaitable next method of the async generator.

```python
import asyncio

async def func():
    for i in range(10):
        yield i

async def main():
    gen = func()
    print("first")
    # In Python 3.10 this is: print(await anext(gen))
    print(await gen.__anext__())
    print("second")
    print(await gen.__anext__())

asyncio.run(main())
first
0
second
1
```

jreiberkyle:
**Related Issue(s):**

Closes #706 

Moves to wontfix: #476 and #779

**Proposed Changes:**

1. The following functions which handle paged results were changed to async generator functions and renamed:
  * OrdersClient.list_orders -> list_orders_aiter
  * DataClient.search -> search_aiter
  * DataClient.run_search -> run_search_aiter
  * DataClient.list_searches -> list_searches_aiter
3. The following were renamed to clarify that they are async generator functions:
  * SubscriptionsClient.list_subscription_results -> list_subscription_results_aiter
  * SubscriptionsClient.list_subscriptions -> list_subscriptions_aiter

Not intended for changelog:

1. 

**Diff of User Interface**

Old behavior:

e.g. for OrdersClient.list_orders:

```python
orders = await cl.list_orders()
orders_list = [o async for o in orders]
```

New behavior:

*note that cl.list_orders_aiter() is not awaited*

```python
orders_aiter = cl.list_orders_aiter()
orders_list = [o async for o in orders_aiter]
```

**PR Checklist:**

@JuLeeAtPlanet @cholmes

### DIFF
--- a/docs/python/sdk-guide.md
+++ b/docs/python/sdk-guide.md
@@ -109,11 +109,17 @@ Otherwise, the JSON blob is a list of the individual results.
 >>> async def main():
 ...     async with Session() as sess:
 ...         client = OrdersClient(sess)
-...         orders = client.list_orders()
-...         orders_list = collect(orders)
+...         orders_aiter = client.list_orders_aiter()
+...         orders_list = collect(orders_aiter)
 ...
 >>> asyncio.run(main())
 
+```
+
+Alternatively, these results can be converted to a list directly with
+
+```python
+>>>         orders_list = [o async for o in client.list_orders_aiter()]
 ```
 
 
--- a/planet/cli/data.py
+++ b/planet/cli/data.py
@@ -272,12 +272,12 @@ async def search(ctx, item_types, filter, limit, name, sort, pretty):
     parameter will be applied to the stored quick search.
     """
     async with data_client(ctx) as cl:
-        items = await cl.search(item_types,
-                                filter,
-                                name=name,
-                                sort=sort,
-                                limit=limit)
-        async for item in items:
+        items_aiter = cl.search_aiter(item_types,
+                                      filter,
+                                      name=name,
+                                      sort=sort,
+                                      limit=limit)
+        async for item in items_aiter:
             echo_json(item, pretty)
 
 
--- a/planet/cli/orders.py
+++ b/planet/cli/orders.py
@@ -67,8 +67,8 @@ async def list(ctx, state, limit, pretty):
     optionally pretty-printed.
     '''
     async with orders_client(ctx) as cl:
-        orders = await cl.list_orders(state=state, limit=limit)
-        async for o in orders:
+        orders_aiter = cl.list_orders_aiter(state=state, limit=limit)
+        async for o in orders_aiter:
             echo_json(o, pretty)
 
 
--- a/planet/cli/subscriptions.py
+++ b/planet/cli/subscriptions.py
@@ -42,8 +42,9 @@ async def list_subscriptions_cmd(ctx, status, limit, pretty):
     """Prints a sequence of JSON-encoded Subscription descriptions."""
     async with CliSession(auth=ctx.obj['AUTH']) as session:
         client = SubscriptionsClient(session)
-        filtered_subs = client.list_subscriptions(status=status, limit=limit)
-        async for sub in filtered_subs:
+        subs_aiter = client.list_subscriptions_aiter(status=status,
+                                                     limit=limit)
+        async for sub in subs_aiter:
             echo_json(sub, pretty)
 
 
@@ -157,8 +158,8 @@ async def list_subscription_results_cmd(ctx,
     """Gets results of a subscription and prints the API response."""
     async with CliSession(auth=ctx.obj['AUTH']) as session:
         client = SubscriptionsClient(session)
-        filtered_results = client.get_results(subscription_id,
-                                              status=status,
-                                              limit=limit)
-        async for result in filtered_results:
+        results_aiter = client.get_results_aiter(subscription_id,
+                                                 status=status,
+                                                 limit=limit)
+        async for result in results_aiter:
             echo_json(result, pretty)
--- a/planet/clients/data.py
+++ b/planet/clients/data.py
@@ -332,10 +332,10 @@ class DataClient:
         response = await self._session.request(method='GET', url=url)
         return response.json()
 
-    async def run_search(self,
-                         search_id: str,
-                         limit: int = 100) -> AsyncIterator[dict]:
-        """Execute a saved search.
+    async def run_search_aiter(self,
+                               search_id: str,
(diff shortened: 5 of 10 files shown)


docs/python/sdk-guide.md
```diff
@@ -109,13 +109,19 @@ Otherwise, the JSON blob is a list of the individual results.
...
+...         orders_list = collect(orders_aiter)
 ...
 >>> asyncio.run(main())
 
 ```
 
+Alternatively, these results can be converted to a list directly with
+
+```python
+>>>         orders_list = [o async for o in client.list_orders_aiter()]
```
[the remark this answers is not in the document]
kevinlacaille: Let's get rid of the large, confusing blank space.

jreiberkyle:
@sgillies and @kevinlacaille - going ahead and merging this, mentioning you for an FYI

kevinlacaille:
I'm not sure I see the value in adding `_aiter` to function & variable names. It seems like it only adds more confusion and verbosity. Especially since we're just appending `_aiter` to these names, it seems redundant. For instance, since there is no longer a `list_orders()` function why do we need to call it `list_orders_aiter()`?

kevinlacaille (commented):
I'm a little uncertain about the `_aiter` suffix. See my comment above.

docs/python/sdk-guide.md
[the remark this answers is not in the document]
jreiberkyle: yeah it's not great but I included it so the line could be plugged into the code example above. @JuLeeAtPlanet what are your thoughts on this?

jreiberkyle:
@kevinlacaille I guess I can't reply to your review comment directly like in-line comments, which is unfortunate, but I have added an entry in the PR description that explains the reasoning behind the name change (see **Why the function name changes?** in the Proposed Changes section) . Thank you for digging deeper, it seems appropriate to provide more context around that change as it has a big impact.

docs/python/sdk-guide.md
```diff
@@ -109,13 +109,19 @@ Otherwise, the JSON blob is a list of the individual results.
 >>> async def main():
 ...     async with Session() as sess:
 ...         client = OrdersClient(sess)
-...         orders = client.list_orders()
-...         orders_list = collect(orders)
+...         orders_aiter = client.list_orders_aiter()
+...         orders_list = collect(orders_aiter)
```
sgillies, line 113: @jreiberkyle @kevinlacaille let's not lead users to assign client.list_orders_aiter() to a variable. The best usage is just one single line.

```
orders = [order for order in client.list_orders()]
```

Lets recommend this and do it everywhere in this project.

I'm also :-1: on `_aiter`. There's no synchronous alternative and the fact that it's a generator is documented in the method docstring and in the type hint.

planet/cli/data.py
```diff
@@ -272,12 +272,12 @@ async def search(ctx, item_types, filter, limit, name, sort, pretty):
...
-                                name=name,
-                                sort=sort,
-                                limit=limit)
-        async for item in items:
+        items_aiter = cl.search_aiter(item_types,
+                                      filter,
+                                      name=name,
+                                      sort=sort,
+                                      limit=limit)
+        async for item in items_aiter:
```
sgillies, line 280: Lines 275 could be a single line and improve clarity.

docs/python/sdk-guide.md
sgillies: A note about the history of "iter" in names in In Python: whenever we've seen "iter" used as a prefix on method names, it's been part of a transition. Once, Python dicts had `keys()`, and it was a list. But then it was decided to make the keys method a generator function. To help the transition, `iterkeys()` was added. It returned a generator while `keys()` returned a list. And then in Python 3, `keys()` returned a generator and `iterkeys()` was deprecated.

Issue #396, opened 2022-02-04
in CLI, add non-login path to storing API key


jreiberkyle:
Currently, the CLI only allows one path to storing the API key: login with username and password. Provide an alternate path for storing the API key, e.g. something like `planet auth key set <>`

PR #809, opened 2022-12-02
Clarify auth source, CLI only sources from config file

### LINKED ISSUE #808: Clarify auth source with CLI only sourcing from config file and additional log messages
Users and maintainers troubleshooting with the users alike have been stuck wondering why their Planet SDK authentication was not recognized when they had just performed the recommended `planet auth init` ([example](https://github.com/planetlabs/planet-client-python/issues/643)). Eventually they determined that the user had the PL_API_KEY environment variable set, which was not updated when `planet auth init` was run (aka, configuration file), but takes priority over the value stored with `planet auth init`.

Use #699 as inspiration for how to implement CLI-specific auth.

Change the CLI authentication source to differ from the API source in the following way: the CLI source is only the configuration file, environment variables are ignored. Further, add `planet auth set` to allow an alternative to `planet auth login` for storing auth information. `planet auth value` would always report the CLI auth value, which is from the configuration file.

To support the user in understanding and maintaining auth sources, do the following:
* When CLI login (or set) is run, determine if the environment variable exists and, if so, raise a warning with a suggestion for how to update the environment variable from the CLI configuration (i.e. using `planet auth value`)
* Clarify the Python library auth source (which could be environment variable or configuration file) by adding an INFO log specifying the source.

This pattern is inspired by, and closely follows, the gcloud authorization pattern. That is, the CLI exclusively uses the authentication information managed by the CLI. The Python API first looks for an environment variable then falls back to authentication information managed by the CLI.

The primary alternative to this solution is to have the CLI obtain authentication information from an environment variable first, then the configuration file, and to have the CLI report the  authentication that would actually be used (aka, if t[here](https://github.com/planetlabs/planet-client-python/pull/699) is an environment variable, report that, if not, report what is in the configuration file). This solution is documented here. This alternative solution has the benefit of already having a draft PR. It has the additional benefit of being similar to the AWS authentication scheme. It has the drawback that it is a different authentication scheme than gcloud and maxar (though this is inevitable with aws and google/maxar having different schemes). It has the additional drawback that, although it is close to AWS, it differs from AWS in ways that impact usage. 

While AWS has the CLI use authentication information from the environment variable first, then the configuration, it only reports what is in the configuration file. The AWS command is `aws configure get` which makes it pretty clear it is configuration specific, while our CLI command is `planet auth value` which is more auth-specific. The alternative solution also has the CLI use authentication information from the environment variable first, but differs in that it reports the primary authentication information, which will be the environment variable if it is set. This approach makes it difficult to update the environment variable if it exists (a likely scenario when using a docker environment e.g. for notebooks). The currently-recommended approach is: Set configuration with `planet auth init`, set environment variable with `PL_API_KEY=(planet auth value)` then share the environment variable with docker. When `planet auth value` reports the environment variable, that workflow simply sets the environment variable to the value of the environment variable. One would need to delete the environment variable to allow for updating the value. Alternatively, we could add a CLI command for reporting what is in the configuration similar to `aws configure get`.

Compared to the alternative solution, the solution proposed here has the benefit that it would remove potentially confusing functionality (multiple sources of auth for the CLI), it would allow for ease and clarity in setting the environment variable using the CLI authentication functionality, and would be aligned with the authentication scheme of gcloud and maxar, making it easier for new users to orient. It has the drawback that, when the environment variable is set, the CLI and the SDK will potentially have two different auth values.

**Additional context**

The three major SDKs in the Planet SDK sphere handle authentication in the following ways:
* Gcloud [CLI](https://cloud.google.com/sdk/docs/authorizing) does not use environment variables, only configuration file. [SDK](https://cloud.google.com/docs/authentication/application-default-credentials) uses environment variables before looking at the (different from CLI) configuration file.
* Maxar [CLI](https://ard.maxar.com/docs/sdk/cli/api/) does not use environment variables, only configuration file. [SDK](https://ard.maxar.com/docs/sdk/sdk/authentication/) uses environment variables before looking at the configuration file.
* AWS [CLI](https://docs.aws.amazon.com/cli/latest/userguide/cli-configure-files.html) uses environment variables before looking at the configuration file and boto3 ([SDK](https://boto3.amazonaws.com/v1/documentation/api/latest/guide/configuration.html)) does the same. When users run `[aws configure get](https://docs.aws.amazon.com/cli/latest/reference/configure/get.html)`, they always get the value in the configuration file, which does not consider other configurations e.g. environment variable.

jreiberkyle:
**Related Issue(s):**

Closes #808 and #699 and #396

**Proposed Changes:**

1. CLI now obtains authorization from secret file exclusively.
2. Add `planet auth store` to allow storing authorization in the secret file. This command raises a confirmation prompt.
3. When `planet auth store` or `planet auth init` are used and the authorization environment variable exists, a warning message is printed with instructions for how to update the environment variable, if desired
4. `auth.Auth.write()` renamed to `auth.Auth.store()` to match CLI command

Not intended for changelog:

1. `CliSession()` now manages authorization information in addition to headers instead of managing authorization information in Click context. This has the benefit of only obtaining authorization when a `CliSession()` is created so not every time something like `planet data --help` is called.
2. Add tests for `CLISession()` to ensure auth information is obtained from the file and that expected headers are sent.
3. Update docs on CLI and SDK authorization priorities. I still think that CLI authorization needs its own section, similar to SDK, so this information can be looked up quickly.

**Diff of User Interface**

Scenario 1: Authorization environment variable exists with value `ENVAPIKEY` and config file exists with value `CONFIGAPIKEY`

Old behavior:

```console
$ planet auth value
ENVAPIKEY
```

New behavior:

```console
$ planet auth value
CONFIGAPIKEY
```

Scenario 2: Authorization environment variable exists. It is desired to update the config file

Old behavior:

```console
$ planet auth init
Email: email
Password: 
Initialized
```

New behavior:

```console
$ planet auth init
Email: email
Password: 
Initialized
Warning - Environment variable PL_API_KEY already exists. To update, with the new value, use the following:
export PL_API_KEY=$(planet auth value)
```

Scenario 3: It is desired to update the config file authorization from the environment variable.

Old behavior:

```python
from planet.auth import Auth

Auth.from_env().write()
```

New behavior:

```python
from planet.auth import Auth

Auth.from_env().store()
```

**PR Checklist:**

### DIFF
--- a/docs/get-started/quick-start-guide.md
+++ b/docs/get-started/quick-start-guide.md
@@ -83,8 +83,8 @@ And you can see that the value was stored successfully as an environment variabl
 echo $PL_API_KEY
 ```
 
-!!!note "The API Key environment variable is always used first in the Planet SDK"
-    If you do create a `PL_API_KEY` environment variable, the SDK will use this value. `PL_API_KEY` overrides the value that was retrieved using your Planet login with a call to `planet auth init`. The `planet auth value` call currently does not reflect that `PL_API_KEY` overrides the `auth init` value (this should be fixed in 2.0-beta.1 with [issue 643](https://github.com/planetlabs/planet-client-python/issues/643))
+!!!note "The API Key environment variable is ignored by the CLI but used by the Python library"
+    If you do create a `PL_API_KEY` environment variable, the CLI will be unaffected but the Planet library will use this as the source for authorization instead of the value stored in `planet auth init`.
 
 ## Step 5: Search for Planet Imagery
 
--- a/docs/python/sdk-guide.md
+++ b/docs/python/sdk-guide.md
@@ -46,15 +46,18 @@ asyncio.run(main())
 
 ## Authenticate with Planet services
 
-There are two steps to managing authentication information, obtaining the
-authentication information from Planet and then managing it for local retrieval
-for authentication purposes.
+A `Session` requires authentication to communicate with Planet services. This
+authentication information is retrieved when a `Session` is created. By default,
+a `Session` obtains authorization from the following sources with order
+indicating priority:
 
-The recommended method for obtaining authentication information is through
-logging in, using `Auth.from_login()` (note: using something like the `getpass`
-module is recommended to ensure your password remains secure). Alternatively,
-the api key can be obtained directly from the Planet account site and then used
-in `Auth.from_key()`.
+1. The environment variable `PL_API_KEY`
+2. The secret file
+
+The SDK provides the `auth.Auth` class for managing authentication information.
+This module can be used to obtain authentication information from username
+and password with `Auth.from_login()`. Additionally, it can be created with
+the API key obtained directly from the Planet account site with `Auth.from_key(<API_KEY>)`.
 
 Once the authentication information is obtained, the most convenient way of
 managing it for local use is to write it to a secret file using `Auth.write()`.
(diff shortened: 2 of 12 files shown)


jreiberkyle:
@cholmes can you review the UI changes?
@JuLeeAtPlanet can you review the doc changes?
@kevinlacaille can you review the code implementation?

ty!

jreiberkyle:
actually, changing review structure so it's just on @kevinlacaille for the implementation. And @cholmes and @JuLeeAtPlanet FYI on the docs and UI

PR #816, opened 2022-12-03
fix subscriptions list_results paging bug

### LINKED ISSUE #777: `planet subscriptions results` not returning more than 20 results
**Expected behavior**
`planet subscriptions results <sub-id>` to return 100 results as specified as the default in the documentation. Or to return 
the proper number of results if `--limit` is greater than 20 (it works properly if the limit is less than 20)

Only 20 results are returned, even if the `--limit` is set to be greater than 20. 

It seems like the cli client is not actually paging through results - it just does the first 20 and stops. I verified that the http endpoint has more results. It has 20 on the first page, but it has a 'next' link that has 20 per page.

**Workaround**
Going to the http end point directly and following the 'next' links?

**Minimum, Complete, Viable Code Sample**

```console
> planet subscriptions results <sub-id> --limit 21
```

Should result in 21 calls returned, only 20 are.

**Environment Information**
* Operation System Information: macOS-12.2.1-arm64-arm-64bit
* Python version 3.9.15: (main, Oct 11 2022, 21:39:54)  [Clang 14.0.0 (clang-1400.0.29.102)]
* Planet package version: planet, version 2.0a7dev

**Installation Method**
* pip

jreiberkyle:
**Related Issue(s):**

Closes #777 

**Proposed Changes:**

1. Fixes bug where list_results would only return first page

Not intended for changelog:

1. 

**Diff of User Interface**

Old behavior:

```console
planet-client-python$> planet subscriptions results abebfd94-4845-46a1-a96e-a911ca9dbf9b | wc
      20     400   10643
```

New behavior:

```console
planet-client-python$> planet subscriptions results abebfd94-4845-46a1-a96e-a911ca9dbf9b | wc
     100    2000   53146
```

**PR Checklist:**

### DIFF
--- a/planet/clients/subscriptions.py
+++ b/planet/clients/subscriptions.py
@@ -221,7 +221,6 @@ class SubscriptionsClient:
 
         class _ResultsPager(Paged):
             """Navigates pages of messages about subscription results."""
-            NEXT_KEY = '_next'
             ITEMS_KEY = 'results'
 
         params = {'status': [val for val in status or {}]}
--- a/tests/integration/test_subscriptions_api.py
+++ b/tests/integration/test_subscriptions_api.py
@@ -122,7 +122,7 @@ def result_pages(status=None, size=40):
         pm = datetime.now().isoformat()
         if len(results) == size:
             url = f'https://api.planet.com/subscriptions/v1/42/results?pm={pm}'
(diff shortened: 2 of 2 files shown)


jreiberkyle:
@cholmes FYI

PR #817, opened 2022-12-05
Allow httpx>=0.23

sgillies:
So that users can use httpx 0.23.1 etc.

Packages like planet can't fix the upper end of requirement specs. If they do, they can be uninstallable from time to time.


sgillies:
0.23.1 has breaking changes https://github.com/encode/httpx/blob/master/CHANGELOG.md#0231. This will require actual work.

PR #834, opened 2023-01-19
Have calculate_wait stop just shy of 1sec offset to optimize distribution

### LINKED ISSUE #723: An http test, test__calculate_wait, randomly fails
The test `test__calculate_wait()`, which compares a randomly generated number with a list of hard-coded numbers, randomly fails.

Here's the traceback:

```
tests/unit/test_http.py:277: AssertionError

    def test__calculate_wait():
        max_retry_backoff = 20
        wait_times = [
            http.Session._calculate_wait(i + 1, max_retry_backoff)
            for i in range(5)
        ]
    
        # (min, max): 2**n to 2**n + 1, last entry hit threshold
        expected_times = [2, 4, 8, 16, 20]
    
        for wait, expected in zip(wait_times, expected_times):
            # this doesn't really test the randomness but does test exponential
            # and threshold
>           assert math.floor(wait) == expected
E           assert 17 == 16
E            +  where 17 = <built-in function floor>(17.0)
E            +    where <built-in function floor> = math.floor
```

"random is, unpredictably, often trouble"
-@jreiberkyle

jreiberkyle:
The step size between calls is 1s and the random offset is called to avoid waves, so have it stop just shy of 1s. This addresses the random failing test in test_http caused when the random offset was 1s.

**Related Issue(s):**

Closes #723

**Proposed Changes:**

1. 

Not intended for changelog:

1. Have calculate_wait stop just shy of 1sec offset

**Diff of User Interface**

Old behavior:

running the test 5000 times resulted in 37 failed tests

New behavior:

running the test 5000 times resulted in 0 failed tests

**PR Checklist:**

### DIFF
--- a/planet/http.py
+++ b/planet/http.py
@@ -335,7 +335,7 @@ class Session(BaseSession):
         """Calculates retry wait
 
         Base wait period is calculated as a exponential based on the number of
-        tries. Then, a random jitter of up to 1s is added to the base wait
+        tries. Then, a random jitter of up to 999ms is added to the base wait
         to avoid waves of requests in the case of multiple requests. Finally,
         the wait is thresholded to the maximum retry backoff.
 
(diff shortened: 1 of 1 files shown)


kevinlacaille (approved):
Nice and easy solution to this issue, thanks for the fix!

jreiberkyle:
LOL it turns out the test was actually telling us something (that we shouldn't be going up to 1000s for optimal distribution) instead of just failing. Good test!

Issue #682, opened 2022-08-30
Planet Cli2 SDK-Data API: Search command is giving error as invalid item type for a valid one


rajashekarmamidala:
**Environment**: planet, version 2.0a2

**Test steps**:
1. Do a search with data api
`planet data search --limit 2 item_types PSScene`

**Expected**: It should display the data for PSScene 

**Actual**: It is displaying error as below

```
Usage: planet data search [OPTIONS] ITEM_TYPES [FILTER]
Try 'planet data search --help' for help.

Error: Invalid value for 'ITEM_TYPES': ['item_types'] should be one of {'REOrthoTile', 'SkySatCollect', 'PSScene4Band', 'PSOrthoTile', 'Sentinel1', 'SkySatScene', 'Sentinel2L1C', 'Landsat8L1G', 'PSScene3Band', 'MYD09GQ', 'MOD09GA', 'PSScene', 'REScene', 'MOD09GQ', 'MYD09GA'}
```

cholmes:
Thanks for the report @rajashekarmamidala! 

So for this one the proper is `planet data search --limit 2 PSScene` - you don't include the `item_types` word in the request.

You still need to include a filter though - if you just enter the above it'll hang, waiting for a filter.

jreiberkyle:
fixed with #840

Issue #598, opened 2022-06-21
Using `planet orders create` by itself feels like it hangs


cholmes:
If I just do `planet orders create` it seems like nothing happens. Maybe it's waiting for stdin? All the other commands seem to do something, like say that it's expecting a file. Maybe there's a good reason it's not doing it, but as a user I'm not sure what I'm supposed to try next, while the others tend to flag what it's expecting next, and point people towards using the command with `--help`

jreiberkyle:
Being new to piping myself, I was also surprised by this behavior. I can say that it is the same behavior one gets if they call `tee` or `grep a` (here, 'a' can be replaced by any pattern) without a pipe in, they just wait for user input.

jreiberkyle:
yeah, the UI of less does seem more friendly. I agree with spending time to see if we can implement that approach in the SDK CLI.

cholmes:
Moving to beta.2, as #835 should solve this.

jreiberkyle:
fixed with #840

PR #840, opened 2023-01-24
Remove stdin from default for CLI commands, require `-` to specify

### LINKED ISSUE #835: Remove `-`/stdin as default for aguments
Having `-` (to specify stdin) as a default for arguments is causing user confusion, as it seems like it 'hangs' when the argument is not specified - #598, #682

Further, there is a desire to make FILTER an optional argument where it is replaced by an empty filter instead of stdin, #600.

Therefore, change the CLI behavior so that stdin is not the default for arguments, and the user must specify it with `-`.

jreiberkyle:
**Related Issue(s):**

Closes #835 #682 #598

**Proposed Changes:**

1. Remove stdin from default for CLI commands, require `-` to specify.

Not intended for changelog:

1. Update subscriptions client to re-use JSON CLI type.

**Diff of User Interface**

Old behavior:
```bash
cat request.json | planet orders create
``` 
would work

```bash
planet orders create
``` 
would appear to hang, as it is waiting for stdin

New behavior:

```bash
planet orders create
``` 
returns the errror `Error: Missing argument 'REQUEST'`

```bash
cat request.json | planet orders create -
``` 
works

**PR Checklist:**

@mkshah605

### DIFF
--- a/tests/integration/test_orders_cli.py
+++ b/tests/integration/test_orders_cli.py
@@ -417,17 +417,8 @@ def test_cli_orders_download_state(invoke, order_description, oid):
     assert 'order state (running) is not a final state.' in result.output
 
 
-# TODO: add tests of "create --pretty" (gh-491).
-@pytest.mark.parametrize(
-    "id_string, expected_ids",
-    [('4500474_2133707_2021-05-20_2419', ['4500474_2133707_2021-05-20_2419']),
-     ('4500474_2133707_2021-05-20_2419,4500474_2133707_2021-05-20_2420',
-      ['4500474_2133707_2021-05-20_2419', '4500474_2133707_2021-05-20_2420'])])
 @respx.mock
-def test_cli_orders_create_basic_success(expected_ids,
-                                         id_string,
-                                         invoke,
-                                         order_description):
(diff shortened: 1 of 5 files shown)


kevinlacaille (approved):
Great call pulling from the `types` module, though I do not like how we have a module with the same name as the [`types` module from Python's standard library](https://docs.python.org/3/library/types.html).

tests/integration/test_orders_cli.py
```diff
@@ -417,25 +417,16 @@ def test_cli_orders_download_state(invoke, order_description, oid):
     assert 'order state (running) is not a final state.' in result.output
 
 
-# TODO: add tests of "create --pretty" (gh-491).
```
kevinlacaille, line 420: What was the purpose of changing this test? Seems not relevant to the tickets.

jreiberkyle: I was considering changing this test to test the inputs (json string, etc) and while I was in there I noticed this parameterization isn't used (id_string is never used and there is no logic to handle differences in expected_ids). I realized that the various inputs are all tested in the types module (for JSON type) so didn't need to actually add to this test but in the interest of cleaning as we go, I left the change to the test.

Issue #704, opened 2022-10-03
Possible to create a filter with conflicting rules


bosth:
`planet data filter --permission False --string-in quality_category test` produces the following filter:

`{"type": "AndFilter", "config": [{"type": "StringInFilter", "field_name": "quality_category", "config": ["test"]}, {"type": "StringInFilter", "field_name": "quality_category", "config": ["standard"]}]}`

I think it would be nice if we weren't setting filters by default (i.e. quality of `standard`) anyway, but this issue is another reason to reconsider.

jreiberkyle:
We went with not setting filters by default, see #842. That resolves this issue.

Issue #637, opened 2022-07-17
user-specified quality category should override default


cholmes:
Just tried to specify a category filter and got a filter that had both in it:

```
(planet-v2-cli) cholmes@x9t37m3hgt requests % planet data filter --string-in quality_category standard,test | jq
{
  "type": "AndFilter",
  "config": [
    {
      "type": "StringInFilter",
      "field_name": "quality_category",
      "config": [
        "standard",
        "test"
      ]
    },
    {
      "type": "PermissionFilter",
      "config": [
        "assets:download"
      ]
    },
    {
      "type": "StringInFilter",
      "field_name": "quality_category",
      "config": [
        "standard"
      ]
    }
  ]
}
```

jreiberkyle:
Oops! Good catch

PR #842, opened 2023-01-25
Remove filter requirement for data quick search, planet data filter cli command outputs empty filter by default

### LINKED ISSUE #600: Don't require a filter for a quick search.

jreiberkyle:
**Related Issue(s):**

Closes #600 #704 #637

**Proposed Changes:**

1. Data client CLI and API no longer require filter to be specified, and will use an empty filter (no filtering) if it is not. Filter moved from a positional argument for the CLI and API to an optional key word argument, `--filter` for the CLI and `search_filter` for the API.
2. Data filter CLI command now outputs an empty filter by default, it no longer includes permissions and standard quality filters by default, these are easily added with the `--permissions` and `--std-quality` flags

Not intended for changelog:

1. Update CLI docs to use new interface. Change wording some to discuss permissions and standard quality filters not being defaults but being easy to add. @JuLeeAtPlanet FYI

**Diff of User Interface**

#### Planet data filter
Old behavior:

`planet data filter` outputs a permissions and standard quality filter by default
```
> planet data filter
{"type": "AndFilter", "config": [{"type": "PermissionFilter", "config": ["assets:download"]}, {"type": "StringInFilter", "field_name": "quality_category", "config": ["standard"]}]}
```

to get an empty filter use
```
> planet data filter 
{"type": "AndFilter", "config": []}
```

New behavior:
`planet data filter` outputs an empty filter by default
```
> planet data filter
{"type": "AndFilter", "config": []}
```
to get a permissions/std-quality filter:
```
> planet data filter --permission --std-quality
{"type": "AndFilter", "config": [{"type": "PermissionFilter", "config": ["assets:download"]}, {"type": "StringInFilter", "field_name": "quality_category", "config": ["standard"]}]}
```

#### Planet quick search, CLI
Old behavior:

`planet data search PSScene filter.json` applies filter to search, returns at most 100 results

`planet data search PSScene` errors out due to FILTER being a required argument

New behavior:

`planet data search PSScene -filter filter.json` applies filter to search

`planet data search PSScene` returns 100 (default value of limit) results

#### Planet quick search, API
Old behavior:
```python
data_client.search(['PSScene'], filter)
```
applies filter to search

```python
data_client.search(['PSScene'])
```
Errors out due to search_filter being a required argument

New behavior:

```python
data_client.search(['PSScene'], search_filter=filter)
```
applies filter to search

```python
data_client.search(['PSScene'])
```
Returns 100 (default value of limit) results

**PR Checklist:**

@cholmes @JuLeeAtPlanet

### DIFF
--- a/docs/cli/cli-data.md
+++ b/docs/cli/cli-data.md
@@ -17,10 +17,10 @@ At this point you should have completed [Step 5](../get-started/quick-start-guid
 of the quick start guide, and run your first full data search command:
 
 ```
-planet data search PSScene filter.json > recent-psscene.json
+planet data search PSScene --filter filter.json > recent-psscene.json
 ```
 
-This saves the latest 100 scenes in a file, that you can open and look at.
+This saves the descriptions of the latest 100 standard-quality scenes you have permissions to download in a file, that you can open and look at.
 
 ### Pretty printing
 
@@ -28,7 +28,7 @@ You will likely notice that this file is quite wide, with one very long line for
 item returned. You can make for a more readable file by using the `--pretty` flag:
 
 ```
-planet data search --pretty PSScene filter.json > recent-psscene.json
+planet data search --pretty PSScene --filter filter.json > recent-psscene.json
 ```
 
 The `--pretty` flag is built into most of the CLI calls. But you can also achieve the
@@ -38,7 +38,7 @@ piping any JSON output through it prints it in a more readable form. So the foll
 command will do the same thing as the previous one:
 
 ```
-planet data search PSScene filter.json | jq > recent-psscene.json
+planet data search PSScene --filter filter.json | jq > recent-psscene.json
 ```
 
 You can read a bit [more about jq]((cli-intro.md#jq) in the CLI intro.
@@ -49,14 +49,14 @@ You also don't have to save the output to a file. If you don't redirect it into 
 it will just print out on the console.
 
 ```
-planet data search PSScene filter.json 
+planet data search PSScene --filter filter.json
 ```
 
 If you enter this command you'll see the output stream by. Here you can use jq again, and
 it'll often give you nice syntax highlighting in addition to formatting.
 
 ```
-planet data search PSScene filter.json | jq
+planet data search PSScene --filter filter.json | jq
 ```
 
 ### Create filter and search in one call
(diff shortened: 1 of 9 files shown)


jreiberkyle:
@julee could you look at content changes?
@cholmes could you look at usage pattern?
@kevinlacaille could you do a code review?
TY!

cholmes (approved):
Looks great, works great in my limited testing.

I think the only thing in here that will be subtly confusing, especially in evolution from the current thing, is that:

`planet data filter --number-in cloud_percent 4 | planet data search PSScene` won't actually get the filter applied. You need to remember to add `--filter -`. If you've got any ideas on how to nudge people towards that it could be good. Like can we detect if they're attempting to pipe something and warn them that it won't be applied?

jreiberkyle:
@cholmes hmm... yeah, that's a hard one to protect against. There is the option of prompting the user for confirmation (overridden with an `-y` option) when a search is run without a filter. Requires an opt-in for searching without a filter.

cholmes:
Could it just raise a little warning when searching without a filter? 
Like just says 'executing search with no filter'. I don't think we need to keep that warning around forever, just for a few releases to help people adopt new behavior. 

Even better would be if we could detect if there's anything being piped in. Like if it detects some attempted pipe then it would raise that message (or an even stronger warning, since it wouldn't come up every time)

kevinlacaille (approved):
Everything looks good to me! I can confirm that all tests pass on 3.10 and 3.11.

jreiberkyle:
@cholmes I don't know of a way to check for an attempt to pipe in. Or, if there is a way, it would take a lot of work. I'm also not sure of how to output a warning that wouldn't mess up any further processing of the search results - like if we pipe it to jq then jq will probably croak on the warning message before it can process the results that are the actual outputs.

cholmes:
>  I'm also not sure of how to output a warning that wouldn't mess up any further processing of the search results - like if we pipe it to jq then jq will probably croak on the warning message before it can process the results that are the actual outputs.

Ah, I had been thinking that a warning maybe wouldn't go to stdout in the same way, but you're probably right. Ok, sounds like no easy solution - seems like we should just ship it and be sure to document it all well. Maybe a note / warning somewhere in the docs, and in the release notes, to notify people to watch out for piping without the '-'. New users shouldn't have any trouble with it, and we only have a small number of users today, so it should be fine.

PR #850, opened 2023-02-08
Implement CLI commands planet data search-delete

### LINKED ISSUE #500: Implement CLI command: planet data search-delete
As specified in https://github.com/planetlabs/planet-client-python/blob/main/design-docs/CLI-Data.md#search-delete

kevinlacaille:
**Related Issue(s):**

Closes #500 

**Proposed Changes:**

1. Implement CLI commands planet data search-delete

Not intended for changelog:

1. 

**Diff of User Interface**

Old behavior:

N/A

New behavior:
```console
❯ planet data search-delete 14aab90bb19749f483bfa929938a8382
# try and delete the same search ID again
❯ planet data search-delete 14aab90bb19749f483bfa929938a8382 
Error: {"general": [{"message": "The requested search id does not exist"}], "field": {}}
```

**PR Checklist:**

### DIFF
--- a/tests/integration/test_data_cli.py
+++ b/tests/integration/test_data_cli.py
@@ -701,9 +701,18 @@ def test_search_get_id_not_found(invoke, search_id):
     assert 'Error: {"message": "Error message"}\n' == result.output
 
 
+@respx.mock
+def test_search_delete(invoke, search_id, search_result):
+    delete_url = f'{TEST_SEARCHES_URL}/{search_id}'
+    mock_resp = httpx.Response(HTTPStatus.NO_CONTENT, json=search_result)
+    respx.delete(delete_url).return_value = mock_resp
+
+    result = invoke(['search-delete', search_id])
+
+    assert not result.exception
+
 # TODO: basic test for "planet data search-create".
 # TODO: basic test for "planet data search-update".
-# TODO: basic test for "planet data search-delete".
 # TODO: basic test for "planet data search-get".
 # TODO: basic test for "planet data search-list".
 # TODO: basic test for "planet data search-run".
--- a/tests/integration/test_data_cli.py
+++ b/tests/integration/test_data_cli.py
@@ -703,7 +703,7 @@ def test_search_get_id_not_found(invoke, search_id):
 
 
 @respx.mock
-def test_search_delete(invoke, search_id, search_result):
+def test_search_delete_success(invoke, search_id, search_result):
     delete_url = f'{TEST_SEARCHES_URL}/{search_id}'
     mock_resp = httpx.Response(HTTPStatus.NO_CONTENT, json=search_result)
     respx.delete(delete_url).return_value = mock_resp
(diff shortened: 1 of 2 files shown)


tests/integration/test_data_cli.py
```diff
@@ -701,9 +702,31 @@ def test_search_get_id_not_found(invoke, search_id):
     assert 'Error: {"message": "Error message"}\n' == result.output
 
 
+@respx.mock
+def test_search_delete(invoke, search_id, search_result):
```
jreiberkyle, line 706: ok this is just a style thing, but I love to use pytest's `-k` option to select a test to run. That option just matches the beginning of a test name so `-k test_search_delete` would run this test as well as the next test. I tend to work around that by adding a unique identifier to the end of the name - like `test_search_delete_success`. Bonus of a little more descriptive name!

tests/integration/test_data_cli.py
```diff
@@ -701,9 +702,31 @@ def test_search_get_id_not_found(invoke, search_id):
...
+    delete_url = f'{TEST_SEARCHES_URL}/{search_id}'
+    mock_resp = httpx.Response(HTTPStatus.NO_CONTENT, json=search_result)
```
jreiberkyle, line 708: Ohh, cool catch on the NO_CONTENT status code. This is the code the API returns?

tests/integration/test_data_cli.py
```diff
@@ -701,9 +702,31 @@ def test_search_get_id_not_found(invoke, search_id):
...
+@respx.mock
+def test_search_delete_invalid_id(invoke, search_id, search_result):
+    delete_url = f'{TEST_SEARCHES_URL}/{search_id}'
+    mock_resp = httpx.Response(404, json=search_result)
+    respx.delete(delete_url).return_value = mock_resp
+
+    result = invoke(['search-delete', search_id])
+
+    assert result.exception
+    assert result.exit_code == 1
```
jreiberkyle, line 725: Hrm... Ideally, with an invalid parameter (search_id, here), we'd return code of 2 ([ref](https://github.com/planetlabs/planet-client-python/blob/main/DESIGN.md#cli-return-codes)). But we aren't pre-validating the search-id before we send it to the API. Maybe just change the name of the test to say the search_id is nonexistant instead of invalid? I *think* the existence is really what we're checking here, no?

tests/integration/test_data_cli.py
jreiberkyle: I just checked the api and it looks like it returns a 204 - is that what NO_CONTENT is? 
https://developers.planet.com/docs/apis/data/reference/#tag/Item-Search/operation/DeleteSearch

jreiberkyle (approved):
Looks awesome. I mostly have some namechange suggestions. Great job!

tests/integration/test_data_cli.py
kevinlacaille: You got it!

tests/integration/test_data_cli.py
kevinlacaille: 204 is indeed NO_CONTENT! https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/204

tests/integration/test_data_cli.py
kevinlacaille: I think you're right

PR #858, opened 2023-02-26
Add client method to Session class

### LINKED ISSUE #857: Add a method to get a client from a Session instance
As in AWS boto3: https://boto3.amazonaws.com/v1/documentation/api/latest/guide/session.html#custom-session. For planet, this would look like

```python
session = planet.Session(...)
client = session.client("data")
```

Currently, we need to do the following.

```
session = planet.Session(...)
client = DataClient(session)
```

There's nothing wrong with this kind of dependency injection, and we're going to keep this usage. But what I'm proposing is more handy and is easy to remember.

sgillies:
It gets a client by the name of the client module. `orders` -> `planet.clients.orders.OrdersClient`.

The intent is to simplify the API for casual users. The API centers the Session class. It feels natural to let Session be a factory for clients. And it goes well with code autocompletion. Type `session.` in VS Code and we get a prompt for `.client()`.

Screenshot from 2023-02-27 10-28-38

Type `session.client(` and we get a prompt to call it with a name like data, orders, or subscriptions. Actually, I think the docs could be slightly improved to help even more.

Screenshot from 2023-02-27 10-29-30

The new method is intended to be a convenience for users and documentation writers. The Session and three client classes are not otherwise changed. They don't need to be and much testing implementation hinges on the existing relationship between client classes and Session. The pattern used by `OrdersClient(session)` is a form of dependency injection; it really does make writing good tests easier.

The CLI commands are left as they are. 

Under docs/, all examples of `client = SomethingClient(session)` have been changed to `client = session.client('something')`.

Resolves #857.

### DIFF
--- a/docs/get-started/upgrading.md
+++ b/docs/get-started/upgrading.md
@@ -24,7 +24,7 @@ The best way of doing this is wrapping any code that invokes a client class in a
 
 ```python
 async with Session() as session:
-    client = OrdersClient(session)
+    client = session.client('orders')
     result = await client.create_order(order)
 # Process result
 ```
@@ -40,12 +40,12 @@ In V2, all `*Client` methods (for example, `DataClient().search`, `OrderClient()
 ```python
 import asyncio
 from datetime import datetime
-from planet import Session, DataClient
+from planet import Session
 from planet import data_filter as filters
  
 async def do_search():
     async with Session() as session:
-        client = DataClient(session)
+        client = session.client('data')
         date_filter = filters.date_range_filter('acquired', gte=datetime.fromisoformat("2022-11-18"), lte=datetime.fromisoformat("2022-11-21"))
         cloud_filter = filters.range_filter('cloud_cover', lte=0.1)
         download_filter = filters.permission_filter()
@@ -74,11 +74,11 @@ Is now
 
 ```python
 async with Session() as session:
-    items = [i async for i in planet.DataClient(session).search(["PSScene"], all_filters)]
+    items = [i async for i in session.client('data').search(["PSScene"], all_filters)]
 ```
 
 ## Orders API
 
-The Orders API capabilities in V1 were quite primitive, but those that did exist have been retained in much the same form; `ClientV1().create_order` becomes `OrderClient(session).create_order`. (As with the `DataClient`, you must also use `async` and `Session` with `OrderClient`.)
+The Orders API capabilities in V1 were quite primitive, but those that did exist have been retained in much the same form; `ClientV1().create_order` becomes `OrdersClient(session).create_order`. (As with the `DataClient`, you must also use `async` and `Session` with `OrdersClient`.)
 
 Additionally, there is now also an order builder in `planet.order_request`, similar to the preexisting search filter builder. For more details on this, refer to the [Creating an Order](../../python/sdk-guide/#creating-an-order).
--- a/planet/http.py
+++ b/planet/http.py
@@ -431,10 +431,10 @@ class Session(BaseSession):
 
         """
         # To avoid circular dependency.
-        from planet.clients import client_directory
+        from planet.clients import _client_directory
 
         try:
-            return client_directory[name](self, base_url=base_url)
+            return _client_directory[name](self, base_url=base_url)
         except KeyError:
             raise exceptions.ClientError("No such client.")
 
--- a/planet/http.py
+++ b/planet/http.py
@@ -24,6 +24,7 @@ import time
 from typing import AsyncGenerator, Optional
 
 import httpx
+from typing_extensions import Literal
 
 from .auth import Auth, AuthType
 from . import exceptions, models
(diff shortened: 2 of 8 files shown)


docs/get-started/upgrading.md
```diff
@@ -24,7 +24,7 @@ The best way of doing this is wrapping any code that invokes a client class in a
 
 ```python
 async with Session() as session:
```
sgillies, line 26: This is what I mean by "centers the Session class". Creating one is the first thing we do and it's the manager for the context of a block of code.

planet/http.py
```diff
@@ -413,6 +414,34 @@ async def stream(
...
+            A client instance.
+
+        Raises:
+            ClientError when no such client can be had.
+
+        """
+        # To avoid circular dependency.
+        from planet.clients.data import DataClient
+        from planet.clients.orders import OrdersClient
+        from planet.clients.subscriptions import SubscriptionsClient
```
sgillies, line 434: Session and clients aren't fully decoupled because the client modules import `http`. We can fix that in the future. For now, we can tolerate some imports here. They don't have a significant impact on performance. By the time this method is called, those modules will already have been loaded.

planet/http.py
```diff
@@ -413,6 +414,34 @@ async def stream(
...
+
+        try:
+            return {
+                'data': DataClient,
+                'orders': OrdersClient,
+                'subscriptions': SubscriptionsClient
```
sgillies, line 440: For now, I'm statically configuring the mapping between names and clients here. Getting too fancy would be a mistake. We have no plans to have multiple clients per API or support user-supplied client plugins.

tests/unit/test_session.py
```diff
@@ -0,0 +1,23 @@
...
+    session = Session()
+    client = session.client(client_name)
+    assert isinstance(client, client_class)
+
+
+def test_session_get_client_error():
+    """Get an exception when no such client exists."""
+    session = Session()
+    with pytest.raises(ClientError):
+        _ = session.client('bogus')
```
sgillies, line 23: Test coverage of the new code is 100%.

planet/http.py
```diff
@@ -414,12 +415,13 @@ async def stream(
...
-    def client(self, name: str, base_url: Optional[str] = None) -> object:
+    def client(self,
+               name: Literal['data', 'orders', 'subscriptions'],
+               base_url: Optional[str] = None) -> object:
         """Get a client by its module name.
 
         Parameters:
-            name: the name of the client module: data, orders, or
-                subscriptions.
+            name: one of 'data', 'orders', or 'subscriptions'.
```
sgillies, line 424: This is pretty nice, I think. A user is told that they should pass 'data', 'orders', or 'subscription' as the argument.

Screenshot from 2023-02-27 10-44-36

setup.py
```diff
@@ -30,6 +30,7 @@
     'jsonschema',
     'pyjwt>=2.1',
     'tqdm>=4.56',
+    'typing-extensions',
```
sgillies, line 33: This module provides a backport of `Literal` for Python 3.7.

planet/http.py
JuLeeAtPlanet: Do you want to file a ticket for fixing that in the future?

JuLeeAtPlanet (approved):
One comment for you, but good to go!

planet/http.py
sgillies: No, not now. I'm not sure it's important. I changed the text above to reflect that.

planet/http.py
sgillies: Looking at my PR as if I was a reviewer, it occurs to me that this static mapping could be moved to `planet/client/__init__.py`. That would let us reduce those 3 imports to 1.

planet/http.py
sgillies: Note: the client modules import http in order to use http.Session in their type hints.

planet/http.py
```diff
@@ -431,10 +431,10 @@ def client(self,
 
         """
         # To avoid circular dependency.
-        from planet.clients import client_directory
+        from planet.clients import _client_directory
 
         try:
-            return client_directory[name](self, base_url=base_url)
+            return _client_directory[name](self, base_url=base_url)
```
sgillies, line 437: This way the Session class only needs to know that it can get a client with a name, it's not responsible for the directory. Looser coupling.

sgillies:
Restoring the branch, we might be reverting this.

planet/http.py
JuLeeAtPlanet: I saw that!

planet/http.py
jreiberkyle: This typing hint IS nice! This would be great to use this for all of the other arguments that have a set number of 'valid' entries, we already have them defined either at the top of the client module or in the specs module.

sgillies:
Reverted pending more discussion.

jreiberkyle:
This PR looks great and is ready to go. I really like the use of `Literals` to allow specifying valid entries. That is nice!

PR #859, opened 2023-02-27
Implement subscription request helper cli commands

### LINKED ISSUE #614: Helper methods for subscriptions
For the MVP of subscriptions we just have users directly form the entire json request, so we could ship something and iterate.

Now that it's out we should consider what we want to do to help users. To have one or more options like `planet data filter` or `planet order request`.

A user should be able to construct a decent subscription without having to get into JSON. When I have some time I'll map out some proposals, but just opening this to make sure we track and prioritize. 

#613 #601 are related - ideas on generating some of the JSON parts.

jreiberkyle:
**Related Issue(s):**

Closes #614 

**Proposed Changes:**

1. Add `planet subscriptions request` and `planet subscriptions request-catalog` cli commands to help with creating subscription requests

Not intended for changelog:

1. 

**Diff of User Interface**

Old behavior:
N/A

New behavior:

```bash
planet subscriptions request-catalog \
        --item-types PSScene \
        --asset-types ortho_analytic_8b_sr,ortho_udm2 \
        --geometry aoi.geojson \
        --start-time 2022-01-01
```

```
planet subscriptions request \
    --source $(planet subscriptions request-catalog \
        --item-types PSScene \
        --asset-types ortho_analytic_8b_sr,ortho_udm2 \
        --geometry aoi.geojson \
        --start-time 2022-01-01) \
    --delivery delivery.json | planet subscriptions create -
```

aoi.geojson
```
{
  "type": "FeatureCollection",
  "features": [
    {
      "type": "Feature",
      "properties": {},
      "geometry": {
        "type": "Polygon",
        "coordinates": [
          [
            [
              7.05322265625,
              46.81509864599243
            ],
            [
              7.580566406250001,
              46.81509864599243
            ],
            [
              7.580566406250001,
              47.17477833929903
            ],
            [
              7.05322265625,
              47.17477833929903
            ],
            [
              7.05322265625,
              46.81509864599243
            ]
          ]
        ]
      }
    }
  ]
}
```

delivery.json
```
{
  "type": "amazon_s3",
  "parameters": {
    "aws_access_key_id": "aws_access_key_id",
    "aws_secret_access_key": "aws_secret_access_key",
    "bucket": "bucket",
    "aws_region": "aws_region"
  }
}
```

**PR Checklist:**

### DIFF
--- a/design-docs/CLI-Subscriptions.md
+++ b/design-docs/CLI-Subscriptions.md
@@ -259,7 +259,7 @@ Generate a subscriptions request source JSON for another product.
 Options:
   --type                        Type.
   --id                          Id.
-  --geom JSON                      geometry of the area of interest of the subscription that will be used to determine matches.
+  --geometry JSON                      geometry of the area of interest of the subscription that will be used to determine matches.
                                   Can be a json string, filename, or '-' for stdin.
   --start-time DATETIME           Start date and time to begin subscription.
   --end-time DATETIME             Date and time to end the subscription.
--- a/design-docs/CLI-Subscriptions.md
+++ b/design-docs/CLI-Subscriptions.md
@@ -207,7 +207,10 @@ Options:
 ### Usage Examples
 
 ```
-planet subscription request --source source.json --clip geom.json --delivery delivery.json | planet subscriptions create -
+planet subscription request \
+    --name test \
+    --source source.json \
+    --delivery delivery.json | planet subscriptions create -
 ```
 
 ## Request-catalog
@@ -236,13 +239,12 @@ Options:
 ### Usage Examples
 
 ```
-planet subscriptions request \
-    --source $(planet subscriptions request-catalog \
+planet subscriptions request-catalog \
         --item-types PSScene \
         --asset-types ortho_analytic_8b_sr,ortho_udm2 \
         --geometry aoi.geojson \
         --start-time 2022-01-01) \
-    --delivery delivery.json | planet subscriptions create -
+    --delivery delivery.json > source.json
 ```
 
 ## Request-other
--- a/design-docs/CLI-Subscriptions.md
+++ b/design-docs/CLI-Subscriptions.md
@@ -184,7 +184,7 @@ planet subscriptions request [OPTIONS]
 
 Generate a subscriptions request.
 
-This command provides support for building the subscription request JSON used to create or 
+This command provides support for building the subscription request JSON used to create or
 update a subscription. It outputs the subscription request.
 
 Options:
@@ -222,13 +222,13 @@ Generate a subscriptions request source JSON for a catalog.
 Options:
   --asset-types TEXT              One or more comma-separated asset types. Required.
   --item-types TEXT               One or more comma-separated item-types. Required.
-  --geom JSON                 geometry of the area of interest of the subscription that will be used to determine matches. 
+  --geometry JSON                 geometry of the area of interest of the subscription that will be used to determine matches.
                                   Can be a json string, filename, or '-' for stdin.
   --start-time DATETIME           Start date and time to begin subscription.
   --end-time DATETIME             Date and time to end the subscription.
-  --rrule TEXT                    iCalendar recurrance rule to specify recurrances. 
-  --filter JSON                   A search filter can be specified a json string, 
-                                  filename, or '-' for stdin. 
+  --rrule TEXT                    iCalendar recurrance rule to specify recurrances.
+  --filter JSON                   A search filter can be specified a json string,
+                                  filename, or '-' for stdin.
   --pretty                        Format JSON output.
   --help                          Show this message and exit.
 ```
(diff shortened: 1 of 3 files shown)


design-docs/CLI-Subscriptions.md
```diff
@@ -237,11 +237,11 @@ Options:
 
 ```
 planet subscriptions request \
```
kevinlacaille, line 239: May I suggest having an example that includes all of the options? In this case, from the docstring alone it's unclear to me what an appropriate input would be for `rrule`.

design-docs/CLI-Subscriptions.md
```diff
@@ -222,13 +222,13 @@ Generate a subscriptions request source JSON for a catalog.
 Options:
```
kevinlacaille, line 222: 1. The example above, line 210, does not work. It is missing a `--name` flag. 
2. `--clip` is no longer an option, looks like it got replaced by `--tools`.

A working example would look something like this 
```
planet subscription request --name "some name" --source source.json --tools geom.json --delivery delivery.json | planet subscriptions create -
```

design-docs/CLI-Subscriptions.md
```diff
@@ -237,11 +237,11 @@ Options:
 
 ```
 planet subscriptions request \
-    --source $(planet subscriptions request-catalog
-        --item-type PSScene
-        --asset-types ortho_analytic_8b_sr,ortho_udm2
-        --geom aoi.json
-        --start-time 05/01/2022)
+    --source $(planet subscriptions request-catalog \
```
kevinlacaille, line 240: - The `request` call is missing its `--name` flag.
- I really like this example, but I would recommend splitting this example into 2 parts:
1. A direct call to `request-catalog`, which you wrote in your PR:
```
planet subscriptions request-catalog \
        --item-types PSScene \
        --asset-types ortho_analytic_8b_sr,ortho_udm2 \
        --geometry aoi.geojson \
        --start-time 2022-01-01
```
Maybe even have this write to `source.json`?

2. Either the `request` example you wrote here, showing how you can call the function inline OR a `request` command with `source.json` as an output from the example 1. For instance, either
```planet subscriptions request \
    --name "some name" \
    --source $(planet subscriptions request-catalog \
        --item-types PSScene \
        --asset-types ortho_analytic_8b_sr,ortho_udm2 \
        --geometry aoi.geojson \
        --start-time 2022-01-01) \
    --delivery delivery.json | planet subscriptions create -
```
OR
```
planet subscriptions request \
    --name "some name" \
    --source source.json \
    --delivery delivery.json | planet subscriptions create -
```

tests/integration/test_subscriptions_cli.py
```diff
@@ -250,3 +249,60 @@ def test_subscriptions_results_success(invoke, options, expected_count):
 
     assert result.exit_code == 0  # success.
     assert result.output.count('"id"') == expected_count
+
+
+def test_request_base_success(invoke, geom_geojson):
```
kevinlacaille, line 254: May I also suggest a failure test for `request`?

tests/integration/test_subscriptions_cli.py
```diff
@@ -250,3 +249,60 @@ def test_subscriptions_results_success(invoke, options, expected_count):
...
+        '--name=test',
+        f'--source={source}',
+        f'--delivery={delivery}'
+    ])
+
+    assert source in result.output
+    assert result.exit_code == 0  # success.
+
+
+def test_request_catalog_success(invoke, geom_geojson):
```
kevinlacaille, line 288: May I also suggest a failure test for `request-catalog`?

kevinlacaille (changes requested):
Your CLI commands look great!

1. I found a few issues in the design-docs, but I have recommended some fixes.
2. Your success tests for `request` and `request-catalog` look great, but if you've got the time, I would recommend throwing in some simple failure tests for them.

design-docs/CLI-Subscriptions.md
jreiberkyle: Great suggestion. However, since this is just a design doc, and not actual user-faced documentation, we don't need to get into documenting every option so I won't go into rrule for now.

tests/integration/test_subscriptions_cli.py
jreiberkyle: I didn't include a failure test for two reasons:
1. The success tests fully cover the code. 
2. Looking at the code, there isn't any logic other than calling a few functions that are fully tested. So, I'm not sure what failure I'd test for. Even testing for BadParameter would be basically testing the typing library.

Can you think of a beneficial test to include here?

tests/integration/test_subscriptions_cli.py
jreiberkyle: noted and I'll defer discussion here to the `request` case above

tests/integration/test_subscriptions_cli.py
kevinlacaille: Ah yes, I realize that now. A success test seems to be the only thing that makes sense here 👍

kevinlacaille:
Looks great, thanks for all the efforts, Jen!

kevinlacaille (approved):
Looks great, thanks for all the efforts, Jen!

PR #898, opened 2023-03-24
change search-update name/filter options/parameters to match search UI

### LINKED ISSUE #890: Change `search-update` to be consistent with new search.
I just realized the `search-update` should also be updated for consistency. Elsewhere (#887) I mentioned `planet data stats` could also use an update. I think that one is 'nice to have', but probably not essential. But `search-update` seems like it really should have the same behavior. I'm only trying it out now - I had thought it might just take JSON instead of having arguments, but just realized it has the arguments, and it'd be weird to have them different.

_Originally posted by @cholmes in https://github.com/planetlabs/planet-client-python/issues/883#issuecomment-1478903549_

Note I'd see an ideal `search-update` not requiring all the parameters - if you left them off then it'd assume that you want to keep things the same. So like if you just wanted to update the `name` then you could leave the other values blank. That seems like it'd probably be a complicated implementation, since we'd need to get the JSON and edit the right bits - just putting it here as I think that'd be the ideal experience. But this function is very little used, so not really worth optimizing.

jreiberkyle:
**Related Issue(s):**

Closes #890 

**Proposed Changes:**

1. Breaking change: Bring the UI of `planet data search-update` CLI command inline with `planet data search` and `DataClient.search()` by moving name and filter to required options.

Not intended for changelog:

1. 

**Diff of User Interface**

Old behavior:

```console
planet data filter | planet data search-update d7c24f298c4943078cb7cb7808e989a5 test PSScene -
```

New behavior:

```console
planet data filter | planet data search-update d7c24f298c4943078cb7cb7808e989a5 PSScene --filter - --name test
```

**PR Checklist:**

### DIFF
--- a/planet/cli/data.py
+++ b/planet/cli/data.py
@@ -468,20 +468,28 @@ async def search_delete(ctx, search_id):
 @translate_exceptions
 @coro
 @click.argument('search_id')
-@click.argument('name')
 @click.argument("item_types",
                 type=types.CommaSeparatedString(),
                 callback=check_item_types)
-@click.argument('filter', type=types.JSON())
+@click.option(
+    '--filter',
+    type=types.JSON(),
+    required=True,
+    help="""Filter to apply to search. Can be a json string, filename,
+         or '-' for stdin.""")
+@click.option('--name',
+              type=str,
+              required=True,
+              help='Name of the saved search.')
 @click.option('--daily-email',
               is_flag=True,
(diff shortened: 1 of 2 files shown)


cholmes (approved):

kevinlacaille (approved):
LGTM!

kevinlacaille:
Are we hoping to merge this into `main` or `rc3dev-fixes`?

PR #921, opened 2023-04-13
do not show source in python api reference

jreiberkyle:
Docs no longer include the sourcecode in the python api reference page


JuLeeAtPlanet (approved):
LGTM

PR #923, opened 2023-04-13
Add proper GitHub Actions badge

sgillies:
We're not using Travis CI.


sgillies:
Fixed! See https://github.com/planetlabs/planet-client-python/blob/sgillies-patch-1/README.md

JuLeeAtPlanet (approved):
thank you!

PR #939, opened 2023-04-26
Code snippets for all APIs

### LINKED ISSUE #540: Content Plan Task--Code snippets
For the sake of this document, a code snippet is a block of code in the documentation that provides a code example of how to use the command being documented. A code snippet is not the output of the source code in the SDK. A code snippet for the SDK is not the CLI string.

Provide a code snippet for each call, using real-life terms (i.e., not foo & bar).

Provide a CLI cheat sheet (separate from the SDK) with examples using real-life terms. (See, for example, [The gcloud CLI cheat sheet](https://cloud.google.com/sdk/docs/cheatsheet).)

### LINKED ISSUE #936: Code snippet integration in markdown
https://squidfunk.github.io/mkdocs-material/setup/extensions/python-markdown-extensions/#snippets

### LINKED ISSUE #937: Testing and coverage for code snippets
- Create tests and coverage for the snippets
- Ensure the snippets work properly and that we have snippets for all of the functions

kevinlacaille:
**Related Issue(s):**

Closes #540 #937  

**Proposed Changes:**

1. Code snippets for Orders API, Data API, and Subscriptions API.

Not intended for changelog:

1. 

**Diff of User Interface**

Old behavior:
N/A

New behavior:
N/A

**PR Checklist:**

- [ ] This PR does not break any examples or I have updated them

@jreiberkyle


kevinlacaille:
@cholmes, @sgillies, @jreiberkyle I've pushed this back to a draft PR to work on the markdown ticket (#936) in parallel, as we'll need to add the Snippets extension to add the ability to embed these snippets directly into the markdown.

See: 
- https://squidfunk.github.io/mkdocs-material/setup/extensions/python-markdown-extensions/#snippets 
- https://facelessuser.github.io/pymdown-extensions/extensions/snippets/

examples/snippets/data_api.py
```diff
@@ -0,0 +1,222 @@
...
+async def search(item_types, search_filter, name, sort, limit):
+    '''Code snippet for search.'''
+    async with planet.Session() as sess:
+        client = sess.client('data')
+        async for item in client.search(item_types=item_types,
+                                        search_filter=search_filter,
+                                        name=name,
+                                        sort=sort,
+                                        limit=limit):
+            return item
```
sgillies, line 40: @kevinlacaille What's the intent here? Why wrap lines 33-40 up in a new function?

kevinlacaille: I believe the intent behind the snippets is to show how to use each of the clients' functions, individually. But please correct me if I'm wrong, @sarasafavi @JuLeeAtPlanet. I'm happy to write these in which ever way we intend to use them.

sgillies: I'm 100% sure that if we show to users that they should wrap 3 SDK statements up in a new function with the same name, it will confuse them as much as it helps. Whenever possible, let's try to show Python SDK usage with no wrappers and no helpers, just to keep things clear. Also returning from an async for loop doesn't quite work.

sgillies: Concrete suggestion using snippet sections https://facelessuser.github.io/pymdown-extensions/extensions/snippets/#snippet-sections.

```suggestion
# --8<-- [start:search]
# Search example.
async with planet.Session() as sess:
    client = sess.client('data')
    async for item in client.search(item_types=item_types, search_filter=search_filter, name=name, sort=sort, limit=limit):
        print(item)
# --8<-- [end:search]
```

kevinlacaille: Yes, this is probably a lot more useful. Also, thanks for adding the embedded markdown stuff!

kevinlacaille: I'll change the snippets to this format 👍

sgillies:
@kevinlacaille I've realized a complication for testing against the production APIs: the test runner can only access saved searches and orders that it has made using the same credentials. It looks to me like your tests rely on your own saved searches and orders and that's not going to work.

So it looks like you'll have to have a snippet to make a search, then get that search's id to go back to it. Same for orders.

sgillies:
@kevinlacaille the main source of errors in the new tests is that the clients are pointed at the production API while the credentials stored in GitHub are for staging. See https://github.com/planetlabs/planet-client-python/actions/runs/5525726227/jobs/10079664599#step:5:616.

I think we should reconsider running these snippets against our API. It's not very complicated for the Data API. But it gets more complicated for APIs that are stateful. Is it going to be feasible to keep making and downloading orders every time these tests run? Not to mention subscriptions, which need a cloud destination.

kevinlacaille:
NOTE: Maybe add in CONTRIBUTING.md on how to add to the snippets if a new function is added or if functionality changes.

PR #971, opened 2023-06-30
Add a clip_to_source kwarg to subscript_request.build_request

### LINKED ISSUE #967: Reuse subscription source geometry in clip tool
Specifying a clip geometry will be optional. If one is provided, it will be ignored in favor of the subscription source geometry.

sgillies:
This could be a preview of the behavior of the next API version.

Resolves #967

### DIFF
--- a/planet/subscription_request.py
+++ b/planet/subscription_request.py
@@ -63,12 +63,14 @@ def build_request(name: str,
         tools: Tools to apply to the products. The order of operation
             is determined by the service.
         clip_to_source: whether to clip to the source geometry or not
-            (the default). If True a clip configuration will be added
-            to the list of requested tools unless an existing clip tool
-            exists. NOTE: the next version of the Subscription API will
-            remove the clip tool option and always clip to the source
-            geometry. Thus this is a preview of the next API version's
-            default behavior.
+            (the default). If True a clip configuration will be added to
+            the list of requested tools unless an existing clip tool
+            exists.  NOTE: Not all data layers support clipping, please
+            consult the Product reference before using this option.
+            NOTE: the next version of the Subscription API will remove
+            the clip tool option and always clip to the source geometry.
+            Thus this is a preview of the next API version's default
+            behavior.
 
     Returns:
         A Python dict representation of a Subscriptions API request for
@@ -99,7 +101,6 @@ def build_request(name: str,
     ...
 
     ```
-
     """
     # Because source and delivery are Mappings we must make copies for
     # the function's return value. dict() shallow copies a Mapping
--- a/planet/subscription_request.py
+++ b/planet/subscription_request.py
@@ -50,7 +50,7 @@ def build_request(name: str,
                   notifications: Optional[Mapping] = None,
                   tools: Optional[List[Mapping]] = None,
                   clip_to_source=False) -> dict:
-    """Prepare a subscription request.
+    """Construct a Subscriptions API request.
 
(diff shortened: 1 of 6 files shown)


planet/cli/subscriptions.py
```diff
@@ -205,10 +205,10 @@ async def list_subscription_results_cmd(ctx,
...
     type=types.JSON(),
     help='Notifications JSON. Can be a string, filename, or - for stdin.')
-@click.option('--tools',
-              type=types.JSON(),
-              help='Toolchain JSON. Can be a string, filename, or - for stdin.'
-              )
+@click.option(
+    '--tools',
+    type=types.JSON(),
+    help='Toolchain JSON. Can be a string, filename, or - for stdin.')
```
sgillies, line 211: This is all on yapf and not explained by anything in its change log :unamused:

planet/subscription_request.py
```diff
@@ -45,12 +45,27 @@
...
+        notifications: Specify notifications via email/webhook.
+        tools: Tools to apply to the products. The order of operation
+            is determined by the service.
+        clip_to_source: whether to clip to the source geometry or not
+            (the default). If True a clip configuration will be added
+            to the list of requested tools unless an existing clip tool
+            exists. NOTE: the next version of the Subscription API will
+            remove the clip tool option and always clip to the source
+            geometry. Thus this is a preview of the next API version's
+            default behavior.
```
sgillies, line 68: Moving this above the example.

planet/subscription_request.py
```diff
@@ -45,12 +45,27 @@
 
 
 def build_request(name: str,
-                  source: dict,
-                  delivery: dict,
-                  notifications: Optional[dict] = None,
-                  tools: Optional[List[dict]] = None) -> dict:
+                  source: Mapping,
```
sgillies, line 48: Typing best practices (https://typing.readthedocs.io/en/latest/source/best_practices.html) is to accept mappings (most general) and return dicts (most specific). This is in accord with Postel's Law.

planet/subscription_request.py
```diff
@@ -77,31 +92,36 @@ def build_request(name: str,
...
-        name: Name of the subscription.
-        source: A source for the subscription, i.e. catalog.
-        delivery: A delivery mechanism e.g. GCS, AWS, Azure, or OCS.
-        notifications: Specify notifications via email/webhook.
-        tools: Tools to apply to the products. Order defines
-            the toolchain order of operatations.
     """
-    details = {"name": name, "source": source, "delivery": delivery}
+    details = {
+        "name": name, "source": dict(source), "delivery": dict(delivery)
```
sgillies, line 97: Here is the consequence of accepting mappings: we need to explicitly convert them to dicts here. The upside is that we will never modify dicts that users pass in to this function, which can cause all kinds of subtle bugs.

planet/cli/subscriptions.py
sgillies: And it's version dependent. With Python 3.7, yapf puts '--tools' on line 208. WTH.

sgillies: We're now using Python 3.8 for the CI linting checks.

planet/subscription_request.py
sgillies: We copy information now, which has some overhead, but makes calling code more reliable. And building a subscriptions request doesn't need to be super optimized.

PR #972, opened 2023-07-04
Update CI to add 3.12 and change yapf check to use 3.8

sgillies:
Yapf 0.40.1 has different behavior for Python versions 3.7 and 3.8. Since 3.7 is EOL, we'll make 3.8 the new baseline.

### DIFF
--- a/noxfile.py
+++ b/noxfile.py
@@ -37,7 +37,7 @@ def coverage(session):
     session.run('coverage', 'report')
 
 
-@nox.session(python=["3.7", "3.8", "3.9", "3.10", "3.11"])
+@nox.session(python=["3.8", "3.9", "3.10", "3.11", "3.12"])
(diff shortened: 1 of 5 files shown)


noxfile.py
```diff
@@ -37,15 +37,25 @@ def coverage(session):
...
     # https://docs.python.org/3/library/warnings.html#temporarily-suppressing-warnings
-    session.run('pytest', '--ignore', 'examples/', '-v', '-W=error', *options)
+    session.run('python',
+                '-m',
+                'pytest',
+                '--ignore',
+                'examples/',
+                '-v',
+                '-Werror',
+                '-Wignore::DeprecationWarning:tqdm.std',
```
sgillies, line 57: tqdm is using a datetime method that is deprecated in Python 3.12.

Issue #933, opened 2023-04-19
Support for upcoming csv timeseries subscription results


cpaulik:
**Is your feature request related to a problem? Please describe.**

With the introduction of PVs to the subscription API we are also supporting timeseries in csv format.

This is available on api.planet.com/subscriptions/v1 but not yet documented offically.

**Describe the solution you'd like**

I would like to use the `format=csv` option of the subscription API. 

Example request:

```
https://api.planet.com/subscriptions/v1/<uuid>/results?format=csv
```

**Additional context**

[Internal document](https://docs.google.com/document/d/1dj8qPeHy4a7vR6Mu2HVuO2bfPaDU_m4y9ExBC_RFpXY/edit#heading=h.pc8fh9xlg1ef) explaining the endpoint.

PR #976, opened 2023-07-07
Add a planetary_variable_source subscription helper

### LINKED ISSUE #966: Add a Planetary Variable source helper to the subscription_request module
To generate source objects like the one shown at https://developers.planet.com/docs/subscriptions/pvs-subs/#create-a-planetary-variables-subscription. It should be based on the existing `catalog_source()` helper https://github.com/planetlabs/planet-client-python/blob/main/planet/subscription_request.py#L99.

sgillies:
Resolves #966. API doc improvements for `catalog_source()` are included in this PR.

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -1,10 +1,11 @@
 2.1.0 (TBD)
 
 Added:
-- A subscription_request.planetary_variable_source function has been added ().
+- A subscription_request.planetary_variable_source function has been added
+  (#976).
 - The subscription_request.build_request function has a new option to clip to
-  the subscription's source geometry. This is a preview of the default
-  behavior of the next version of the Subscriptions API ().
+  the subscription's source geometry. This is a preview of the default behavior
+  of the next version of the Subscriptions API (#971).
 
 2.0.3 (2023-06-28)
 
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -1,9 +1,10 @@
 2.1.0 (TBD)
 
 Added:
+- A subscription_request.planetary_variable_source function has been added ().
 - The subscription_request.build_request function has a new option to clip to
   the subscription's source geometry. This is a preview of the default
-  behavior of the next version of the Subscriptions API.
+  behavior of the next version of the Subscriptions API ().
 
 2.0.3 (2023-06-28)
 
--- a/planet/subscription_request.py
+++ b/planet/subscription_request.py
@@ -13,7 +13,7 @@
 # the License.
 """Functionality for preparing subscription requests."""
 from datetime import datetime
-from typing import Any, Dict, Optional, List, Mapping
+from typing import Any, Dict, Optional, List, Literal, Mapping
 
 from . import geojson, specs
 from .exceptions import ClientError
@@ -73,12 +73,12 @@ def build_request(name: str,
             behavior.
 
     Returns:
-        A Python dict representation of a Subscriptions API request for
-        a new subscription.
+        dict: a representation of a Subscriptions API request for
+            a new subscription.
 
     Raises:
-        ClientError when a valid Subscriptions API request can't be
-        constructed.
+        ClientError: when a valid Subscriptions API request can't be
+            constructed.
 
     Examples:
     ```python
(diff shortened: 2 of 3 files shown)


sgillies:
@cpaulik while I work on #933, would you be willing to see if this is what you're looking for in a PV source helper?

planet/subscription_request.py
```diff
@@ -212,6 +219,86 @@ def catalog_source(
...
+    The return value can be passed to
+    [planet.subscription_request.build_request][].
+
+    Note: this function does not validate variable types and ids.
+
+    Parameters:
+        var_type: one of "biomass_proxy", "land_surface_temperature",
+            "soil_water_content", or "vegetation_optical_depth".
+        var_id: a value such as "SWC-AMSR2-C_V1.0_100" for soil water
+            content derived from AMSR2 C band.
```
kevinlacaille, line 248: Would you be opposed to pointing a user to all of the possible `var_id`s? I.e., 
```suggestion
        var_id: a value such as "SWC-AMSR2-C_V1.0_100" for soil water
            content derived from AMSR2 C band. See all options in the [Planetary Variables docs](https://developers.planet.com/docs/subscriptions/pvs-subs/#create-a-planetary-variables-subscription).
```

planet/subscription_request.py
```diff
@@ -212,6 +219,86 @@ def catalog_source(
...
+    Returns:
+        dict: a representation of a subscription source.
+
+    Raises:
+        ClientError: if a source can not be
+            configured.
+
+    Examples:
+        ```python
+        >>> source = planetary_variables_source(
```
kevinlacaille, line 265: Typo
```suggestion
        >>> source = planetary_variable_source(
```

tests/unit/test_subscription_request.py
```diff
@@ -346,3 +346,23 @@ def test_toar_tool_success():
...
+
+
+def test_pv_source_success(geom_geojson):
+    """Configure a planetary variable subscription source."""
+    # NOTE: this function does not yet validate type and id.
+    # The nonsense values are intended to fail when the function does
+    # add validation.
+    source = subscription_request.planetary_variable_source(
+        "var1",
+        "VAR1-abcd",
```
kevinlacaille, line 358: Would you be opposed to having real types and IDs here? That way:
1. when someone reads the test they can have an example of working code
2. the test will already pass when we do implement validation of types and ids.

I do appreciate keeping things simple, but I wonder if this small change is worth it.
```suggestion
        "soil_water_content",
        "SWC-AMSR2-C_V1.0_100",
```

kevinlacaille (approved):
A few small suggestions and 1 must-fix typo in the example code. Otherwise, looks good to me!

CHANGES.txt
```diff
@@ -1,9 +1,11 @@
 2.1.0 (TBD)
 
 Added:
+- A subscription_request.planetary_variable_source function has been added
+  (#976).
 - The subscription_request.build_request function has a new option to clip to
-  the subscription's source geometry. This is a preview of the default
-  behavior of the next version of the Subscriptions API.
+  the subscription's source geometry. This is a preview of the default behavior
```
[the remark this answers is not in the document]
sgillies: :canada:

planet/subscription_request.py
```diff
@@ -212,6 +219,86 @@ def catalog_source(
...
+) -> dict:
+    """Construct a Planetary Variable subscription source.
+
+    Planetary Variables come in 4 types and are further subdivided
+    within these types. See [Subscribing to Planetary
+    Variables](https://developers.planet.com/docs/subscriptions/pvs-subs/#planetary-variables-types-and-ids)
+    for details.
+
+    The return value can be passed to
+    [planet.subscription_request.build_request][].
```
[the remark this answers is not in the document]
sgillies: @kevinlacaille it does! https://mkdocstrings.github.io/usage/#cross-references

planet/subscription_request.py
sgillies: @kevinlacaille you mean link to https://developers.planet.com/docs/subscriptions/pvs-subs/#planetary-variables-types-and-ids instead? Or something else?

tests/unit/test_subscription_request.py
sgillies: @kevinlacaille I get what you mean, but I'm intending that this test should fail if/when we do have some validation. Maybe the thing to do is both.

planet/subscription_request.py
```diff
@@ -212,6 +219,86 @@ def catalog_source(
     return {"type": "catalog", "parameters": parameters}
 
 
+def planetary_variable_source(
+    var_type: Literal["biomass_proxy",
```
cpaulik, line 223: Just wanted to point out that internally we are already using more sources for upcoming forest products.

I'm not sure if there is a way to not duplicate these lists.

planet/subscription_request.py
kevinlacaille: Oh yes, that's what I meant to send. I suppose this is optional, but especially for non-intuitive names, this would be helpful. Thoughts?

tests/unit/test_subscription_request.py
kevinlacaille: Makes sense to me!

planet/subscription_request.py
sgillies: @cpaulik I'm going off what we have at https://developers.planet.com/docs/subscriptions/pvs-subs/#planetary-variables-types-and-ids. Do you know if we publish a machine-readable version of this table for customers?

I'm using the Literal type hints here so that developers get value hints from their IDEs. The SDK does not at this time validate the values at runtime. Thus you can use any values you want, including unavailable PV types and ids. See for example https://github.com/planetlabs/planet-client-python/pull/982/files#diff-661d60b3722e519e66594c89a655b8aa191b919ac27c98ecffde58f984ab120eR355.

cpaulik: You sparked my curiosity. Since our API docs spell them out I did some digging and found that https://api.planet.com/subscriptions/v1/spec includes the sources types under `schemas.PVSourceTypes`

PR #981, opened 2023-07-11
Support for CSV formatted results of subscriptions

### LINKED ISSUE #933: Support for upcoming csv timeseries subscription results

sgillies:
Resolves #933. The big choice here is to make CSV format access its own method instead of a flag on the existing method. I feel like this is warranted by the API split between JSON (paged) and CSV (not paged).

Sanity check @kevinlacaille @cpaulik ?

### DIFF
--- a/planet/clients/subscriptions.py
+++ b/planet/clients/subscriptions.py
@@ -1,7 +1,7 @@
 """Planet Subscriptions API Python client."""
 
 import logging
-from typing import AsyncIterator, Optional, Set
+from typing import AsyncIterator, Literal, Optional, Sequence
 
 from planet.exceptions import APIError, ClientError
 from planet.http import Session
@@ -58,7 +58,7 @@ class SubscriptionsClient:
             self._base_url = self._base_url[:-1]
 
     async def list_subscriptions(self,
-                                 status: Optional[Set[str]] = None,
+                                 status: Optional[Sequence[str]] = None,
                                  limit: int = 100) -> AsyncIterator[dict]:
         """Iterate over list of account subscriptions with optional filtering.
 
@@ -216,16 +216,21 @@ class SubscriptionsClient:
 
     async def get_results(self,
                           subscription_id: str,
-                          status: Optional[Set[str]] = None,
+                          status: Optional[Sequence[Literal[
+                              "created",
+                              "queued",
+                              "processing",
+                              "failed",
+                              "success"]]] = None,
                           limit: int = 100) -> AsyncIterator[dict]:
         """Iterate over results of a Subscription.
 
-        Note:
+        Notes:
             The name of this method is based on the API's method name. This
             method provides iteration over results, it does not get a
             single result description or return a list of descriptions.
 
-        Args:
+        Parameters:
             subscription_id (str): id of a subscription.
             status (Set[str]): pass result with status in this set,
                 filter out results with status not in this set.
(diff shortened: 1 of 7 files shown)


planet/clients/subscriptions.py
```diff
@@ -263,3 +267,45 @@ class _ResultsPager(Paged):
...
+            ClientError: on a client error.
+        """
+        url = f'{self._base_url}/{subscription_id}/results'
+        params = {'status': [val for val in status or {}], 'format': 'csv'}
+
+        # Note: retries are not implemented yet. This project has
+        # retry logic for HTTP requests, but does not handle errors
+        # during streaming. We may want to consider a retry decorator
+        # for this entire method a la stamina:
+        # https://github.com/hynek/stamina.
```
sgillies, line 307: I think that our model for retrying requests doesn't cover streaming cases cases.

tests/unit/test_subscription_request.py
```diff
@@ -346,3 +346,23 @@ def test_toar_tool_success():
 
     expected = {"type": "toar", "parameters": {"scale_factor": 12345}}
     assert res == expected
+
```
sgillies, line 349: Fuuu... not sure how this got in here.

planet/subscription_request.py
```diff
@@ -212,6 +219,86 @@ def catalog_source(
     return {"type": "catalog", "parameters": parameters}
 
 
+def planetary_variable_source(
+    var_type: Literal["biomass_proxy",
```
cpaulik, line 223: Just wanted to point out that internally we are already using more sources for upcoming forest products. 

I'm not sure if there is a way to not duplicate these lists.

cpaulik:
Thanks. Using this via `planet subscriptions results UUID --csv` makes sense to me. 

Are we clear in the documentation about how to get the output into a CSV file? I think we have users for which that might not be obvious.

cpaulik (approved):

planet/subscription_request.py
cpaulik: Just saw #976 after this. Moving to that MR.

tests/integration/test_subscriptions_api.py
```diff
@@ -276,6 +281,18 @@ async def test_get_results_success():
...
 
 
+@pytest.mark.anyio
+@res_api_mock
+async def test_get_results_csv():
+    """Subscription CSV fetched, has the expected items."""
+    async with Session() as session:
+        client = SubscriptionsClient(session, base_url=TEST_URL)
+        results = [res async for res in client.get_results_csv("42")]
+        import csv
```
kevinlacaille, line 291: I'm surprised `yapf` doesn't care about this import's placement.

sgillies:
@guyon-planet in the future, let me merge these. I have a couple changes to make yet.

guyon-planet:
@sgillies, sorry about that

Issue #963, opened 2023-05-30
Unable to receive zipped deliveries to GCP


kevinlacaille:
**Expected behavior**
Given this style `cloudconfig.json` file, as specified in order_request ([here](https://github.com/planetlabs/planet-client-python/blob/main/planet/order_request.py#L165-L200)) and validated in specs ([here](https://github.com/planetlabs/planet-client-python/blob/main/planet/specs.py#L88-L91)) I expect to be able to receive a zipped file delivered my GCS bucket.

```console
{
    "google_cloud_storage": {
        "bucket": "pfed_external_share",
        "path_prefix": "zip_test",
        "credentials": "my-creds"
    },
    "archive_type": "zip"
}
```

Untested, but this might be a global issue (i.e., an issue with AWS, GCS, Azure, etc.).

**Actual behavior (describe the problem)**
My order is not delivered zipped.

**Related Issues**
N/A

**Workaround**
N/A

**Minimum, Complete, Viable Code Sample**
Apparently, one must also set a `"archive_filename"`  (e.g. `"archive_filename": "{{name}}_{{order_id}}.zip"` ) or use `"single_archive": true` to actually get zipped deliveries.

```console
{
    "google_cloud_storage": {
        "bucket": "pfed_external_share",
        "path_prefix": "zip_test",
        "credentials": "my-creds"
    },
    "archive_type": "zip",
    "archive_filename": "{{name}}_{{order_id}}.zip"
}
```
or
```console
{
    "google_cloud_storage": {
        "bucket": "pfed_external_share",
        "path_prefix": "zip_test",
        "credentials": "my-creds"
    },
    "archive_type": "zip",
    "single_archive": true
}
```

jreiberkyle:
Yep, looks like we need some more logic added to `delivery` for handling zipping.

JSee98:
Hi @kevinlacaille, this seems like a good first-time issue to me. I looked at the following function that updates the download config with the archive file name and other related details https://github.com/planetlabs/planet-client-python/blob/79d9a3cb952fcd4f75e5e935ee067455580c779d/planet/order_request.py#L165-L200

The issue seems to be the fact that archive_name and single_archive are optional fields and a check isn't performed, specifically, `"archive_type": "zip"` is defined and the following two fields are undefined, thus, causing this issue.

Let me know if I can pick this up.

kevinlacaille:
Hi, @JSee98! Glad to see you back at it here :) 

[...]

PR #990, opened 2023-07-13
Add archive options to planet-orders-request

### LINKED ISSUE #970: The --cloudconfig CLI option  has an innaccurate name
The `--cloudconfig` option of the `planet orders create` CLI command has an inaccurate name, as it is not limited to cloud delivery configurations. 

It is entirely possible to provide a local delivery config using the `--cloudconfig` option, e.g.:
```
{  
    "archive_type": "zip",
    "single_archive": true,
    "archive_filename": "{{name}}_{{order_id}}.zip"
}
```

The option should be named `--delivery` or `--deliveryconfig`, and [the docs](https://planet-sdk-for-python-v2.readthedocs.io/en/latest/cli/cli-orders/#cloud-delivery) updated accordingly. The [equivalent SDK method](https://planet-sdk-for-python-v2.readthedocs.io/en/latest/python/sdk-reference/#planet.order_request.delivery) is more accurately named `delivery`

sgillies:
And make --delivery and alias for --cloudconfig.

Resolves #970

### DIFF
--- a/planet/cli/orders.py
+++ b/planet/cli/orders.py
@@ -266,13 +266,14 @@ async def create(ctx, request: str, pretty):
               show_default=True,
               help="Optionally zip archive all item bundles together.")
 @click.option(
-    '--cloudconfig',
     '--delivery',
+    '--cloudconfig',
     type=types.JSON(),
-    help=("Credentials for cloud storage provider to enable cloud delivery of "
-          "data. Can be a json string, filename, or '-' for stdin. "
-          "Optionally, archive parameters can be set using the same JSON. "
-          "The --delivery option is an alias for this use case."))
+    help=("Delivery configuration, which may include credentials for a cloud "
+          "storage provider, to enable cloud delivery of data, and/or "
+          "parameters for bundling deliveries as zip archives. Can be a JSON "
+          "string, a filename, or '-' for stdin. The --cloudconfig option is "
+          "an alias for this use case."))
 @click.option(
     '--stac/--no-stac',
     default=True,
@@ -292,7 +293,7 @@ async def request(ctx,
                   archive_type,
                   archive_filename,
                   single_archive,
-                  cloudconfig,
+                  delivery,
                   stac,
                   pretty):
     """Generate an order request.
@@ -324,7 +325,7 @@ async def request(ctx,
     delivery = planet.order_request.delivery(archive_type=archive_type,
                                              archive_filename=archive_filename,
                                              single_archive=single_archive,
-                                             cloud_config=cloudconfig)
+                                             cloud_config=delivery)
 
     if stac and "google_earth_engine" not in delivery:
         stac_json = {'stac': {}}
--- a/planet/cli/orders.py
+++ b/planet/cli/orders.py
@@ -253,11 +253,26 @@ async def create(ctx, request: str, pretty):
               default=False,
               is_flag=True,
               help='Send email notification when order is complete.')
+@click.option('--archive-type',
+              type=click.Choice(['zip']),
+              help="Optionally zip archive each item bundle.")
+@click.option('--archive-filename',
+              default='{{name}}-{{order_id}}.zip',
+              show_default=True,
+              help="Templated filename for archived bundles or orders.")
+@click.option('--single-archive',
+              is_flag=True,
+              default=False,
+              show_default=True,
+              help="Optionally zip archive all item bundles together.")
 @click.option(
     '--cloudconfig',
+    '--delivery',
     type=types.JSON(),
-    help="""Credentials for cloud storage provider to enable cloud delivery of
-    data. Can be a json string, filename, or '-' for stdin.""")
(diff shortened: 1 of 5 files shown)


planet/cli/orders.py
```diff
@@ -253,11 +253,26 @@ async def create(ctx, request: str, pretty):
...
+              show_default=True,
+              help="Templated filename for archived bundles or orders.")
+@click.option('--single-archive',
+              is_flag=True,
+              default=False,
+              show_default=True,
+              help="Optionally zip archive all item bundles together.")
 @click.option(
     '--cloudconfig',
+    '--delivery',
```
sgillies, line 270: Here's the alias.

planet/cli/orders.py
```diff
@@ -253,11 +253,26 @@ async def create(ctx, request: str, pretty):
...
     type=types.JSON(),
-    help="""Credentials for cloud storage provider to enable cloud delivery of
-    data. Can be a json string, filename, or '-' for stdin.""")
+    help=("Credentials for cloud storage provider to enable cloud delivery of "
+          "data. Can be a json string, filename, or '-' for stdin. "
+          "Optionally, archive parameters can be set using the same JSON. "
+          "The --delivery option is an alias for this use case."))
```
sgillies, line 275: Too confusing? Or okay?

planet/cli/orders.py
```diff
@@ -303,14 +321,12 @@ async def request(ctx,
         except planet.exceptions.ClientError as e:
             raise click.BadParameter(e)
 
-    if cloudconfig:
-        delivery = planet.order_request.delivery(cloud_config=cloudconfig)
-        if "google_earth_engine" in cloudconfig:
```
sgillies, line 308: I moved this test closer to where it is needed.

planet/order_request.py
```diff
@@ -199,13 +199,17 @@ def delivery(archive_type: Optional[str] = None,
     if archive_type:
         archive_type = specs.validate_archive_type(archive_type)
 
-    fields = ['archive_type', 'single_archive', 'archive_filename']
-    values = [archive_type, single_archive, archive_filename]
+    config: Dict[str, Any] = {}
 
-    config = dict((k, v) for k, v in zip(fields, values) if v)
+    if archive_type:
```
sgillies, line 204: This covers some of the needs of #963, I think.

planet/order_request.py
```diff
@@ -199,13 +199,17 @@ def delivery(archive_type: Optional[str] = None,
...
+        config.update(archive_type=archive_type,
+                      archive_filename=archive_filename
+                      or '{{name}}-{{order_id}}.zip',
```
sgillies, line 207: Here's a default archive filename.

sgillies: Note that unless the archive type is given, there are no archive items in the "delivery" config.

sgillies:
@kevinlacaille @tbarsballe what do you think of this attempt to solve #970 by adding the missing individual archive options that complement `--cloudconfig` as well as a `--delivery` alias?

planet/cli/orders.py
tbarsballe: If feels a little unclear as is. I would reword such that --cloudconfig is the alias, and merge the archive parameter line into the first sentence. E.g.:
```
"Delivery configuration, which may include credentials for a cloud storage "
"provider, to enable cloud delivery of data, and/or parameters for bundling "
"deliveries as zip archives. Can be a json string, filename, or '-' for stdin. "
"The --cloudconfig option is an alias for this use case."
```

planet/cli/orders.py
```diff
@@ -253,11 +253,26 @@ async def create(ctx, request: str, pretty):
...
+              help="Optionally zip archive each item bundle.")
+@click.option('--archive-filename',
+              default='{{name}}-{{order_id}}.zip',
+              show_default=True,
+              help="Templated filename for archived bundles or orders.")
+@click.option('--single-archive',
+              is_flag=True,
+              default=False,
+              show_default=True,
+              help="Optionally zip archive all item bundles together.")
```
tbarsballe, line 267: This is a great addition, thanks for including it.

tbarsballe (commented):
Looks pretty good to me. I like keeping --cloudconfig an having --delivery as an alias - that obviate's my concern around changing the parameter being too user-unfriendly.

I did suggest a rewording of the parameter help, though it is alright as is.

planet/cli/orders.py
sgillies: Yes indeed, this was an oversight. It's easy to imagine users only using, for example, `--archive-type zip` and no cloud config at all.

sgillies:
Thanks for the feedback @tbarsballe !

planet/cli/orders.py
tbarsballe: Great, looks good to me.

tbarsballe (approved):

PR #1007, opened 2023-07-29
release 2.1.2

jreiberkyle:
2.1.2 (2023-06-28)

Fixed:
- A single HTTP error response would cause an entire file download to error out.
  Now, the download is retried using the same logic as retries of other HTTP
  requests.


jreiberkyle:
@sgillies This release is ready to go! The next step in the Release Workflow is `Create a new github release:`

sgillies:
We're not quite ready to release.

PR #1011, opened 2023-08-01
New take on retrying order asset downloads

### LINKED ISSUE #1010: Eliminate double download of assets and orders
#1002 introduces a double download. One at https://github.com/planetlabs/planet-client-python/pull/1002/files#diff-eb40861f30b566f7a4a569eb3e052b9edccf30400cbe3a309c34f914aab79cd1R600 and one at https://github.com/planetlabs/planet-client-python/pull/1002/files#diff-eb40861f30b566f7a4a569eb3e052b9edccf30400cbe3a309c34f914aab79cd1R620. The implementation of #1002 only requires the first to be an initial connection to get the Content-length header. We may want to change the implementation, but changing the `get` to `stream` is a good fix for the moment.

sgillies:
Wraps the entire orders.download_asset() in a retry, eliminating the double download and concentrating the retry complexity instead of spreading it over several modules.

Resolves #1010

### DIFF
--- a/planet/clients/orders.py
+++ b/planet/clients/orders.py
@@ -14,19 +14,21 @@
 # the License.
 """Functionality for interacting with the orders api"""
 import asyncio
-import logging
-from pathlib import Path
-import time
-from typing import AsyncIterator, Callable, List, Optional
-import uuid
-import json
 import hashlib
+import json
+import logging
+import re
+import time
+import uuid
+from pathlib import Path
+from typing import AsyncIterator, Callable, List, Optional
 
+import stamina
 from tqdm.asyncio import tqdm
 
 from .. import exceptions
 from ..constants import PLANET_BASE_URL
-from ..http import Session
+from ..http import RETRY_EXCEPTIONS, Session
 from ..models import Paged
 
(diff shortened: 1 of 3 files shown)


tests/integration/test_orders_api.py
```diff
@@ -587,9 +587,7 @@ async def _stream_img():
     # an error caused by respx and not this code
     # https://github.com/lundberg/respx/issues/130
     respx.get(dl_url).side_effect = [
-        httpx.Response(HTTPStatus.OK,
-                       headers=img_headers,
-                       request='donotcloneme'),
```
sgillies, line 592: Behavior of this PR no longer depends on multiple download requests.

planet/clients/orders.py
```diff
@@ -257,31 +260,49 @@ async def download_asset(self,
                 limit is exceeded.
 
         """
-        response = await self._session.request(method='GET', url=location)
-        filename = filename or response.filename
-        length = response.length
-        if not filename:
-            raise exceptions.ClientError(
-                f'Could not determine filename at {location}')
+        async with self._session._limiter, self._session._client.stream('GET', location) as resp:
```
sgillies, line 263: Rate limiter is used here.

PR #1016, opened 2023-11-22
Bugfix datetime serialization utc zero offset

stefan-jozefowicz-bayer:
**Proposed Changes:**

1.

Not intended for changelog:

1. Change how timezone aware datetimes with UTC tz are converted to datetime string

**Diff of User Interface**

Old behavior:
passing a timezone aware datetime from the UTC zone to some of the request constructors resulted in malformed datetime strings, e.g: `2022-06-01T01:00:00+00:00Z` - i.e. strings that have BOTH `+00:00` and `Z` at the end. 

At the root of it was `value.utcoffset()` >> `timedelta(hours=0)` evaluating as False even though there is a defined offset here:
```
def _datetime_to_rfc3339(value: datetime) -> str:
    """Converts the datetime to an RFC3339 string"""
    iso = value.isoformat()
    if not value.utcoffset():
        # rfc3339 needs a Z if there is no timezone offset
        iso += 'Z'
    return iso
```

New behavior:
The Z is tagged on only if the utcoffset() is None.

When the TZ is UTC the dt string is:
 `2022-06-01T01:00:00+00:00` 

The previous test setup for timezone aware datetimes obfuscated this bug, so I changed the tests as well.

**PR Checklist:**

### DIFF
--- a/planet/data_filter.py
+++ b/planet/data_filter.py
@@ -149,7 +149,7 @@ def date_range_filter(field_name: str,
 def _datetime_to_rfc3339(value: datetime) -> str:
     """Converts the datetime to an RFC3339 string"""
     iso = value.isoformat()
-    if not value.utcoffset():
+    if value.utcoffset() is None:
         # rfc3339 needs a Z if there is no timezone offset
         iso += 'Z'
     return iso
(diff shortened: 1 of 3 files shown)


pl-gideon (approved):
Thank you for fixing this issue!

PR #1027, opened 2024-03-20
Added field_boundaries_sentinel_2_p1m for Field Boundaries PV support

### LINKED ISSUE #1026: Add support for Field Boundaries PV
Sorta? I'd like to request field boundaries PV subscription, but I can't with `request-pv`

Add field boundaries to the --var-type (and any other places it's needed)

Using another var-type and then hand modifying the json.

**Additional context**
It'd also be awesome if the client would provide feedback on the time ranges that are acceptable, as for field boundaries they are a bit narrow. But that's likely a lot more complexity and would likely mean having to keep the client in sync with the server.

cholmes:
**Related Issue(s):**

Closes #1026

I modeled this off of https://github.com/planetlabs/planet-client-python/commit/4c920fde74706da7724323522f606586815ab583 which didn't have any tests, so hopefully none are needed. Just adds an option for the new https://developers.planet.com/docs/planetary-variables/field-boundaries/

**Proposed Changes:**

1. Support for field_boundaries_sentinel_2_p1m in Subscriptions API

Not intended for changelog:

1. 

**Diff of User Interface**

Old behavior: No mention of `field_boundaries_sentinel_2_p1m`, and failure if you try to use it.

New behavior: `field_boundaries_sentinel_2_p1m` can be provided to a PV subscription, and it shows up as an option in the CLI.

**PR Checklist:**

- [] This PR is as small and focused as possible
- [] If this PR includes proposed changes for inclusion in the changelog, the title of this PR summarizes those changes and is ready for inclusion in the Changelog.
- [] This PR does not break any examples or I have updated them

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -1,3 +1,8 @@
+
+
+Added:
+ - Support for field_boundaries_sentinel_2_p1m in Subscriptions API (#1026)
+
 2.4.0 (2024-03-19)
 
 Added:
--- a/mkdocs.yml
+++ b/mkdocs.yml
@@ -49,7 +49,6 @@ extra_css:
     
 plugins:
   - search
-  - with-pdf
   - mkdocstrings:
       handlers:
         python:
--- a/mkdocs.yml
+++ b/mkdocs.yml
@@ -49,6 +49,7 @@ extra_css:
     
 plugins:
   - search
+  - with-pdf
   - mkdocstrings:
       handlers:
         python:
--- a/planet/cli/subscriptions.py
+++ b/planet/cli/subscriptions.py
@@ -373,7 +373,8 @@ def request_catalog(item_types,
         "land_surface_temperature",
         "soil_water_content",
         "vegetation_optical_depth",
-        "forest_carbon_diligence_30m"
+        "forest_carbon_diligence_30m",
+        "field_boundaries_sentinel_2_p1m"
     ]),
 )
 @click.option('--var-id', required=True, help='Planetary variable id.')
--- a/planet/subscription_request.py
+++ b/planet/subscription_request.py
@@ -276,7 +276,8 @@ def planetary_variable_source(
                       "land_surface_temperature",
                       "soil_water_content",
                       "vegetation_optical_depth",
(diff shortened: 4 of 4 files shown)


CHANGES.txt
```diff
@@ -1,3 +1,8 @@
+
+
+Added:
```
asonnenschein, line 3: nit: The changelog update can also wait until the release PR.  Not a blocker (more of an FYI) because this PR is branched off dev version [2.5dev](https://github.com/planetlabs/planet-client-python/blob/fb-support-1026/planet/__version__.py#L1).  Just remember to add the release version to this changelog entry in the release PR!

asonnenschein (approved):

mkdocs.yml
```diff
@@ -49,6 +49,7 @@ extra_css:
     
 plugins:
   - search
+  - with-pdf
```
asonnenschein, line 52: This addition is causing the [check docs](https://github.com/planetlabs/planet-client-python/actions/runs/8354173225/job/22867090232?pr=1027#step:5:51) build step to fail, because `with-pdf` is not installed.  I haven't estimated the level of effort for adding this plugin; do we need it here or is it safe to remove `with-pdf` from this PR and include it in a follow up PR?

CHANGES.txt
cholmes: Cool. I tried to look a bit to see what the best practice was - to put it in now or wait till release. Feel free to remove if the standard is just to have whoever is releasing write the notes - just figured I'd try to be helpful.

mkdocs.yml
cholmes: Ah, my bad. That was an experiment I did (and forgot about - I haven't used my local repo in a year). I'll remove.

cholmes: Removed. The plugin was to be able to output a big pdf to then be able to put all the docs into a custom GPT, which was nice. I think it did depend on some alternate installation (like not easily done with pip install), so probably not worth the effort of including in the main stuff. It was more of an experiment, I'd just totally forgot about it - even when I saw the build error.

PR #1032, opened 2024-04-15
Allow use of feature references in the sdk for subscriptions + orders

angaither:
**Related Issue(s):**

Closes #

**Proposed Changes:**

1. Allow [feature references](https://developers.planet.com/docs/apis/features/) to be used in the sdk for creating subscriptions (with or without a source block) and for orders. 

Not intended for changelog:

1. 

**Diff of User Interface**

Old behavior:
geojson was required to make a subscription or order. 

New behavior:
geojson or a geometry feature reference can be used 
```json
{
        "type": "ref",
        "content": "pl:features/my/water-fields-RqB0NZ5/rmQEGqm",
} 
```

**PR Checklist:**


ischneider:
There's a number of `geometry: Optional[dict]` though the typing intention is now including `str` (ref)

I haven't dug enough to see if it's even possible (or useful) to try to specify a GeoJSON type (like by applying more type variables to dict keys, though honestly it doesn't feel useful)...

But something like this might be a nice compromise (untested in mypy or other static checker)

```
from typing import Union

# GeoJSON is a python dict that conforms to the specification
GeoJSON = dict

# GeoRef is a reference to a feature, either as a file or URI, that if resolved, would yield GeoJSON
GeomRef = str

# Geom is some inline or reference to a Geometry
Geom = Union[GeoJSON, GeomRef]
```

planet/cli/subscriptions.py
```diff
@@ -437,6 +451,7 @@ def request_pv(var_type, var_id, geometry, start_time, end_time, pretty):
     Variables](https://developers.planet.com/docs/subscriptions/pvs-subs/#planetary-variables-types-and-ids)
     for details.
     """
+    # print("Geom is", geometry)
```
asonnenschein, line 454: Can this print statement be removed?

tests/integration/test_subscriptions_cli.py
```diff
@@ -321,12 +321,17 @@ def test_request_base_success(invoke, geom_geojson):
...
 
 
-def test_request_base_clip_to_source(invoke, geom_geojson):
+@pytest.mark.parametrize("geom_fixture",
+                         [('geom_geojson'), ('geom_reference'),
+                          ("str_geom_reference")])
+def test_request_base_clip_to_source(geom_fixture, request, invoke):
     """Clip to source using command line option."""
+    geom = request.getfixturevalue(geom_fixture)
+    print("geom is", geom)
```
asonnenschein, line 330: nit: remove print statement.

planet/cli/subscriptions.py
angaither: done!

planet/cli/subscriptions.py
```diff
@@ -29,6 +30,17 @@ def check_item_types(ctx, param, item_types) -> Optional[List[dict]]:
         raise click.BadParameter(str(e))
 
 
+def check_geom(ctx, param, geometry) -> Optional[dict]:
```
asonnenschein, line 33: This method looks like it's defined in both `cli/data.py` and `cli/subscriptions.py`.  Is there a place where we can define `check_geom()` once and share it as needed across different files?

asonnenschein:
I see this MR includes integration tests for the Data API/CLI, and the Subscriptions CLI.  @angaither Do you think we also need tests for the Subscriptions API and Orders API/CLI?

planet/cli/subscriptions.py
angaither: that seemed to be the pattern for other cli methods. [check_item_type/s](https://github.com/planetlabs/planet-client-python/blob/main/planet/cli/data.py#L73) in data, [check_item_type/s](https://github.com/planetlabs/planet-client-python/blob/main/planet/cli/subscriptions.py#L21) in subs, both loops through and call a validator in another file.

asonnenschein: Hmm, that's a pattern I think we should move away from where possible.  Would you please move `check_geom()` to a place where it could be shared?  I believe that following DRY patterns will help make the SDK more maintainable going forward.

angaither: moved! I started moving check_item_type and check_item_types as well, but that turned into a can of 🪱s

asonnenschein: Thanks!  No worries about check_item_type/s, I would consider those methods to be outside the scope of this PR.

asonnenschein (approved):

Issue #1038, opened 2024-05-27
Geometry filter that worked in 2.6 does not work in 2.7


s6hebern:
**Expected behavior**
In version 2.6, the attached search filter worked fine. Attaching the original geojson as well (had to rename it to "json" to be able to upload).
[bunthaus.json](https://github.com/planetlabs/planet-client-python/files/15455674/bunthaus.json)
[planet_search_filter.json](https://github.com/planetlabs/planet-client-python/files/15455480/planet_search_filter.json)

**Actual behavior (describe the problem)**
In version 2.7, that same search filter returns the following error:
[planet_error.txt](https://github.com/planetlabs/planet-client-python/files/15455481/planet_error.txt)

**Minimum, Complete, Viable Code Sample**
Tried my best to extract the relevant functions from a much larger script.
```python
import datetime
import asyncio
import geopandas as gpd
from planet import Auth, data_filter, Session


def prepare_aoi(aoi):
    gdf = gpd.read_file(aoi)
    # make MultiPolygon
    if gdf.geometry.count() > 1:
        log.warn('Found multiple features! Morphing them into one multipart feature ...')
        multipart = gpd.tools.collect(gdf.geometry)
        crs = gdf.crs
        gdf = gpd.GeoDataFrame({'id': [0], 'geometry': [multipart]}, crs=crs)
        gdf.to_crs(epsg=4326, inplace=True)
    return json.loads(gdf.to_json())


async def search_scenes(geojson, start, end, api_key):
    search_filter = data_filter.and_filter([
        data_filter.permission_filter(),
        data_filter.string_in_filter('instrument', ['PSB.SD', ]),
        data_filter.date_range_filter('acquired', gte=start),
        data_filter.date_range_filter('acquired', lte=end),
        data_filter.geometry_filter(geojson),
    ])
    async with Session(auth=Auth.from_key(api_key)) as session:
        client = session.client('data')
        return {item.get('id'): item.get('properties').get('cloud_cover') * 100 async for item in
                client.search(['PSScene'], search_filter, limit=0)}


if __name__ == '__main__':
    # "MY_GEOJSON" needs to be the real file
    geojson = prepare_aoi(MY_GEOJSON)
    start = datetime.datetime(2023, 9, 25)
    end = datetime.datetime(2023, 9, 30)
    # provide working API key here
    await search_scenes(geojson, start, end, API_KEY)
```

**Workaround**
Downgrade to version 2.6

**Environment Information**
[...]

PR #1041, opened 2024-05-30
add Data API search tests using positional args

stephenhillier:
**Related Issue(s):**

#1038 

**Proposed Changes:**

Adds a test to the Data API client calling the search method using positional args.

### DIFF
--- a/tests/integration/test_data_api.py
+++ b/tests/integration/test_data_api.py
@@ -225,6 +225,43 @@ async def test_search_filter(item_descriptions,
 
 @respx.mock
 @pytest.mark.anyio
+async def test_search_filter_positional_args(item_descriptions,
+                                             search_filter,
+                                             search_response,
+                                             session):
+
+    quick_search_url = f'{TEST_URL}/quick-search'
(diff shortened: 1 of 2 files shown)


tests/integration/test_data_api.py
```diff
@@ -223,6 +224,43 @@ async def test_search_filter(item_descriptions,
     assert items_list == item_descriptions
 
 
+@respx.mock
+@pytest.mark.anyio
+async def test_search_filter_positional_args(item_descriptions,
```
asonnenschein, line 229: nit: add a short docstring for this test.

asonnenschein: Also, compared to the below tests it looks like this tests 'quick-search' and pagination rather than vanilla search?  If so what do you think about renaming this test to reflect quick-search + pagination?

asonnenschein (approved):
LGTM!  I left a couple of nitpick comment, up to you to take the feedback or not @stephenhillier.  Otherwise I don't see anything that should block this MR.

tests/integration/test_data_api.py
stephenhillier: Pagination removed from test and docstring added

Issue #1042, opened 2024-06-04
Add hosting block for Orders API


pl-gideon:
**Is your feature request related to a problem? Please describe.**

**Describe the solution you'd like**
Add hosting block for Orders API

**Describe alternatives you've considered**

**Additional context**

PR #1044, opened 2024-06-05
Release 2.8.0

pl-gideon:
2.8.0 (2024-06-05)

Added:
- Support for hosting parameter in Orders API (#1042)
- New Data API tests (#1041)

### DIFF
--- a/planet/__version__.py
+++ b/planet/__version__.py
@@ -1,1 +1,1 @@
-__version__ = '2.8'
+__version__ = '2.9dev'
--- a/planet/__version__.py
+++ b/planet/__version__.py
@@ -1,1 +1,1 @@
-__version__ = '2.8dev'
+__version__ = '2.8'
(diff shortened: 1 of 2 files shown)


planet/__version__.py
```diff
@@ -1 +1 @@
-__version__ = '2.8dev'
+__version__ = '2.9dev'
```
asonnenschein, line 1: Should be version `2.8` when you tag the release from this branch.  Then push another commit after you tag the release to update to `2.9dev` before merging to main (https://github.com/planetlabs/planet-client-python/blob/main/RELEASE.md#release-workflow).  IMO this release workflow is verbose and overcomplicated.  We'll have a simplified release workflow before the end of this half (hopefully).

asonnenschein (approved):

planet/__version__.py
asonnenschein: ☝️ disregard, I see you already did all of this!

PR #1049, opened 2024-07-03
Planet sync client

stephenhillier:
**Proposed Changes:**

Adds a `Planet` sync client that has feature parity with the existing async clients.

**Diff of User Interface**

New behavior:

The `Planet` client has members `data`, `subscriptions`, `orders`. Users can interact with these APIs without setting up individual clients or a Session.

Example:

```python
from planet import Planet

client = Planet()

for item in client.subscriptions.list_subscriptions():
    print(item)

results = client.data.search(["PSScene"], limit=10)
for item in results:
    print(item.get("id"))
```

Minor changes from the async clients to the sync client:
* for the Subscriptions client, instead of having both `get_results` and `get_results_csv` methods, the sync client has one method that optionally accepts `format="csv"` as a parameter. It defaults to the normal json output.

At the risk of scope creep, one additional nice-to-have for an initial MVP sync client would be some simple dataclass/attrs based return types instead of dicts. That way it's easier for users to understand what they're getting back in e.g. a search or order.


ischneider:
Nice work!

I'm definitely a big :+1: on the single Planet class.

I did a small test listing my searches and then concurrently executing each search using both sync and async (threadpool and queue/task approach) and found the performance to be essentially equivalent w/ both...

---

I'm less keen on duplication of all the documentation and boiler plate in testing, though I understand this is due to function "color".

One approach I just tested was using a metaclass to dynamically generate sync wrappers. The advantage of this is documentation is mirrored and there's much less code. One disadvantage is type hints aren't handled properly - the actual return types are correct but IDE doc shows up incorrectly... Would be happy to share this but I'm not satisfied with the lack of type hint propagation (and maybe this can be addressed?) but otherwise, beyond some trickery, it's simple and it works.

---

Another approach we might consider is top-level "request objects" [1] - this way the interface to the request (the name of the operation and the arguments) can always be sync and the results are where dispatch can vary. This way the docs for the operation parameters, exceptions, narrative etc are all in one spot and not repeated and the only code variation is essentially the `call_sync` (as it is now).

So `list_searches` might be implemented like

```
search = data.ListSearches(session, limit=500)
# sync
searches = search.get_data()
# async
searches = await search.aget_data()
```

Advantage here is parity w/ SH API.

Another approach might be to use a request object but require a specific client, e.g.

```
client = AsyncClient(session)
# list returns an async generator
searches = await client.list(data.ListSearches(limit=500))
```

The downside is we still have the old client classes, docs, etc. to maintain for some period of time...

[1] see `SentinelHubRequest` https://sentinelhub-py.readthedocs.io/en/latest/examples/process_request.html

ischneider:
Another approach - by extending DataClient and overriding every function (as sync and with altered type hints), we get the best of both (can inherit doc strings and get correct type hints) with less boilerplate...

vscode at least makes the overriding part simpler but then there's still maintenance (adding an optional parameter, for example) and the boilerplate of testing though at least w/ some creativity, I'd guess we could reduce this some...

stephenhillier:
I do like the idea of inheriting the docstrings where possible. If we can give ourselves a headstart on a simple MVP wrapper, with minimal duplication of what's already there, then that sounds good to me. 

The reason I chose to start with the explicit approach is because this is a standalone client and it shouldn't necessarily be a violation of DRY merely to reuse the descriptions (IMO). They will probably need to be modified independently of each other as changes are made. 

That being said I will see what I can come up with for sharing the docstrings, I should have some time to round off the week to try a couple approaches :+1:

stephenhillier:
It occurs to me while going down the path of creating shared docstrings, that we might be needlessly coupling the sync and async implementations together in order to save space on the docstrings. Yes each method calls out to the async method, but the implementation of any method can be changed (if needed), optional args added, and async or sync-specific examples can be added to the docstrings. Examples are sorely needed and those can't really be shared. 

The one-time port is simple and all the docs and type hints are correct (and the grunt work has been done). Thoughts?

planet/client.py
```diff
@@ -0,0 +1,52 @@
...
+    """
+
+    def __init__(self, session: Optional[Session] = None) -> None:
+        self._session = session or Session()
+        self._session._client.headers.update(
+            {"X-Planet-App": SYNC_CLIENT_AGENT})
+
+        self.data = DataAPI(self._session)
+        self.orders = OrdersAPI(self._session)
+        self.subscriptions = SubscriptionsAPI(self._session)
```
asonnenschein, line 52: Do all of these clients need to be instantiated at the same time?  I wonder if there is a way to 'opt-in' to the individual clients, e.g. a Subscriptions API user who only wants to use the Subscriptions client.  As coded, this user will absolutely be able to use the Subscriptions client, but they will also instantiate Data API and Orders API clients.

stephenhillier: I don't think there's much overhead for an end user to have these classes instantiated behind the scenes when the Planet client is created. They're really just providing methods and not holding any state. I think my question would be if there's any benefit to being able to opt-in/out of Data/Orders/Subscriptions on the client- there should be no harm in a user creating a Planet class and ignoring e.g. subscriptions until they need it (IMO).

docs/get-started/sync-client-quick-start.md
```diff
@@ -0,0 +1,237 @@
...
+The Planet SDK for Python makes it easy to access Planet’s massive repository of satellite imagery and add Planet
+data to your data ops workflow.
+
+**Note:** This is the new, non-asyncio client. If you want to take advantage of asyncio, check the [asyncio client quick start guide](quick-start-guide.md).
+
+Your feedback on this version of our client is appreciated. Please raise an issue on [GitHub](https://github.com/planetlabs/planet-client-python/issues) if you encounter any problems.
+
+## Dependencies
+
+This package requires [Python 3.8 or greater](https://python.org/downloads/). A virtual environment is strongly recommended.
```
asonnenschein, line 14: nit:  The SDK is pinned to Python v >=3.9 now https://github.com/planetlabs/planet-client-python/blob/main/pyproject.toml#L19

docs/get-started/sync-client-quick-start.md
```diff
@@ -0,0 +1,237 @@
...
+    pl.orders.wait(order['id'])
+
+    pl.orders.download_order(order['id'], overwrite=True)
+```
+
+### Creating a subscription
+
+#### Prerequisites
+
+Subscriptions will require a delivery destination. The following example uses Amazon S3.
```
asonnenschein, line 197: > Subscriptions will require a delivery destination.

Not always, subscriptions can be created without a delivery destination.  This would be a results only subscription.

docs/get-started/sync-client-quick-start.md
```diff
@@ -0,0 +1,237 @@
...
+You will need your ACCESS_KEY_ID, SECRET_ACCESS_KEY, bucket and region name.
+
+#### Scene subscription
+
+To subscribe to scenes that match a filter, use the `subscription_request` module to build a request, and
+pass it to the `subscriptions.create_subscription()` method of the client.
+
+Warning: the following code will create a subscription, consuming quota based on your plan.
```
asonnenschein, line 205: Good idea to add a warning here!

planet/clients/orders.py
```diff
@@ -502,3 +509,269 @@ async def list_orders(self,
...
+        Parameters:
+            order_id: The ID of the order
+
+        Returns:
+            JSON description of the order
+
+        Raises:
+            planet.exceptions.ClientError: If order_id is not a valid UUID.
+            planet.exceptions.APIError: On API error.
+        """
```
asonnenschein, line 571: This docstring is exactly the same as the docstring for it's async counterpart, `OrdersClient().get_order()`.  I wonder if we could simplify things by applying the docstring from `OrdersClient.get_order()` to this method so that the same docstring isn't defined in two places?  By 'apply' I mean something like 'share' the docstring between the two methods rather than 'inherit' because the implementation in this MR doesn't follow a standard Python inheritance pattern; the `OrdersClient` is passed to `OrdersAPI` via the `OrdersAPI._client` property.  Here's an example of what I am thinking:

```python
class OrdersClient:
    def get_order(self):
        """Docstring for get_order()."""
        pass

class Child:
    def get_order(self):
        self.get_order.__doc__ = self._client.get_order.__doc__
        # Add any additional docstring here
        pass
```

planet/http.py
```diff
@@ -272,6 +275,20 @@ async def alog_response(*args, **kwargs):
...
+        # create a dedicated event loop for this httpx session.
+        def _start_background_loop(loop):
+            asyncio.set_event_loop(loop)
+            loop.run_forever()
+
+        self._loop = asyncio.new_event_loop()
+        self._loop_thread = threading.Thread(target=_start_background_loop,
+                                             args=(self._loop, ),
+                                             daemon=True)
+        self._loop_thread.start()
```
asonnenschein, line 287: What does this do?

docs/get-started/sync-client-quick-start.md
asonnenschein: Resolving, I pushed a commit to update this to 3.9.

planet/http.py
stephenhillier: Since the methods of the sync client generally just call out to the async client methods, we need to use those async methods within an event loop. An easy way to do this is to invoke the methods using `asyncio.run` or `asyncio.get_running_loop`, however one issue with this approach is that the httpx session (which is coupled to our own Session class) is tied to a single event loop. If we close the event loop, our Session instance will no longer be usable.

Creating a new event loop in a thread is a way to ensure that the httpx session will have a persistent loop for as long as our sdk's Session is alive. This lets us manage the asyncio loop (and the Session instance) behind the scenes without the user needing to as they would with the async client.

planet/clients/orders.py
stephenhillier: I tend to think the negatives outweigh the positives here. Readers of the code would not have a docstring to refer to in the usual place and we also would not be able to use examples in shared docstrings since the methods are called differently.

planet/http.py
stephenhillier: Also, you won't hear me argue that this is an ideal long term solution. I think it is a good step in the right direction for users asking for an easier to use client.

If adoption of the sync client is promising, it may be worthwhile to transition the sync client to have its own non-async http session with no event loop in the picture. Something like @ischneider's idea of shared business logic and request setup, with either sync or async request dispatching (and my preference would be to abstract all that from users). We might be able to use the Features and/or Basemaps clients to test some ideas out.

asonnenschein: Thanks for explaining!  I understand why this is needed in the current state of the world, and agree that maybe in the future when we have less restraints for backwards compatibility that we could try another approach.  Otherwise  for now this is a clever solution that will help ship the sync client sooner rather than later!

planet/clients/orders.py
asonnenschein: Understood!  FWIW I don't think dynamic vs statically defined docstrings should block this MR.  Check out https://github.com/planetlabs/planet-client-python/pull/1077 where I implemented a dynamic approach.  If that dynamic approach passes review, we can merge to main and ship it as an enhancement at a later date.

tests/integration/test_data_api.py
```diff
@@ -70,6 +72,11 @@ def search_response(item_descriptions):
     return response
 
 
+@pytest.fixture(scope="module")
+def data_api():
+    return DataAPI(Session(), TEST_URL)
```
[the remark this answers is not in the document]
stephenhillier: Agreed, it makes sense to access the Data API methods the same way.

ischneider (approved):
Let's squash and merge this and move forward!

PR #1052, opened 2024-07-23
Add a redirect to the v2 docs to the v1 gh-pages site

tbarsballe:
Recently, we had some users landing on the old v1 docs rather than the v2 docs. As we do still have some users on v1, I think it is important to keep those docs around, but they should have a link to v2 so that new users don't get lost. This PR adds a deprecation notice and link to the top of the v1 docs:

Screenshot 2024-07-23 at 10 56 41 AM

The old v1 docs appear to be entirely static rather than the more typical Jekyll, so this change is done in JS, so it applies to all pages with minimal effort and duplication. Not optimal, but for a deprecated doc site that isn't going to change anytime soon I think its fine.

**Proposed Changes:**

1. Update the v1 gh-pages documentation with a deprecation notice and a link to the v2 documentation

**PR Checklist:**

### DIFF
--- a/_static/custom.css
+++ b/_static/custom.css
@@ -51,3 +51,23 @@ pre span {
 {
   color: #f2f2f2;
 }
+
+#redirect {
+  padding: 5px;
+  font-size: 18px;
+  text-align: center;
+  color: white;
+  background-color: rgb(0, 127, 153);
+  border-bottom: 3px solid ;
+
+}
+
+@media screen and (max-width: 875px) {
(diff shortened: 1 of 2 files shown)


tbarsballe:
@ischneider Yeah, that'd definitely be the cleaner solution. I poked around the repo and frankly didn't want to try untangling the 4+ year-old doc build situation, this was the simplest and fastest solution I could throw together to solve this while still looking half-decent.

PR #1060, opened 2024-09-30
subscriptions get_results_csv uses session.stream

### LINKED ISSUE #1059: the get_results_csv subscriptions operation does not throttle itself
**Expected behavior**

`get_results_csv` should use the client's session's `stream` function rather than the lower-level httpx client's `stream`

while the stream itself cannot be retried, this results in the initial request being throttled (and retried)

there is no throttling and basic unlimited usage will result in TooManyRequest errors

**Related Issues**

n/a

**Workaround**

manual throttling and/or retries

**Minimum, Complete, Viable Code Sample**

n/a

ischneider:
**Related Issue(s):**

Closes #1059

**Proposed Changes:**

1. get_results_csv throttled and retried

### DIFF
--- a/planet/clients/subscriptions.py
+++ b/planet/clients/subscriptions.py
@@ -336,7 +336,6 @@ class SubscriptionsClient:
         # during streaming. We may want to consider a retry decorator
         # for this entire method a la stamina:
         # https://github.com/hynek/stamina.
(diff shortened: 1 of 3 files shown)


stephenhillier (approved):
Is the `Note: retries are not implemented yet ...` comment still needed?

PR #1067, opened 2024-10-18
modernize with pyproject, use scm for version

ischneider:
**Related Issue(s):**

N/A

**Proposed Changes:**

Modernize project metadata, use scm for versioning

**Diff of User Interface**

N/A

**PR Checklist:**


planet/__version__.py
```diff
@@ -1 +1,3 @@
-__version__ = '2.11.0dev'
+from importlib.metadata import version
+
+__version__ = version("planet")
```
asonnenschein, line 3: Slick!  I haven't seen this before.  From the docs it looks like `version()` will pull the version number of the package as a string from the pypi distribution package?  Does that mean we don't have to bump the version number manually anymore and this will just pull whatever version we tag?

ischneider: Yes, this makes releasing much simpler!

asonnenschein (approved):
LGTM!  After this MR is merged I will follow up with an update to the release process that reflects the changes to __version__.py.  Before you merge could you update the Python version here https://github.com/planetlabs/planet-client-python/blob/main/CONTRIBUTING.md?plain=1#L174-L195 ?

ischneider:
Updated various references to 3.7 with recent versions... I stuck w/ 3.12 for the git actions as this is what ubuntu-latest provides? (don't think it matters which version, as long as it runs...)

PR #1071, opened 2024-10-24
Document and clean up subscriptions client limit behaviour

### LINKED ISSUE #1069: Limit behaviour not documented in the subscriptions API
**Expected behavior**
I would expect the [Subscriptions Client Reference](https://planet-sdk-for-python-v2.readthedocs.io/en/latest/python/sdk-reference/#planet.clients.subscriptions.SubscriptionsClient) to describe how to allow no limit (set limit=0), but it does not.
The data client documentation does describe this behaviour.

### LINKED ISSUE #1070: get_results_csv limit parameter has no effect
**Expected behavior**
As this request is returning a single document, I would not expect this parameter to be present

The limit parameter of [SubscriptionsClient().get_results_csv()](https://planet-sdk-for-python-v2.readthedocs.io/en/latest/python/sdk-reference/#planet.clients.subscriptions.SubscriptionsClient.get_results_csv) is not described in the documnetation.
More importantly, the parameter is [never actually used in the code](https://github.com/planetlabs/planet-client-python/blob/main/planet/clients/subscriptions.py#L331-L342)

While this is technically a backwards-incompatible change in the SDK, I think the cleanest solution would be to just remove the parameter

**Related Issues**
#1069

tbarsballe:
**Related Issue(s):**

Closes #1069 
Closes #1070 

**Proposed Changes:**

1. Removed the unused `limit` parameter of `SubscriptionsClient.get_results_csv`

Not intended for changelog:

1. Added documentation of the limit=0 behaviour to other SubscriptionsClient methods where this is still used(`list_subscriptions`, `get_results`)

**PR Checklist:**

### DIFF
--- a/planet/clients/subscriptions.py
+++ b/planet/clients/subscriptions.py
@@ -74,7 +74,7 @@ class SubscriptionsClient:
                 set, filter out subscriptions with status not in this
                 set.
             limit (int): limit the number of subscriptions in the
-                results.
+                results. When set to 0, no maximum is applied.
             TODO: user_id
 
         Yields:
@@ -270,7 +270,7 @@ class SubscriptionsClient:
             status (Set[str]): pass result with status in this set,
                 filter out results with status not in this set.
             limit (int): limit the number of subscriptions in the
-                results.
+                results. When set to 0, no maximum is applied.
             TODO: created, updated, completed, user_id
 
         Yields:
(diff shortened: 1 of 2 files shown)


planet/clients/subscriptions.py
```diff
@@ -310,8 +310,7 @@ async def get_results_csv(
                                           "queued",
                                           "processing",
                                           "failed",
-                                          "success"]]] = None,
-        limit: int = 100,
```
stephenhillier, line 314: We had a Slack discussion and found that this arg is unused, undocumented and was never required. This theoretically could be a breaking change for somebody, but removing it seems reasonable

stephenhillier (approved):

asonnenschein:
Confirming after reviewing this PR and Subscription API's /results endpoint that `limit` is undocumented and unused, and I agree that this should be removed.

asonnenschein (approved):

PR #1073, opened 2024-10-29
Features client

tbarsballe:
Includes: list_collections, create_collection, list_features, create_features

This came out of an internal need for a project I was working on incorporating the SDK, and represents about the extent of time I have to contribute to Features API support at this time. If someone else can pick up from where this started, that would be great.

I have programmatically and manually tested all of the the FeaturesClient methods in the context of the project described above, and everything seems to be working. In particular, the API has different pagination scheme, currently handled by overriding Paged.__next_link, that could perhaps be incorporated directly into Paged, since the as-of-yet-unimplemented Analytics API uses the same pagination scheme.

**Related Issue(s):**

Oddly enough, we do not seem to have an issue for Features API support yet.

**Proposed Changes:**

1. Add FeaturesClient to support the Features API

**PR Checklist:**

- [ ] This PR is as small and focused as possible
- [ ] If this PR includes proposed changes for inclusion in the changelog, the title of this PR summarizes those changes and is ready for inclusion in the Changelog.
- [ ] This PR does not break any examples or I have updated them

### DIFF
--- a/docs/python/async-sdk-guide.md
+++ b/docs/python/async-sdk-guide.md
@@ -458,6 +458,59 @@ async def download_and_validate():
         cl.validate_checksum(asset, path)
 ```
 
+### Features API Collections and Features
+
+The Python SDK now supports Features API Collections and Features (note: in the SDK and API, Features are often referred to as items in a collection).
+
+Collections and Features/items that you create in in the SDK will be visible in Features API and Features Manager.
+
+#### Creating a collection
+
+You can use the Python SDK to create Features API collections. 
+
+```python
+async with Session() as sess:
+  client = FeaturesClient(sess)
+  new_collection = await client.create_collection(title="my collection", description="a new collection")
+```
+
+#### Listing collections
+
+```python
+async with Session() as sess:
+  client = FeaturesClient(sess)
+  collections = client.list_collections()
+  async for collection in collections:
+    print(collection)
+```
+
+#### Listing features/items in a collection
+
+```python
+async with Session() as sess:
+  client = FeaturesClient(sess)
+  items = client.list_items(collection_id)
+  async for item in items:
+    print(items)
+```
+
+#### Using items as geometries for other methods
+
+You can pass collection items/features directly to other SDK methods. Any method that requires a geometry will accept
+a Features API Feature.
+
+```python
+async with Session() as sess:
+  client = FeaturesClient(sess)
+  items = client.list_items(collection_id)
+  example_feature = await anext(items)
+
+  data_client = DataClient(sess)
+  results = data_client.search(["PSScene"], geometry=example_feature, limit=1)
+  async for result in results:
+    print(result)
+```
+
 ## API Exceptions
 
 When errors occur, the Planet SDK for Python exception hierarchy is as follows:
--- a/planet/cli/cmds.py
+++ b/planet/cli/cmds.py
@@ -14,6 +14,7 @@
 """Decorators for Click commands"""
 import asyncio
 from functools import wraps
+from typing import Optional
 
 import click
 
@@ -21,7 +22,7 @@ from planet import exceptions
 from planet.cli.options import pretty
 
 
-def command(group: click.Group):
+def command(group: click.Group, name: Optional[str]=None):
     """a decorator that adds common utilities/options to a click command
 
     usage:
@@ -78,7 +79,7 @@ def command(group: click.Group):
         # Doing this last (outside the wrapper) allows click to
         # pick up the options/arguments added to the command by e.g.
         # `@click.option()`
-        return group.command(wrapper)
+        return group.command(name=name)(wrapper)
 
     return decorator
 
--- a/planet/cli/cmds.py
+++ b/planet/cli/cmds.py
@@ -47,13 +47,8 @@ def command(group: click.Group,
     # the decorators to add to the function **when the function is run**
     # (as opposed to when the function is registered as a click command)
     decorators = [
-        translate_exceptions,
         coro,
-    ]
-
-    # click specific functionality e.g. any decorators that
-    # register an extra argument for the function
-    click_fns = [
+        translate_exceptions,
         click.pass_context,
         pretty,
     ] + extra_args
@@ -64,7 +59,7 @@ def command(group: click.Group,
     def decorator(f):
 
         # run any click-specific registration decorators
-        for fn in click_fns:
+        for fn in decorators:
             f = fn(f)
 
         @wraps(f)
--- a/planet/cli/cmds.py
+++ b/planet/cli/cmds.py
@@ -14,7 +14,7 @@
 """Decorators for Click commands"""
 import asyncio
 from functools import wraps
-from typing import Optional
+from typing import Callable, Optional
 
 import click
 
@@ -22,7 +22,9 @@ from planet import exceptions
 from planet.cli.options import pretty
 
 
-def command(group: click.Group, name: Optional[str] = None):
+def command(group: click.Group,
+            name: Optional[str] = None,
+            extra_args: list[Callable] = []):
     """a decorator that adds common utilities/options to a click command
 
     usage:
@@ -54,12 +56,16 @@ def command(group: click.Group, name: Optional[str] = None):
     click_fns = [
         click.pass_context,
         pretty,
-    ]
+    ] + extra_args
 
     # since we want to use `command` as a function with an arg: `@command(group)`,
     # we need to create and return an "real" decorator that takes the function as its
     # arg.
     def decorator(f):
+
+        # run any click-specific registration decorators
+        for fn in click_fns:
+            f = fn(f)
 
         @wraps(f)
         def wrapper(*args, **kwargs):
@@ -70,10 +76,6 @@ def command(group: click.Group, name: Optional[str] = None):
                 cmd = d(f)
 
             return cmd(*args, **kwargs)
-
-        # run any click-specific registration decorators
-        for fn in click_fns:
-            f = fn(f)
 
         # register the whole thing as a Click command.
         # Doing this last (outside the wrapper) allows click to
--- a/planet/cli/cmds.py
+++ b/planet/cli/cmds.py
@@ -22,7 +22,7 @@ from planet import exceptions
 from planet.cli.options import pretty
 
 
-def command(group: click.Group, name: Optional[str]=None):
+def command(group: click.Group, name: Optional[str] = None):
     """a decorator that adds common utilities/options to a click command
 
     usage:
--- a/planet/cli/cmds.py
+++ b/planet/cli/cmds.py
@@ -18,6 +18,69 @@ from functools import wraps
 import click
 
 from planet import exceptions
+from planet.cli.options import pretty
+
+
+def command(group: click.Group):
+    """a decorator that adds common utilities/options to a click command
+
+    usage:
+
+    @command(features)  # pass a group created with @click.group
+    def my_command():
+        ...
+
+    this single decorator replaces a list of decorators we would otherwise
+    add to every function e.g.:
+
+    @features.command
+    @coro
+    @translate_exceptions
+    @click.pass_context
+    def my_command():
+        ...
+    """
+
+    # the decorators to add to the function **when the function is run**
+    # (as opposed to when the function is registered as a click command)
+    decorators = [
+        translate_exceptions,
+        coro,
+    ]
+
+    # click specific functionality e.g. any decorators that
+    # register an extra argument for the function
+    click_fns = [
+        click.pass_context,
+        pretty,
+    ]
+
+    # since we want to use `command` as a function with an arg: `@command(group)`,
+    # we need to create and return an "real" decorator that takes the function as its
+    # arg.
+    def decorator(f):
+
+        @wraps(f)
+        def wrapper(*args, **kwargs):
+            cmd = f
+
+            # wrap cmd with all the default decorators
+            for d in decorators:
+                cmd = d(f)
+
+            return cmd(*args, **kwargs)
+
+        # run any click-specific registration decorators
+        for fn in click_fns:
+            f = fn(f)
+
+        # register the whole thing as a Click command.
+        # Doing this last (outside the wrapper) allows click to
+        # pick up the options/arguments added to the command by e.g.
+        # `@click.option()`
+        return group.command(wrapper)
+
+    return decorator
 
 
 # https://github.com/pallets/click/issues/85#issuecomment-503464628
@@ -60,6 +123,6 @@ def translate_exceptions(func):
                 'Auth information does not exist or is corrupted. Initialize '
                 'with `planet auth init`.')
         except exceptions.PlanetError as ex:
-            raise click.ClickException(ex)
+            raise click.ClickException(str(ex))
 
     return wrapper
--- a/planet/cli/features.py
+++ b/planet/cli/features.py
@@ -7,7 +7,7 @@ from planet.cli.io import echo_json
 from planet.clients.features import FeaturesClient
 
 from .cmds import coro, translate_exceptions
-from .options import pretty
+from .options import pretty, compact
 from .session import CliSession
 
 
@@ -63,7 +63,8 @@ async def collection_create(ctx, title, description, pretty):
 @translate_exceptions
 @coro
 @pretty
-async def collections_list(ctx, pretty):
+@compact
+async def collections_list(ctx, pretty, compact):
     """List Features API collections
 
     Example:
@@ -72,7 +73,17 @@ async def collections_list(ctx, pretty):
     """
     async with features_client(ctx) as cl:
         results = cl.list_collections()
-        echo_json([c async for c in results], pretty)
+
+        if compact:
+            compact_fields = ('id', 'title', 'description')
+            output = [{
+                k: v
+                for k, v in row.items() if k in compact_fields
+            } async for row in results]
+        else:
+            output = [c async for c in results]
+
+        echo_json(output, pretty)
 
 
 @features.command()  # type: ignore
--- a/planet/cli/features.py
+++ b/planet/cli/features.py
@@ -1,0 +1,137 @@
+from contextlib import asynccontextmanager
+import json
+import click
+from click.exceptions import ClickException
+
+from planet.cli.io import echo_json
+from planet.clients.features import FeaturesClient
+
+from .cmds import coro, translate_exceptions
+from .options import pretty
+from .session import CliSession
+
+
+@asynccontextmanager
+async def features_client(ctx):
+    async with CliSession() as sess:
+        cl = FeaturesClient(sess, base_url=ctx.obj['BASE_URL'])
+        yield cl
+
+
+@click.group()  # type: ignore
+@click.pass_context
+@click.option('-u',
+              '--base-url',
+              default=None,
+              help='Assign custom base Features API URL.')
+def features(ctx, base_url):
+    """Commands for interacting with Features API"""
+    ctx.obj['BASE_URL'] = base_url
+
+
+@features.command()  # type: ignore
+@click.pass_context
+@translate_exceptions
+@coro
+@click.option("-t",
+              "--title",
+              required=True,
+              help="a title for the collection")
+@click.option("-d",
+              "--description",
+              "--desc",
+              required=False,
+              help="a description for the collection")
+@pretty
+async def collection_create(ctx, title, description, pretty):
+    """Create a new Features API collection.
+
+    Example:
+
+    \b
+    planet features collection-create \\
+      --title "new collection" \\
+      --desc "my new collection"
+    """
+    async with features_client(ctx) as cl:
+        col = await cl.create_collection(title, description)
+        echo_json(col, pretty)
+
+
+@features.command()  # type: ignore
+@click.pass_context
+@translate_exceptions
+@coro
+@pretty
+async def collection_list(ctx, pretty):
+    """List Features API collections
+
+    Example:
+
+    planet features collection-list
+    """
+    async with features_client(ctx) as cl:
+        results = cl.list_collections()
+        echo_json([c async for c in results], pretty)
+
+
+@features.command()  # type: ignore
+@click.pass_context
+@translate_exceptions
+@coro
+@click.argument("collection_id", required=True)
+@pretty
+async def collection_get(ctx, collection_id, pretty):
+    """List Features API collections
+
+    Example:
+
+    planet features collection-list
+    """
+    async with features_client(ctx) as cl:
+        result = await cl.get_collection(collection_id)
+        echo_json(result, pretty)
+
+
+@features.command()  # type: ignore
+@click.pass_context
+@translate_exceptions
+@coro
+@click.argument("collection_id", required=True)
+@pretty
+async def feature_list(ctx, collection_id, pretty):
+    """List features in a Features API collection
+
+    Example:
+
+    planet features feature-list my-collection-123
+    """
+    async with features_client(ctx) as cl:
+        results = cl.list_features(collection_id)
+        echo_json([f async for f in results], pretty)
+
+
+@features.command()  # type: ignore
+@click.pass_context
+@translate_exceptions
+@coro
+@click.argument("collection_id", required=True)
+@click.argument("filename", required=True)
+@pretty
+async def feature_add(ctx, collection_id, filename, pretty):
+    """Add features from a geojson file to a Features API collection
+
+    Example:
+
+    planet features feature-add my-collection-123 ./my_geom.geojson
+    """
+    async with features_client(ctx) as cl:
+        with open(filename) as data:
+            try:
+                res = await cl.add_features(collection_id, json.load(data))
+            except json.decoder.JSONDecodeError:
+                raise ClickException(
+                    "Only JSON (.json, .geojson) files are supported in the CLI. Please use https://planet.com/features to upload other files."
+                )
+
+    echo_json(res, pretty)
--- a/planet/cli/features.py
+++ b/planet/cli/features.py
@@ -5,6 +5,7 @@ from click.exceptions import ClickException
 
 from planet.cli.io import echo_json
 from planet.clients.features import FeaturesClient
+from planet.geojson import split_ref
 
 from .cmds import coro, translate_exceptions
 from .options import pretty, compact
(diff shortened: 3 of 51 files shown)


planet/clients/features.py
```diff
@@ -0,0 +1,183 @@
...
+
+        class _FeaturesPager(Paged):
+            """Navigates pages of messages about features."""
+            ITEMS_KEY = 'features'
+            LINKS_KEY = 'links'
+
+            def _next_link(self, page):
+                next_link = False
+
+                #TODO: Build this into Paged.__next_link directly, other (unimplemented) APIs have a similar page structure
```
stephenhillier, line 108:
```suggestion
                # TODO: Build this into Paged.__next_link directly, other (unimplemented) APIs have a similar page structure
```

planet/clients/features.py
```diff
@@ -0,0 +1,183 @@
...
+                for link in page[self.LINKS_KEY]:
+                    if "rel" in link and link["rel"] == self.NEXT_KEY:
+                        next_link = link["href"]
+                        LOGGER.debug(f'next: {next_link}')
+                if not next_link:
+                    LOGGER.debug('end of the pages')
+                return next_link
+
+        params = {}
```
stephenhillier, line 117:
```suggestion
        params: list[str, Any] = {}
```

This is failing the mypy check since we're adding multiple types into this dict.

planet/clients/features.py
```diff
@@ -0,0 +1,183 @@
...
+    async def add_features(
+            self,
+            collection_id,
+            request: dict,
+            bbox: Optional[list[float]] = None,
+            datetime: Optional[str] = None,
+            property_id: Optional[str] = None) -> AsyncIterator[str]:
+
+        url = f'{self._base_url}/collections/{collection_id}/items'
+        params = {}
```
stephenhillier, line 164:
```suggestion
        params: dict[str, Any] = {}
```

planet/clients/features.py
```diff
@@ -0,0 +1,183 @@
+"""Planet Features API Python client."""
+
+import logging
+from typing import AsyncIterator, Optional, Iterator
```
stephenhillier, line 4:
```suggestion
from typing import Any, AsyncIterator, Optional
```

stephenhillier:
I added some minor suggestions for the type errors. Just remembered your request for somebody picking this up so no need to follow up.

Looks like the biggest piece remaining is some test coverage. Maybe one of us can tackle that during one of the working session time blocks.

tbarsballe:
Thanks for taking a look. Test coverage is definitely a big missing piece. 

CLI integration would be the other major part that's lacking, since this realy only covers the SDK part right now.

There's also a number of other Features API endpoints to implement - so far this covers all the major ones, but not any of the less-used ones (e.g. "alternates" or "validate")

stephenhillier:
I'm spending some time finishing this up, current progress:
* updated methods, fixed up some arguments 
* added tests for all methods
  * Trying a new thing where we have one test for each piece of functionality, which calls both sync and async methods (uses parameterize if we need multiple test cases). @ischneider 
* added sync client
* added docstrings with examples to all methods (sync and async)

I'm also proposing to add a new class Feature, which is a subclass of `dict`. `Feature` has a `.ref` convenience method/property which makes it easier to grab the feature reference (required for using features with Data API, Subscriptions API etc.). With this we can allow users to pass Features directly to other methods e.g. `pl.data.search(["PSScene"], geometry=feature)`. Subclassing `dict` makes it easier to extend this pattern to other object types in the future, since most client methods return dicts right now.

Another proposal is that add_features accepts geojson Features, Geometries as well as objects that implement `__geo_interface__`. This means users can import features directly from other spatial libraries.

stephenhillier:
Added CLI commands with tests

tbarsballe:
I did a little bit of testing of the CLI here (mostly because I happened to need to list some features today), and have a little bit of feedback on the user experience:

- The command `planet features feature-list [collection-id]` seems overly verbose / redundant due to the "features feature" part (though it makes sense from an API structure perspective). Also since the base command (features) is plural, I found myself mistyping feature-list as features-list frequently (same for collection-list/collections-list). 
  
  Perhaps switch to just using the url slug to fix both issues? i.e. `collection-list` -> `collections-list`, `feature-list` -> `items-list`, etc.
- The help message for the collection-list and collection-get commands are identical - this is likely to be confusing for users: 
  ```
  Commands:
  collection-create  Create a new Features API collection.
  collection-get     List Features API collections
  collection-list    List Features API collections
  feature-add        Add features from a geojson file to a Features API...
  feature-list       List features in a Features API collection
  ```
- This last one probably isn't worth solving right now, but extracting a collection id from the output of collection-list to use in feature-list is a little unwieldy, especially if you've got a lot of collections - this is basically the same problem we've got going from search results to orders, and the same solution of an --ids-only option, or maybe a -h (human-readable) option that also included names, would probably solve it.

stephenhillier:
Thanks @tbarsballe! 

I had the same issue as you- I regularly typed `collections-list` instead of of `collection-list` while working on this and testing. I was trying to follow the pattern that `planet data search-list` used, but perhaps we should just go with what feels most natural. Does `collections-list` sound good (and the same deal for the other actions)?

For your 3rd point, is the use case to list collections, then pick a collection and list those features? `--pretty` could really help there; would a `--compact` option help as well to just show title/description/id? Or would you prefer just formatted text output?

tbarsballe:
> Does collections-list sound good (and the same deal for the other actions)?

Yeah, `collections-list` sounds good to me (and maybe `items-list` instead of features list to avoid the double "features", and match the API endpoint)

> For your 3rd point, is the use case to list collections, then pick a collection and list those features? --pretty could really help there; would a --compact option help as well to just show title/description/id? Or would you prefer just formatted text output?

Yes, that's exactly the use case. I think a --compact option like you describe would most cleanly solve it.

stephenhillier:
Here are some options for the CLI:

For listing items, we could go with either of the following (both are equivalent):
```
planet features items-list  <collection_id>
planet features collection-items <collection_id>
```

For creating an item/feature, I think we would stick with item-add unless you have another idea:
```
planet features item-add <collection_id> feature.geojson
```

Alternately we could use another layer of sub-commands to represent collections and items/features:
```
planet features collections list
planet features collections create
planet features items list <collection_id>
planet features items add
```

Something like this is also possible, but may make it harder to present the CLI in terms of logical groups in the docs:
```
planet features collections list
planet features collections create
planet features list
planet features add
```

I'm not sure about an extra layer sub-commands myself, I personally would rather have one flat list of 6-7 commands in `planet features --help`  than two extra sub-commands to figure out. But open to opinions.

tbarsballe:
I like your first option. And yes, sticking `item-add` seems fine to me.

I had also considered another layer of commands, but it just seems like added complexity for not much gain.

stephenhillier:
Added `--compact` mode to `planet features collections-list` and updated some of the CLI command names.

Output of `--help` showing available commands:

```
> planet features --help
Usage: planet features [OPTIONS] COMMAND [ARGS]...

  Commands for interacting with Features API

Options:
  -u, --base-url TEXT  Assign custom base Features API URL.
  --help               Show this message and exit.

Commands:
  collection-create  Create a new Features API collection.
  collection-get     Get a collection by ID
  collections-list   List Features API collections
  item-add           Add features from a geojson file to a collection
  item-get           Get a feature in a collection.
  items-list         List features in a Features API collection
```

stephenhillier:
Also, I figure we should probably update the SDK methods to say `get_item` instead of `get_feature` for consistency... if everyone agrees I'll push another commit. Otherwise this should be it for the initial Features API SDK release.

tbarsballe:
I tried out the compact mode - looks good, much more legible.
Changing `get_feature` to `get_item` seems sensible to me.

planet/sync/data.py
```diff
@@ -47,7 +49,7 @@ def search(
         name: Optional[str] = None,
         sort: Optional[str] = None,
         limit: int = 100,
-        geometry: Optional[Union[Dict, str]] = None,
+        geometry: Optional[Union[Dict, str, Feature]] = None,
```
ischneider, line 52: I like the idea of being polymorphic and accepting all of these and I think we should expand/formalize this with a top-level type alias https://docs.python.org/3/library/typing.html#type-aliases (and we can hang more doc on this, too, and not repeat it)

Can be a bigger discussion as part of a later change, but as to what to call that type is a harder question, since calling it geometry feels limiting (like by excluding Feature which "has a" geometry) but maybe that's just me

tests/integration/test_features_api.py
```diff
@@ -0,0 +1,276 @@
...
+
+def mock_response(url: str, json: Any, method: str = 'get'):
+    mock_resp = httpx.Response(HTTPStatus.OK, json=json)
+    respx.request(method, url).return_value = mock_resp
+
+
+def to_collection_model(collection: dict) -> dict:
+    """create a collection model as it would appear in a Features API response"""
+    id = collection.get("id")
+    return {
```
ischneider, line 64: less on you and more on our linter config but I find this to be unreadable... we should figure this out and reformat everything to avoid conflict/churn

tests/integration/test_features_api.py
```diff
@@ -0,0 +1,276 @@
...
+    collections_url = f'{TEST_URL}/collections'
+    mock_response(
+        collections_url,
+        list_collections_response([TEST_COLLECTION_1, TEST_COLLECTION_2]))
+
+    def assertf(resp):
+        assert resp[0]["id"] == "collection1"
+        assert resp[1]["id"] == "collection2"
+
+    cl_async = FeaturesClient(session, base_url=TEST_URL)
```
ischneider, line 167: while it might feel right to have the session be a fixture, I'd argue that tear-down is not useful in reality, as all it is basically a no-op when using respx (there's no real connection to close as it's mocked)

so, unless there is variation in the session (no), or base_url (no?), then both clients can/should be module-level non-fixture plain old python objects.

planet/cli/features.py
```diff
@@ -0,0 +1,184 @@
...
+@click.option('-u',
+              '--base-url',
+              default=None,
+              help='Assign custom base Features API URL.')
+def features(ctx, base_url):
+    """Commands for interacting with Features API"""
+    ctx.obj['BASE_URL'] = base_url
+
+
+@features.command()  # type: ignore
```
ischneider, line 33: not critical, but if we're applying the same annotations (command, pass_context, coro, etc.), these can be unified using a single convenience def, something like (where sequence_of_annotations is the "stack" of common ones)

```
def features_command(f):
  for ann in sequence_of_common_annotations:
    f  = ann(f)
  return f
```

then each command is reduced to the set of unique annotations, making it simpler to read...

planet/sync/data.py
stephenhillier: How about something like `GeometryLike` , `GeojsonLike`, or `GeojsonOrFeatureRef`?

stephenhillier: I went with GeojsonLike but added a docstring to explain what that means

tests/integration/test_features_api.py
stephenhillier: updated

asonnenschein:
re: separating vs hyphenating nouns and verbs in CLI commands (e.g. `collection-create` vs `collection create`) -- 

I prefer separating the nouns and verbs vs hyphenating because I think it makes the commands more scalable.  We recently shipped an MR that separates the nouns and verbs so we are already using this pattern in the SDK https://github.com/planetlabs/planet-client-python/pull/1094.

stephenhillier:
I split everything out into `planet features collections` and `planet features items`

I wasn't able to get the flat command list `--help` output like I hoped (click does have `MultiCommand` but they don't seem to fit well into in the middle of a chain of groups), but to be honest it seems like a relatively minimal issue and we can always adjust the help text later.

```
✦ ❯ planet features collections
Usage: planet features collections [OPTIONS] COMMAND [ARGS]...

  commands for interacting with Features API collections

Options:
  --help  Show this message and exit.

Commands:
  create  Create a new Features API collection.
  get     Get a collection by ID
  list    List Features API collections

✦ ❯ planet features items
Usage: planet features items [OPTIONS] COMMAND [ARGS]...

  commands for interacting with Features API items (features within a
  collection)

Options:
  --help  Show this message and exit.

Commands:
  add   Add features from a geojson file to a collection
  get   Get a feature in a collection.
  list  List features in a Features API collection
```

planet/sync/features.py
```diff
@@ -0,0 +1,200 @@
...
+        collection_id = pl.features.create_collection(
+          title="My Collection",
+          description="a test collection"
+        )
+        ```
+        """
+        collection = self._client.create_collection(title, description)
+        return self._client._call_sync(collection)
+
+    def list_items(self, collection_id: str) -> Iterator[Feature]:
```
tbarsballe, line 86: I was using the sync features client today, and discovered this needs to expose `limit`.
The underlying list FeaturesClient defaults to `limit=10`, so it is currently impossible to list more than 10 items using the sync client

I presume list_collections is similar

stephenhillier: added in 70e411b

planet/cli/features.py
stephenhillier: done!

planet/sync/features.py
```diff
@@ -0,0 +1,200 @@
...
+
+        When the feature is added to the collection, it will be given an ID.
+        The ID can be overriden by providing an `id` in the feature's properties.
+        Title and description can also be set this way using the `title` and
+        `description` properties.
+
+        Example:
+
+        ```
+        feature = {
```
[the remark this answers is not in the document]
stephenhillier: removed

stephenhillier:
@tbarsballe @asonnenschein @ischneider anyone able to take another look? I believe all comments should now be addressed but if I missed one let me know.

tbarsballe (commented):
Pulled it down for a test today and ran into a new issue - if you're using an expired API key, the CLI will dump a large stacktrace on you:
```
(.venv) torben.barsballe@torben planet-client-python % planet features collections list
  File "/Users/torben.barsballe/repos/Planet/planet-client-python/planet/http.py", line 296, in _raise_for_status
    response.raise_for_status()
    ~~~~~~~~~~~~~~~~~~~~~~~~~^^
  File "/Users/torben.barsballe/repos/Planet/planet-client-python/.venv/lib/python3.13/site-packages/httpx/_models.py", line 829, in raise_for_status
    raise HTTPStatusError(message, request=request, response=self)
httpx.HTTPStatusError: Client error '401 Unauthorized' for url 'https://api.planet.com/features/v1/ogc/my/collections'
For more information check: https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/401

During handling of the above exception, another exception occurred:

  File "/Users/torben.barsballe/repos/Planet/planet-client-python/.venv/bin/planet", line 8, in <module>
    sys.exit(main())
             ~~~~^^
  File "/Users/torben.barsballe/repos/Planet/planet-client-python/.venv/lib/python3.13/site-packages/click/core.py", line 1157, in __call__
    return self.main(*args, **kwargs)
           ~~~~~~~~~^^^^^^^^^^^^^^^^^
  File "/Users/torben.barsballe/repos/Planet/planet-client-python/.venv/lib/python3.13/site-packages/click/core.py", line 1078, in main
    rv = self.invoke(ctx)
  File "/Users/torben.barsballe/repos/Planet/planet-client-python/.venv/lib/python3.13/site-packages/click/core.py", line 1688, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
                           ~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^
  File "/Users/torben.barsballe/repos/Planet/planet-client-python/.venv/lib/python3.13/site-packages/click/core.py", line 1688, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
                           ~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^
  File "/Users/torben.barsballe/repos/Planet/planet-client-python/.venv/lib/python3.13/site-packages/click/core.py", line 1688, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
                           ~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^
  File "/Users/torben.barsballe/repos/Planet/planet-client-python/.venv/lib/python3.13/site-packages/click/core.py", line 1434, in invoke
    return ctx.invoke(self.callback, **ctx.params)
           ~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/torben.barsballe/repos/Planet/planet-client-python/.venv/lib/python3.13/site-packages/click/core.py", line 783, in invoke
    return __callback(*args, **kwargs)
  File "/Users/torben.barsballe/repos/Planet/planet-client-python/planet/cli/cmds.py", line 78, in wrapper
    return cmd(*args, **kwargs)
  File "/Users/torben.barsballe/repos/Planet/planet-client-python/planet/cli/cmds.py", line 102, in wrapper
    return asyncio.run(func(*args, **kwargs))
           ~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^
  File "/Library/Frameworks/Python.framework/Versions/3.13/lib/python3.13/asyncio/runners.py", line 194, in run
    return runner.run(main)
           ~~~~~~~~~~^^^^^^
  File "/Library/Frameworks/Python.framework/Versions/3.13/lib/python3.13/asyncio/runners.py", line 118, in run
    return self._loop.run_until_complete(task)
           ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^
  File "/Library/Frameworks/Python.framework/Versions/3.13/lib/python3.13/asyncio/base_events.py", line 721, in run_until_complete
    return future.result()
           ~~~~~~~~~~~~~^^
  File "/Users/torben.barsballe/repos/Planet/planet-client-python/planet/cli/features.py", line 82, in collections_list
    output = [c async for c in results]
             ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/torben.barsballe/repos/Planet/planet-client-python/planet/clients/features.py", line 100, in list_collections
    response = await self._session.request(method='GET', url=url)
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/torben.barsballe/repos/Planet/planet-client-python/planet/http.py", line 404, in request
    http_response = await self._retry(self._send, request, stream=False)
                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/torben.barsballe/repos/Planet/planet-client-python/planet/http.py", line 347, in _retry
    raise e
  File "/Users/torben.barsballe/repos/Planet/planet-client-python/planet/http.py", line 332, in _retry
    resp = await func(*a, **kw)
           ^^^^^^^^^^^^^^^^^^^^
  File "/Users/torben.barsballe/repos/Planet/planet-client-python/planet/http.py", line 410, in _send
    http_resp = await self._client.send(request, stream=stream)
                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/torben.barsballe/repos/Planet/planet-client-python/.venv/lib/python3.13/site-packages/httpx/_client.py", line 1629, in send
    response = await self._send_handling_auth(
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    ...<4 lines>...
    )
    ^
  File "/Users/torben.barsballe/repos/Planet/planet-client-python/.venv/lib/python3.13/site-packages/httpx/_client.py", line 1657, in _send_handling_auth
    response = await self._send_handling_redirects(
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    ...<3 lines>...
    )
    ^
  File "/Users/torben.barsballe/repos/Planet/planet-client-python/.venv/lib/python3.13/site-packages/httpx/_client.py", line 1715, in _send_handling_redirects
    raise exc
  File "/Users/torben.barsballe/repos/Planet/planet-client-python/.venv/lib/python3.13/site-packages/httpx/_client.py", line 1697, in _send_handling_redirects
    await hook(response)
  File "/Users/torben.barsballe/repos/Planet/planet-client-python/planet/http.py", line 299, in _raise_for_status
    cls._convert_and_raise(e)
    ~~~~~~~~~~~~~~~~~~~~~~^^^
  File "/Users/torben.barsballe/repos/Planet/planet-client-python/planet/http.py", line 98, in _convert_and_raise
    raise error_type(response.text)
planet.exceptions.InvalidAPIKey: {"detail":"Invalid username/password."}
```

Other endpoints don't have this issue, so I assume there's some error handing missing from the CLI part.

Aside from that, everything seems to be working fine.

I also skimmed over the last batch of changes - looks good to me, though I did add a minor comment on the doc content.

docs/python/async-sdk-guide.md
```diff
@@ -458,6 +458,59 @@ async def download_and_validate():
         cl.validate_checksum(asset, path)
 ```
 
+### Features API Collections and Features
+
+The Python SDK now supports Features API Collections and Features (note: in the SDK and API, Features are often referred to as items in a collection).
```
asonnenschein, line 463: Should we add a link to the platform docs on Features API?

docs/python/async-sdk-guide.md
```diff
@@ -458,6 +458,59 @@ async def download_and_validate():
...
+
+Collections and Features/items that you create in in the SDK will be visible in Features API and Features Manager.
+
+#### Creating a collection
+
+You can use the Python SDK to create Features API collections.
```
asonnenschein, line 469: nit: "You can use the Python SDK to create feature collections in the Features API."

docs/python/async-sdk-guide.md
```diff
@@ -458,6 +458,59 @@ async def download_and_validate():
...
+  client = FeaturesClient(sess)
+  items = client.list_items(collection_id)
+  async for item in items:
+    print(items)
+```
+
+#### Using items as geometries for other methods
+
+You can pass collection items/features directly to other SDK methods. Any method that requires a geometry will accept
+a Features API Feature.
```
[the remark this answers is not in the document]
asonnenschein: +1 and can we make it clear that the `collection_id` variable here refers to a feature collection ref?

docs/python/async-sdk-guide.md
```diff
@@ -458,6 +458,59 @@ async def download_and_validate():
...
+The Python SDK now supports Features API Collections and Features (note: in the SDK and API, Features are often referred to as items in a collection).
+
+Collections and Features/items that you create in in the SDK will be visible in Features API and Features Manager.
+
+#### Creating a collection
+
+You can use the Python SDK to create Features API collections. 
+
+```python
+async with Session() as sess:
```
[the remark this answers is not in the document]
asonnenschein: disregard, I found the sync snippets!

planet/cli/cmds.py
```diff
@@ -14,10 +14,76 @@
...
 from functools import wraps
+from typing import Callable, Optional
 
 import click
 
 from planet import exceptions
+from planet.cli.options import pretty
+
+
+def command(group: click.Group,
```
asonnenschein, line 25: IMO this is a clever solution to the 'too many decorators' problem, however this shouldn't block shipping the Features API client.  This can be it's own task that ships as a fast follow enhancement after we ship the Features API client.

planet/cli/features.py
```diff
@@ -0,0 +1,172 @@
...
+      --title "new collection" \\
+      --desc "my new collection"
+    """
+    async with features_client(ctx) as cl:
+        col = await cl.create_collection(title, description)
+        echo_json(col, pretty)
+
+
+@command(collections, name="list", extra_args=[limit, compact])
+async def collections_list(ctx, pretty, limit, compact):
```
asonnenschein, line 65: Should 'compact' have a default value?

planet/cli/features.py
```diff
@@ -0,0 +1,172 @@
...
+
+@command(items, name="add")
+@click.argument("collection_id", required=True)
+@click.argument("filename", required=True)
+async def item_add(ctx, collection_id, filename, pretty):
+    """Add features from a geojson file to a collection
+
+    Example:
+
+    planet features items add my-collection-123 ./my_geom.geojson
```
asonnenschein, line 161: Is there support, now or planned, to give the GeoJSON here as a JSON string?

planet/clients/features.py
```diff
@@ -0,0 +1,264 @@
...
+        feature.
+        """
+        if isinstance(feature, GeoInterface):
+            # we expect __geo_interface__ to return a Geometry (not a Feature), so
+            # we're using the name `feature` liberally. We'll convert to a feature
+            # at the next step.
+            feature = feature.__geo_interface__
+
+        # convert a geojson geometry into geojson feature
+        if feature.get("type", "").lower() in [
```
asonnenschein, line 245: Should the feature type be restricted to polygons and multi-polygons?  Why would a user send us a point or line geometry?

asonnenschein (approved):
LGTM!  I left a few comments, and I see there are some other open comments, but nothing that I think should block this MR.  The scope of this MR is already pretty huge, I am happy to see this merged as is and some of the lingering feedback can be fast follows.

planet/cli/features.py
stephenhillier: since it's a flag controlled by click I believe it is always passed and uses `False` as the default. This is the CLI command version of the function btw.

planet/cli/features.py
stephenhillier: do you mean like the following:

```
planet features items add my-collection-123 '{"type": "Feature, "geometry": {...}}'
```

That's a good idea and I hadn't thought about that.  If we get feedback that a user would like to do this, we could try to test whether the argument is a filename or valid json. I would probably wait to see if it's a common thing though.

planet/clients/features.py
stephenhillier: Hmm good question. You can search Data API with points and linestrings (and users do), but Features API currently does not accept them.

What do you think about just removing this list? You can submit geojson Geometries as-is, I don't know why I bothered converting them to Features here. You get an error if you try to send a Point:

```
planet features items add "cli-test-123456" test.geojson 
Error: [{"geometry":"invalid geometry provided"}]
```

(If that error is vague I would probably lean towards improving the error msg returned by Features API)

docs/python/async-sdk-guide.md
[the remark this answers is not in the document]
stephenhillier: added note about feature refs:

> When passing a Features API Feature to other methods, the [feature ref](https://developers.planet.com/docs/apis/features/feature-references/) will be used. This means any searches or subscriptions you create will be linked to your feature.


added note about reservations:

>Reserving quota for features is currently not supported in the SDK. However, you may create features within the SDK and then use [Features Manager](https://planet.com/features) to reserve quota.

@asonnenschein also added a quick comment `# collection_id: the ID of a collection in Features API` to clarify what `collection_id` is. Note that you can use the ID, I don't think  `pl:ref/my` style will work for collections here

planet/clients/features.py
stephenhillier: > You can submit geojson Geometries as-is, I don't know why I bothered converting them to Features here

I was mistaken here, you can only submit Features. I put all this back, but I removed the list of geometries (in order to not imply that we accept points and lines)

docs/python/async-sdk-guide.md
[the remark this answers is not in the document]
asonnenschein: Awesome, thanks!

planet/cli/features.py
asonnenschein: Got it, thanks for explaining :)

planet/cli/features.py
asonnenschein: Yup, your example is exactly what I am thinking about.  That way a user could simply copy and paste a GeoJSON geometry instead of having to put it in a file.

planet/clients/features.py
asonnenschein: Understood, perhaps my comment here was too pedantic.  Seeing geometries like point and line in the array threw me off because I am so used to working with polygon and multi polygon geometries with Planet services.  Thanks for explaining!

PR #1074, opened 2024-10-30
Add source_type parameter to list subscriptions

### LINKED ISSUE #1072: Support source_type in SubscriptionsClient list_subscriptions
**Expected behavior**
The `source_type` parameter was added to the GET subscriptions endpoint some time ago

The parameter is not currently supported, making retrieval of subscriptions in organizations with many different sources less efficient.

**Related Issues**
#752 

**Workaround**
Filter subscriptions by source type after retrieving from API

tbarsballe:
**Related Issue(s):**

Closes #1072

**Proposed Changes:**

1. Add source_type parameter to list subscriptions

Not intended for changelog:

1. 

**PR Checklist:**

### DIFF
--- a/planet/cli/subscriptions.py
+++ b/planet/cli/subscriptions.py
@@ -78,14 +78,21 @@ def subscriptions(ctx, base_url):
     multiple=True,
     default=None,
     help="Select subscriptions in one or more states. Default is all.")
+@click.option(
+    '--source-type',
+    default=None,
+    help="Filter subscriptions by source type. See documentation for all "
+    "available types. Default is all.")
 @limit
 @click.pass_context
 @translate_exceptions
 @coro
-async def list_subscriptions_cmd(ctx, status, limit, pretty):
+async def list_subscriptions_cmd(ctx, status, source_type, limit, pretty):
     """Prints a sequence of JSON-encoded Subscription descriptions."""
     async with subscriptions_client(ctx) as client:
-        async for sub in client.list_subscriptions(status=status, limit=limit):
+        async for sub in client.list_subscriptions(status=status,
+                                                   source_type=source_type,
+                                                   limit=limit):
             echo_json(sub, pretty)
 
 
--- a/planet/clients/subscriptions.py
+++ b/planet/clients/subscriptions.py
@@ -1,7 +1,7 @@
 """Planet Subscriptions API Python client."""
 
 import logging
-from typing import AsyncIterator, Optional, Sequence
+from typing import AsyncIterator, Optional, Sequence, Dict, Union
 
 from typing_extensions import Literal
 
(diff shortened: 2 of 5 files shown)


planet/cli/subscriptions.py
```diff
@@ -78,14 +78,22 @@ def subscriptions(ctx, base_url):
     multiple=True,
     default=None,
     help="Select subscriptions in one or more states. Default is all.")
+@click.option(
+    '--source-type',
+    default=None,
+    help=
+    "Filter subscriptions by source type. See documentation for all available types. Default is all."
```
tbarsballe, line 85: yapf --in-place filename (from [CONTRIBUTING.md](https://github.com/planetlabs/planet-client-python/blob/main/CONTRIBUTING.md#yapf)) is evidently insufficient, as it produced:
```
    help=
    "Filter subscriptions by source type. See documentation for all available types. Default is all."
```
Which [failed linting](https://github.com/planetlabs/planet-client-python/actions/runs/11584616504/job/32252092775?pr=1074)

tbarsballe: I was able to fix manually, but it'd be nice if our automated linting actually worked as described.

planet/clients/subscriptions.py
```diff
@@ -89,7 +91,11 @@ class _SubscriptionsPager(Paged):
             """Navigates pages of messages about subscriptions."""
             ITEMS_KEY = 'subscriptions'
 
-        params = {'status': [val for val in status or {}]}
+        params = {
+            'status': [val for val in status or {}]
+        }  # type: Dict[str, Union[str, Sequence[str]]]
+        if source_type is not None:
+            params['source_type'] = source_type
```
tbarsballe, line 98: I can't seem to dig myself out of the typing on this one:

- With no typehint, mypy fails with `error: Incompatible types in assignment (expression has type "str", target has type "list[str]"), because of the initialization for status
- With type hints and import of Dict and Union, mypy passes but yapf fails with a linting error because the imported "aren't used" (since the hint is a comment)
- And removing the Dict and Union gets us back to a mypy failure because Dict and Union aren't imported

Anyone who's a bit more familiar with this toolset have a solution?

tbarsballe: I've currently got it working with `#type: ignore`, which isn't as clean as I'd like, but is the only solution I can figure out.

asonnenschein: Try adding the import statement at the top of the file: 

```python
from typing import Dict, Union, Sequence
```

And defining the type hints like this:

```python
params: Dict[str, Union[str, Sequence[str]]] = {
    'status': [val for val in status or {}]
}
if source_type is not None:
    params['source_type'] = source_type
```

asonnenschein (approved):
LGTM!  I tested the changes here out in my local dev and it all works as expected.  Cool addition @tbarsballe.

planet/clients/subscriptions.py
tbarsballe: That appears to have worked. Thanks!

PR #1078, opened 2024-11-26
Add planet.sync Namespace

asonnenschein:
This PR is branched off of https://github.com/planetlabs/planet-client-python/pull/1049.  It takes the new sync DataAPI, Subscriptions API, and OrdersAPI clients, and Planet wrapper class, and puts them in a new `planet.sync` namespace.  The original async clients still live in their original namespace to support backwards compatibility.  While I am open to the idea of adding an additional namespace, `planet.async` for the async clients, I did not include that in this PR because we need to support backwards compatibility for the existing async client patterns and adding `planet.async` would then be a third namespace which adds more complexity through maintenance overhead.

Examples:
```python
# Using the synchronous Planet wrapper class
from planet.sync import Planet

pl = Planet()
pl.subscriptions.get_subscription()

# Using a sync client directly
from planet.http import Session
from planet.sync import SubscriptionsAPI

client = SubscriptionsAPI(Session())
client.get_subscription()
```


planet/__init__.py
```diff
@@ -29,7 +27,6 @@
     'OrdersClient',
     'order_request',
     'reporting',
-    'Planet',
```
stephenhillier, line 32: What if we kept `Planet` exported at the top-level (e.g. `from planet import Planet` and only move the individual API classes to `planet.sync`? Thinking through this a little bit more since the discussion yesterday, `from planet.sync import Planet` could be harder for users to discover without referencing the docs, and `sync` is probably not as meaningful to end users as it is for us.

`from planet import *`  may be cluttered but I think it's appropriate for Planet/AsyncPlanet (our entrypoints) to get first pick at this namespace.

Otherwise this looks like a good approach to address the docs issue - just thinking out loud about the best way for users to discover how to start using the SDK.

asonnenschein: > from planet import * may be cluttered but I think it's appropriate for Planet/AsyncPlanet (our entrypoints) to get first pick at this namespace.

Good idea @stephenhillier, I updated this PR to promote the sync Planet class wrapper to the top level namespace.  So it can be imported via `from planet import Planet`, and `from planet import *`.  I still have the sync Planet class wrapper living in `./sync/client.py`.

planet/sync/__init__.py
```diff
@@ -0,0 +1,25 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+from .data import DataAPI
+from .orders import OrdersAPI
+from .subscriptions import SubscriptionsAPI
+from .client import Planet
+
+__all__ = ['DataAPI', 'OrdersAPI', 'Planet', 'SubscriptionsAPI']
+
+# Organize client classes by their module name to allow lookup.
```
ischneider, line 22: I don't think we need/want this ... this is a relic of a bizarre pattern that I believe came from dealing w/ circular deps

planet/sync/__init__.py
```diff
@@ -0,0 +1,25 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+from .data import DataAPI
+from .orders import OrdersAPI
+from .subscriptions import SubscriptionsAPI
+from .client import Planet
+
+__all__ = ['DataAPI', 'OrdersAPI', 'Planet', 'SubscriptionsAPI']
```
ischneider, line 20: Nit: the whole idea of Planet is that the user no longer has to consider the other APIs... If they want to, they can import them individually but I think it's confusing the import the "shortcut" (Planet) as well as the thing it shortcuts to...

planet/sync/client.py
```diff
@@ -34,7 +34,8 @@ class Planet:
           The session can be used to control the authentication method. Example:
 
           ```python
-          from planet import Auth, Session, Planet
+          from planet import Auth, Session
+          from planet.sync import Planet
 
           auth = Auth.from_key('examplekey')
```
ischneider, line 40: I know this is existing doc but maybe we should fix the example so folks don't get the idea they should hard-code their key into source code... e.g. 

`auth = Auth.from_key(get_key_from_encrypted_file())` 

or better, just note the Session can be provided allowing for whatever customization but defaults to standard behavior...

planet/sync/__init__.py
asonnenschein: Good idea to keep things simple.  I removed the API clients from the __all__ namespace and just have the `Planet` wrapper.

PR #1080, opened 2024-12-03
Hotfix: Pin httpx to v0.27.2

asonnenschein:
The most recent release of httpx v0.28.0 breaks all of our test pipelines across new and existing PRs with `AllMockedAssertionError`. 

This PR pins httpx to v0.27.2, which is the version that immediately precedes v0.28.0.  We should keep httpx pinned to v0.27.2 until the issue is resolved to ensure that our CI pipelines are not blocked.

respx ticket - https://github.com/lundberg/respx/issues/277
possible fix - https://github.com/lundberg/respx/pull/278

I am subscribed to both of the above respx ticket and PR, and will plan to follow up and configure httpx to pick up the latest version after the issue is resolved.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -9,7 +9,7 @@ description = "Planet SDK for Python"
 dependencies = [
     "click>=8.0",
     "geojson",
-    "httpx>=0.23.0",
+    "httpx==0.27.2",
     "jsonschema",
     "pyjwt>=2.1",
     "tqdm>=4.56",
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -9,7 +9,7 @@ description = "Planet SDK for Python"
 dependencies = [
     "click>=8.0",
     "geojson",
-    "httpx==0.27.2",
+    "httpx<0.28.0",
     "jsonschema",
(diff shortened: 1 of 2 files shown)


ischneider (approved):
I think this is fine but in the future, consider a min/max version specifier ... e.g. `<0.28.0,>=0.23.0`

asonnenschein:
> I think this is fine but in the future, consider a min/max version specifier ... e.g. `<0.28.0,>=0.23.0`

@ischneider Good idea!  Updated.

lundberg:
RESPX version `0.22.0` released

asonnenschein:
> RESPX version `0.22.0` released

Thanks @lundberg!  I opened https://github.com/planetlabs/planet-client-python/pull/1089 to upgrade our versions of `respx` and `httpx`.  In that PR I noticed some weird behavior after upgrading - whitespace is being stripped from byte literals somewhere in the call pattern.  Is this expected behavior?  See https://github.com/planetlabs/planet-client-python/pull/1089#discussion_r1901184316 as an example of what I am talking about.

PR #1089, opened 2025-01-02
Upgrade httpx to >= 0.28.0

asonnenschein:
This MR upgrades the `httpx` dependency version to >= 0.28.0, which was pinned to < 0.28.0 in https://github.com/planetlabs/planet-client-python/pull/1080 to work around a broken `respx` dependency which is now resolved with `respx` version 0.22.0 (https://github.com/planetlabs/planet-client-python/pull/1080#issuecomment-2555899485).

### DIFF
--- a/tests/integration/test_orders_cli.py
+++ b/tests/integration/test_orders_cli.py
@@ -307,8 +307,8 @@ def mock_download_response(oid, order_description):
                                     })
         respx.get(dl_url2).return_value = mock_resp2
 
-        m1_bytes = b'{"key": "value"}'
-        m2_bytes = b'{"key2": "value2"}'
+        m1_bytes = b'{"key":"value"}'
+        m2_bytes = b'{"key2":"value2"}'
         manifest_data = {
             "name": "",
             "files": [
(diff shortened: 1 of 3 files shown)


tests/integration/test_orders_cli.py
```diff
@@ -307,8 +307,8 @@ def _func():
                                     })
         respx.get(dl_url2).return_value = mock_resp2
 
-        m1_bytes = b'{"key": "value"}'
-        m2_bytes = b'{"key2": "value2"}'
+        m1_bytes = b'{"key":"value"}'
+        m2_bytes = b'{"key2":"value2"}'
```
asonnenschein, line 311: Not sure how I feel about this...  On `httpx==0.27.2`/`respx==0.20.2`, the whitespace is supported for these two byte literals.  This MR upgrades to `httpx==0.28.1`/`respx==0.22.0`, the whitespace is NOT supported for these two byte literals.  Somewhere in the call pattern the whitespace is stripped when present.

stephenhillier: It looks like httpx now uses a more compact json serialization. Updating these tests seems like the right way to go.

stephenhillier (approved):

Issue #734, opened 2022-10-30
Update Orders Product Bundles Reference JSON file and link to SSoT JSON file


JuLeeAtPlanet:
1. Note filename of the Orders Product Bundles Reference JSON file:

planet/data/orders_product_bundle_2022_02_02.json

3. In planet/specs.py, note:

PRODUCT_BUNDLE_SPEC_NAME = 'orders_product_bundle_2022_02_02.json'

4. In planet/data/README.md, note:

### Files

* `orders_product_bundle_2022_02_02.json`

Expected: [latest Orders product bundles reference](https://developers.planet.com/theme/js/2022-04-20.json) in the SDK | CLI

Actual: Product Bundle from 2/2/22 (planet/data/orders_product_bundle_2022_02_02.json)

Workaround: Go to latest spec page on [Planet](https://developers.planet.com/apis/orders/product-bundles-reference/) and select from there.

asonnenschein:
Current orders bundle is out of date, most recent is 2024-03-11.json.  This needs to be updated in the SDK, and the product-bundles docs need to be updated to include a line item for updating the SDK.

tbarsballe:
This is partially resolved by https://github.com/planetlabs/planet-client-python/pull/1058

PR #1094, opened 2025-01-16
remove product bundles json from repo, and fetch it dynamically, add helper function to get valid item types and bundles in cli

charcey:
**Description**
Currently we have a copy of the [orders product bundle spec](https://developers.planet.com/apis/orders/product-bundles-reference/) in this repository. When there is an update to the bundles spec _someone_ ends up having to update the spec here. This can cause headaches for users who might expect to be able to order a new item type but can't until the spec is updated.

Recently, a [new endpoint](https://developers.planet.com/apis/orders/reference/#tag/Orders/operation/getBundlesSpec) was added to the Orders API that serves the spec. This presents a great opportunity to dynamically grab the spec instead of managing a version of it statically. 

(editors note: removed old text here, and just laying out the implemented path after internal discussions)

The bundles spec information is lazily loaded into an envvar PRODUCT_BUNDLES. The first time the variable is accessed, the spec is fetched from the previously mentioned public endpoint, and every subsequent access (within the same runtime) will use the previously loaded information. It tries twice to get the spec and since we expect the spec endpoint to be highly available there is no work around if the spec cannot be accessed.

Side effects:
Previously the spec was loaded in (from a static file) on any data, orders, or subscriptions cli help command (e.g. `planet data --help`). This was to list the valid item types which is pulled from the bundles spec (there are other issues around this, but not solving that here). In addition, for orders help commands the bundle names were loaded in. Because having to reach out to the internet to run --help is not a good practice, this information is now available through new commands : `planet data show-item-types`, `planet orders show-item-types`, `planet orders show-bundles`, `planet subscriptions show-item-types`. These changes plus the lazy loading ensures we are only reaching out to the internet for the spec when absolutely necessary. 🥳 

Examples:
```
charcey-C02DP4WKML85:planet-client-python charcey$ planet orders --help
Usage: planet orders [OPTIONS] COMMAND [ARGS]...

  Commands for interacting with the Orders API

Options:
  -u, --base-url TEXT  Assign custom base Orders API URL.
  --help               Show this message and exit.

Commands:
  cancel           Cancel order by order ID.
  create           Create an order.
  download         Download order by order ID.
  get              Get order
  list             List orders
  request          Generate an order request.
  show-bundles     Show valid bundle names for ordering.
  show-item-types  Show valid item types for ordering.
  wait             Wait until order reaches desired state.

charcey-C02DP4WKML85:planet-client-python charcey$ planet orders show-item-types
Valid item types:
- SkySatScene
- TanagerMethane
- REOrthoTile
- PSScene
- TanagerScene
- REScene
- PelicanScene
- SkySatCollect

charcey-C02DP4WKML85:planet-client-python charcey$ planet orders show-bundles
Valid bundles:
- analytic
- radiance_hdf5
- analytic_udm2
...
- methane
- integrated_methane_enhancement
```

**Related Issue(s):**

Closes https://github.com/planetlabs/planet-client-python/issues/734

**Proposed Changes:**

1. Orders product bundles spec is fetched dynamically
2. `planet data show-item-types` to show valid item types for data requests
3. `planet orders show-bundles` to show valid bundle names and `planet orders show-item-types` to show valid item types
4. `planet subscriptions show-item-types` to show valid bundle names for scene subscriptions

Not intended for changelog:

1. Updated _lots_ of tests. Removing all PSOrthoTile references as that has been deprecated, using PSScene instead. New tests for new cli commands. Use mocked spec for all tests. Fixed various typos.
2. Refactored access patterns of spec in specs.py to reduce traversing json so much. Preloading bundle names and item types into new keys in the dictionary.

**Diff of User Interface**

Old behavior: N/A

New behavior: N/A

User behavior is the exact same.

However, if the spec cannot be reached, an error will surface to the user.

Example:
```
>>> from planet.order_request import product
>>> product(
...                     item_ids=[
...                         "20220304_093300_37_2430",
...                         "20220305_093440_25_2429",
...                     ],
...                     product_bundle="analytic_udm2",
...                     item_type="PSScene",)
  File "<stdin>", line 1, in <module>
  File "/Users/charcey/workspace/planet-client-python/planet/order_request.py", line 125, in product
    item_type = specs.validate_item_type(item_type)
                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/charcey/workspace/planet-client-python/planet/specs.py", line 112, in validate_item_type
    PRODUCT_BUNDLES["item_types"],
    ~~~~~~~~~~~~~~~^^^^^^^^^^^^^^
  File "/Users/charcey/workspace/planet-client-python/planet/specs.py", line 62, in __getitem__
    raise FetchBundlesSpecError(
planet.specs.FetchBundlesSpecError: Unable to fetch product bundles spec from API to perform client side validation on item types and product bundles. Please retry!
```

**PR Checklist:**

- [X] This PR is as small and focused as possible
- [X] If this PR includes proposed changes for inclusion in the changelog, the title of this PR summarizes those changes and is ready for inclusion in the Changelog.
- [X] I have updated docstrings for function changes and docs in the 'docs' folder for user interface / behavior changes
- [X] This PR does not break any examples or I have updated them


ischneider:
Here's a way to replace the module import side-effect:

```
class _LazyLoader:

    def __getitem__(self, key):
        cache = getattr(self, "cache", None)
        if cache is None:
            response = httpx.get("https://api.planet.com/compute/ops/bundles/spec")
            response.raise_for_status()
            cache = response.json()
            setattr(self, "cache", cache)
        return cache[key]

PRODUCT_BUNDLES_SPEC_JSON = _LazyLoader()
```

I feel like we might want to introduce minimal retry (mostly to avoid a potential network hiccup) and  clearer handling of unexpected conditions as this could be a source of confusion ...

planet/specs.py
```diff
@@ -36,10 +36,42 @@
...
+
+    def __getitem__(self, key):
+        bundles_spec_url = "https://api.planet.com/compute/ops/bundles/spec"
+        retries = 2
+        cache = getattr(self, "cache", None)
+        if cache is None:
+            for attempt in range(1, retries + 1):
+                try:
+                    response = httpx.get(bundles_spec_url)
+                    response.raise_for_status()
```
ischneider, line 55: make sure to `break` or the retry loop will run again!

planet/specs.py
```diff
@@ -36,10 +36,42 @@
...
+                except:
+                    if attempt == retries:
+                        raise RuntimeError(f"Unable to fetch product bundles spec from API to perform client side validation on item and asset types.\nPlease retry!\nIn persistent cases - disable client side validation by using the skip_client_validation parameter in the SDK --skip-client-validation flag in the CLI.")
+            cache = {
+                'bundles': response.json()['bundles'],
```
ischneider, line 60: minor nit/optimization:
`response.json()`  will parse the content again, so best to do it ahead of time and reuse the dict

planet/specs.py
charcey: added! good eye.

planet/specs.py
charcey: good catch! updated to load it once ahead of time.

planet/cli/data.py
```diff
@@ -652,3 +654,13 @@ async def asset_wait(ctx, item_type, item_id, asset_type, delay, max_attempts):
 
 # TODO: search_run()".
 # TODO: item_get()".
+
+
+@data.command()  # type: ignore
+@click.pass_context
+@translate_exceptions
+def show_item_types(ctx):
```
asonnenschein, line 662: nit: instead of `planet data show-item-types` could this command be renamed `planet data item-types`?  Some personal preference, but I like to try and keep CLI command names short, and keep nouns and verbs separate so it's easier to scale use cases moving forward.  

If `show` is important for the context of the command could it be reworked as a sub-command to `item-types`?  So something like `planet data item-types show`.  Basically separating the noun, `item-types` from the verb, `show`.  The default behavior to `planet data item-types` could be `show` so that way the command is shorter to type, and it will be easy to enhance if we ever want to add more actions to how we interact with item-types (e.g. `planet data item-types filter`, `planet data item-types version`, etc).

planet/cli/orders.py
```diff
@@ -426,3 +464,23 @@ async def request(ctx,
                                                  collection_id=collection_id)
 
     echo_json(request, pretty)
+
+
+@orders.command()  # type: ignore
+@click.pass_context
+@translate_exceptions
+def show_item_types(ctx):
```
asonnenschein, line 472: nit: `planet orders item-types` instead of `planet orders show-item-types`.

planet/cli/orders.py
```diff
@@ -426,3 +464,23 @@ async def request(ctx,
...
+    """Show valid item types for ordering."""
+    click.echo("Valid item types:")
+    for it in get_item_types():
+        click.echo(f"- {it}")
+
+
+@orders.command()  # type: ignore
+@click.pass_context
+@translate_exceptions
+def show_bundles(ctx):
```
asonnenschein, line 482: nit: `planet orders bundles` instead of `planet orders show-bundles`.

planet/cli/subscriptions.py
```diff
@@ -505,3 +493,13 @@ def request_pv(var_type, var_id, geometry, start_time, end_time, pretty):
         end_time=end_time,
     )
     echo_json(res, pretty)
+
+
+@subscriptions.command()  # type: ignore
+@click.pass_context
+@translate_exceptions
+def show_item_types(ctx):
```
asonnenschein, line 501: nit: `planet subscriptions item-types` instead of `planet subscriptions show-item-types`.

planet/specs.py
```diff
@@ -36,10 +34,53 @@
...
+        if cache is None:
+            for attempt in range(1, retries + 1):
+                try:
+                    response = httpx.get(bundles_spec_url)
+                    response.raise_for_status()
+                    break
+                except:  # noqa: E722
+                    if attempt == retries:
+                        raise FetchBundlesSpecError(
+                            "Unable to fetch product bundles spec from API to perform client side validation on item types and product bundles. Please retry!"
```
asonnenschein, line 63:
```suggestion
                            "Unable to fetch product bundles spec. Please retry!"
```
Will users always want to fetch the bundles spec for validation?  This MR looks like it also supports showing the bundles spec for discovery.

asonnenschein, line 63:
```suggestion
                            "Unable to fetch product bundles spec. Please retry!"
```
Will users always want to fetch the bundles spec for validation?  This MR looks like it also supports showing the bundles spec for discovery.

charcey: Ah, good catch. I didn't update this message after adding the new cli commands. However, users don't have a choice lol the spec has to be fetched for any client side validation OR item type/bundle discovery. Maybe it's overkill to mention item types in the error message though.

planet/cli/data.py
charcey: updated!

planet/cli/orders.py
charcey: updated!

planet/cli/orders.py
charcey: updated!

planet/cli/subscriptions.py
charcey: udpated!

planet/specs.py
charcey: updated the language, lmk what you think

charcey: duplicate comment?

asonnenschein (approved):

planet/specs.py
```diff
@@ -36,10 +34,53 @@
...
+                    response = httpx.get(bundles_spec_url)
+                    response.raise_for_status()
+                    break
+                except:  # noqa: E722
+                    if attempt == retries:
+                        raise FetchBundlesSpecError(
+                            "Unable to fetch spec from API to generate valid item types and bundles. Please retry!"
+                        ) from None
+            bundles = response.json()['bundles']
+            cache = {
```
ischneider, line 66: the formatting here is kinda wonky and I'm surprised it didn't make the linter sad ... can we do `"key" : value` ?

charcey: Heh, i had it the way you suggested there and the linter wanted this change. 😭

PR #1096, opened 2025-01-29
Add page_size arg to list_subscriptions()

asonnenschein:
This MR adds a new `page_size` arg to the `SubscriptionsClient.list_subscriptions()` method.  Without specifying a page size, `list_subscriptions()` will fall back on a default page size of 20 subscriptions per page that is defined in the API view handler.  For large numbers of subscriptions, this results in long execution times because `list_subscriptions()` is only getting 20 subscriptions per page, which results in more requests (and latency in transport) than is necessary.  

For example, I tested this out on a sample of 645 subscriptions.  The average execution time to get all 645 subscriptions using the default page size of 20 (`list_subscriptions()`) takes ~36 seconds because `list_subscriptions()` is making ~33 requests to the API:
```
(planet-client-python-3.12) ➜  Desktop python example.py
Average execution time: 36.07816209201701 seconds
Average execution time: 37.37259578681551 seconds
Average execution time: 36.77448189887218 seconds
Average execution time: 36.13904633792117 seconds
Average execution time: 35.90926657919772 seconds
```

When I define a page_size of 700 (`list_subscriptions(page_size=700)`) the average execution time to get all 645 subscriptions takes only ~3 seconds because `list_subscriptions()` only has to make one request to the API with the larger page size:
```
(planet-client-python-3.12) ➜  Desktop python example.py                          
Average execution time: 3.206969585036859 seconds
Average execution time: 2.6358285231981426 seconds
Average execution time: 2.8017055389937013 seconds
Average execution time: 2.6492277118377388 seconds
Average execution time: 2.728695012163371 seconds
```

Here is the contents of `example.py` that I used to test and record execution times:
```python
from planet import Session, SubscriptionsClient

async def get_subscriptions():
    async with Session() as sess:
        cl = SubscriptionsClient(sess)
        subscriptions = cl.list_subscriptions(page_size=700, limit=5000)
        sub_list = [i async for i in subscriptions]
        return sub_list

async def main():
    subs = await get_subscriptions()
    return subs

def run_main():
    return asyncio.run(main())

if __name__ == '__main__':
    import asyncio
    import timeit
    execution_times = timeit.repeat(run_main, number=10, repeat=5)
    for execution_time in execution_times:
        print(f"Average execution time: {execution_time} seconds")
```

### DIFF
--- a/planet/cli/subscriptions.py
+++ b/planet/cli/subscriptions.py
@@ -118,7 +118,7 @@ def subscriptions(ctx, base_url):
     documentation
     for examples.""")
 @limit
-@click.option('--page_size',
+@click.option('--page-size',
               type=click.INT,
               help='Number of subscriptions to return per page.')
 @click.pass_context
--- a/planet/cli/subscriptions.py
+++ b/planet/cli/subscriptions.py
@@ -118,6 +118,9 @@ def subscriptions(ctx, base_url):
     documentation
     for examples.""")
 @limit
+@click.option('--page_size',
+              type=click.INT,
+              help='Number of subscriptions to return per page.')
 @click.pass_context
 @translate_exceptions
 @coro
@@ -133,6 +136,7 @@ async def list_subscriptions_cmd(ctx,
                                  sort_by,
                                  updated,
                                  limit,
+                                 page_size,
                                  pretty):
     """Prints a sequence of JSON-encoded Subscription descriptions."""
     async with subscriptions_client(ctx) as client:
@@ -147,7 +151,8 @@ async def list_subscriptions_cmd(ctx,
                 status=status,
                 sort_by=sort_by,
                 updated=updated,
-                limit=limit):
+                limit=limit,
+                page_size=page_size):
             echo_json(sub, pretty)
 
 
--- a/planet/clients/subscriptions.py
+++ b/planet/clients/subscriptions.py
@@ -65,20 +65,19 @@ class SubscriptionsClient:
         """block on an async function call, using the call_sync method of the session"""
         return self._session._call_sync(f)
 
-    async def list_subscriptions(
-            self,
-            status: Optional[Sequence[str]] = None,
-            limit: int = 100,
-            created: Optional[str] = None,
-            end_time: Optional[str] = None,
-            hosting: Optional[bool] = None,
-            name__contains: Optional[str] = None,
-            name: Optional[str] = None,
-            source_type: Optional[str] = None,
-            start_time: Optional[str] = None,
-            sort_by: Optional[str] = None,
-            updated: Optional[str] = None,
-            page_size: int = 500) -> AsyncIterator[dict]:
+    async def list_subscriptions(self,
+                                 status: Optional[Sequence[str]] = None,
+                                 limit: int = 100,
+                                 created: Optional[str] = None,
+                                 end_time: Optional[str] = None,
+                                 hosting: Optional[bool] = None,
+                                 name__contains: Optional[str] = None,
+                                 name: Optional[str] = None,
+                                 source_type: Optional[str] = None,
+                                 start_time: Optional[str] = None,
+                                 sort_by: Optional[str] = None,
+                                 updated: Optional[str] = None,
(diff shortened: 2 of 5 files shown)


planet/clients/subscriptions.py
```diff
@@ -69,6 +69,7 @@ async def list_subscriptions(
             self,
             status: Optional[Sequence[str]] = None,
             limit: int = 100,
+            page_size: Optional[int] = None,
```
stephenhillier, line 72: any new params should go at the end, since users may be supplying these as positional args.

planet/clients/subscriptions.py
```diff
@@ -156,6 +158,8 @@ class _SubscriptionsPager(Paged):
             params['sort_by'] = sort_by
         if updated is not None:
             params['updated'] = updated
+        if page_size is not None:
+            params['page_size'] = page_size
```
stephenhillier, line 162: Should the default just be larger pages? If requests are most efficient at 100+ results per page I see no reason to default to 20 and make users tune it.

Is it possible we can just bump the default and not even add an option (I'm not necessarily opposed to adding more options if they provide value, but perhaps tweaking the default solves the issue)?

HenryW95: Agreed if we can just change the default that would be most ideal. It might be an odd interface for the user to specify page_size when they're not getting back pages from the client.

planet/clients/subscriptions.py
asonnenschein: Good catch, updated!

planet/clients/subscriptions.py
asonnenschein: Updated the default page size to 500.  I like the idea of increasing the default page size to something more performant, but I also think we should allow the user to set a custom page size.  The `page_size` arg resolves to a `page_size` query parameter on the API endpoint that is configurable, I like the idea of extending that pattern to the SDK.

stephenhillier: sounds good! we should add this to the sync client too. You should be able to just pass the value straight through

asonnenschein: For sure! I added the 500 page_size default to the sync client too :)

planet/clients/subscriptions.py
```diff
@@ -135,7 +136,7 @@ class _SubscriptionsPager(Paged):
             """Navigates pages of messages about subscriptions."""
             ITEMS_KEY = 'subscriptions'
 
-        params: Dict[str, Union[str, Sequence[str], bool]] = {}
+        params: Dict[str, Union[str, Sequence[str], bool, int]] = {}
```
stephenhillier, line 139: I would just use `Dict[str, Any]` but this is just a nit and I will leave it up to you. Otherwise everything looks good :+1:

asonnenschein: Updated

asonnenschein:
> Would it make sense to add page_size to the CLI for parody with the client?

@HenryW95 yes, added!

HenryW95:
> > Would it make sense to add page_size to the CLI for parody with the client?
> 
> @HenryW95 yes, added!

This docs page also lists the CLI params, so it can be added there: https://github.com/planetlabs/planet-client-python/blob/main/docs/cli/cli-subscriptions.md

asonnenschein:
> > > Would it make sense to add page_size to the CLI for parody with the client?
> > 
> > 
> > @HenryW95 yes, added!
> 
> This docs page also lists the CLI params, so it can be added there: https://github.com/planetlabs/planet-client-python/blob/main/docs/cli/cli-subscriptions.md

@HenryW95 Updated 👍

HenryW95 (changes requested):

planet/cli/subscriptions.py
```diff
@@ -118,6 +118,9 @@ def subscriptions(ctx, base_url):
     documentation
     for examples.""")
 @limit
+@click.option('--page_size',
```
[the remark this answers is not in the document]
asonnenschein: Good catch!  This was probably copy pasta from the SDK param.  Updated 👍

Issue #1105, opened 2025-03-13
Order API client support for multiple fallback bundle(s)


ceholden:
**Is your feature request related to a problem? Please describe.**

It would be very convenient for us if the Orders API client in this package supported more than one "fallback bundle(s)". The API documentation and our experience using the Order API directly suggest that this is possible, so it would be great to have support from this Python client.

The [API documentation](https://developers.planet.com/apis/orders/scenes/#fallback-bundles) suggests we can supply multiple "alternative bundle(**s**)",
>To specify a fallback bundle simply add the alternate bundle(s) to the product_bundle field separated by commas.

This example has worked for us in practice - note the 2 fallback bundles,
```python
payload = {
    "name": "mre-test-order",
    "order_type": "partial",
    "products": [
        {
            "item_type": "PSScene",
            "item_ids": ["20250130_035211_69_2516"],
            "product_bundle": "analytic_8b_udm2,analytic_udm2,analytic_3b_udm2",
        }
    ]
}
req = requests.post(
    f"https://api.planet.com//compute/ops/orders/v2",
    json=payload,
    auth=requests.auth.HTTPBasicAuth(PLANET_API_KEY, ""),
)
```

If I try to add multiple fallback bundles using the SDK as either a `list[str]` or comma joined list, I get either,

```python
# given as a list[str]
AttributeError: 'list' object has no attribute 'lower'

# pre-joined with a comma
SpecificationException: bundle - 'analytic_udm2,analytic_3b_udm2' is not one of 'analytic_udm2', 'analytic_3b_udm2', 'analytic_8b_udm2', 'visual', 'basic_analytic_udm2', 'basic_analytic_8b_udm2', 'analytic_sr_udm2', 'analytic_8b_sr_udm2'.
```

**Describe the solution you'd like**

It seems like this addition could be added without much complexity and in a backwards compatible way. One possible solution would be to update the `planet.order_request.product` function's call signature to support `fallback_bundle: Optional[Union[str, List[str]]]` instead of `fallback_bundle: Optional[str]` and add a little more handling logic for the `List[str]` case.

For example,
```python
def product(item_ids: List[str],
            product_bundle: str,
            item_type: str,
            fallback_bundle: Optional[Union[str, List[str]]] = None) -> dict:
    ...

    if fallback_bundle is not None:
        fallback_bundles = [fallback_bundle] if isinstance(fallback_bundle, str) else fallback_bundle
        validated_fallback_bundles = []
[...]

PR #1106, opened 2025-03-19
Orders API Client Support For Multiple Fallback Bundle(s)

asonnenschein:
GH Ticket: https://github.com/planetlabs/planet-client-python/issues/1105

This MR adds support in the Orders SDK and CLI for specifying multiple fallback bundles, updates the docs, and updates the orders CLI and SDK tests with success and failure cases for fallback bundles.  

CLI example:
```
planet orders request --item-type "PSScene" \
    --bundle "analytic_8b_udm2" \
    --fallback-bundle "analytic_udm2,analytic_3b_udm2" \
    --name "my order" \
    "20250130_035211_69_2516"
```

SDK example:
```python
order_request.product(
    ["20250130_035211_69_2516"],
    "analytic_8b_udm2",
    "PSScene",
    "analytic_udm2,analytic_3b_udm2"
)
```

### DIFF
--- a/planet/cli/orders.py
+++ b/planet/cli/orders.py
@@ -361,11 +361,12 @@ async def create(ctx, request, pretty, **kwargs):
               help='Bundle type for the item.',
               type=str,
               callback=check_bundle)
-@click.option('--fallback-bundle',
-              required=False,
-              help='Fallback bundle type(s) for the item if bundle is not available.',
-              type=str,
-              callback=check_bundle)
+@click.option(
+    '--fallback-bundle',
+    required=False,
+    help='Fallback bundle type(s) for the item if bundle is not available.',
+    type=str,
+    callback=check_bundle)
 @click.option('--name',
               required=True,
               help='Order name. Does not need to be unique.',
@@ -451,7 +452,10 @@ async def request(ctx,
     IDs is one or more comma-separated item IDs.
     """
     try:
-        product = planet.order_request.product(ids, bundle, item_type, fallback_bundle)
+        product = planet.order_request.product(ids,
+                                               bundle,
+                                               item_type,
+                                               fallback_bundle)
     except planet.specs.SpecificationException as e:
         raise click.BadParameter(e)
 
--- a/planet/cli/orders.py
+++ b/planet/cli/orders.py
@@ -58,12 +58,14 @@ def check_bundle(ctx, param, bundle) -> Optional[List[dict]]:
     if not item_type:
         raise click.BadParameter("Item type is required to validate a bundle.")
 
-    try:
-        validate_bundle(item_type, bundle)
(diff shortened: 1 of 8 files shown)


pl-gideon (approved):

tests/unit/test_order_request.py
```diff
@@ -404,3 +404,38 @@ def test_sentinel_hub_collection_configuration():
...
     assert sh_config == expected
+
+
+@pytest.mark.parametrize(
+    "bundle, fallback_bundle",
+    [
+        ('analytic_8b_udm2', 'analytic_udm2'),
+        ('analytic_8b_udm2', 'analytic_udm2,analytic_3b_udm2'),
+    ])
+def test_fallback_bundle_success(bundle, fallback_bundle):
```
ceholden, line 415: This looks great! It might also be worth adding a case where the fallback bundle is a `list[str]` for completeness

planet/order_request.py
```diff
@@ -107,7 +107,7 @@ def build_request(name: str,
 def product(item_ids: List[str],
             product_bundle: str,
             item_type: str,
-            fallback_bundle: Optional[str] = None) -> dict:
+            fallback_bundle: Optional[Union[str, List[str]]] = None) -> dict:
```
ceholden, line 110: It might be nice to callout that this can support either 1 or several comma separated bundle names when passing a `str` in the docstring for `fallback_bundle` argument

Something like,
```python
    """
    ...

        fallback_bundle: In case product_bundle not having
            all asset types available, which would result in failed
            delivery, try one or more fallback bundles. Multiple
            fallback bundles may be provided as a list or a comma
            separated string.

    ...
    """
```

ceholden (commented):
I think this looks great and will fulfill the feature request, thank you! I had a couple suggestions about docstrings and a suggestion to cover the `fallback_bundle: list[str]` case in the tests added in this PR, but nothing functional to change. Nice work and thanks for incorporating into the user guide!

tests/unit/test_order_request.py
asonnenschein: Ah good catch!  Updated 👍.

planet/order_request.py
asonnenschein: Updated

planet/cli/orders.py
```diff
@@ -359,6 +364,12 @@ async def create(ctx, request, pretty, **kwargs):
               help='Bundle type for the item.',
               type=str,
               callback=check_bundle)
+@click.option(
+    '--fallback-bundle',
+    required=False,
+    help='Fallback bundle type(s) for the item if bundle is not available.',
```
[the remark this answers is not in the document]
asonnenschein: Updated

PR #1107, opened 2025-03-19
Draft: Make Subscription Source Type Optional

asonnenschein:
This MR updates the Subscriptions SDK and CLI to make the subscription source `type` parameter optional.  For catalog source subscriptions this is a simple change because the `catalog` source type was programmatically appended to the catalog source instead of requiring user input as a positional arg to `catalog_source()`.  To make the `catalog` type optional then I simply removed the line in `catalog_source()` that appends `catalog` source `type`.

For planetary variable source subscriptions this change is a bit more difficult to make without introducing a breaking change, because the source type is variable and is the first required positional arg to `planetary_variable_source()`.  To avoid the breaking change, I added a condition to `planetary_variable_source()` that checks if the `var_type` arg is `None` and only appends the source type if it is not `None`.  This gives us the desired UX for the CLI which no longer requires the `--var-type` parameter, but keeps the positional `var_type` arg in `planetary_variable_source()`.

Note - I did explore the possibility of making the positional `var_type` arg optional by counting the number of input args to `planetary_variable_source()`, checking their types, and inferring whether or not the user specified the `var_type` arg.  While this does technically work, I think it would introduce an anti-pattern to the method because `var_type` is the first positional arg in the method.

### DIFF
--- a/planet/cli/subscriptions.py
+++ b/planet/cli/subscriptions.py
@@ -483,7 +483,7 @@ def request_catalog(item_types,
 @translate_exceptions
 @click.option(
     '--var-type',
-    required=True,
+    required=False,
     help='A Planetary Variable type. See documentation for all available types.'
 )
 @click.option(
--- a/planet/subscription_request.py
+++ b/planet/subscription_request.py
@@ -281,7 +281,7 @@ def catalog_source(
     if time_range_type:
         parameters['time_range_type'] = time_range_type
 
-    return {"type": "catalog", "parameters": parameters}
+    return {"parameters": parameters}
 
 
 def planetary_variable_source(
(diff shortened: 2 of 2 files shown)


tests/integration/test_subscriptions_cli.py
```diff
@@ -398,24 +395,33 @@ def test_subscriptions_results_csv(invoke):
...
         "request-pv",
-        "--var-type=biomass_proxy",
         "--var-id=BIOMASS-PROXY_V3.0_10",
         f"--geometry={geom}",
         "--start-time=2021-03-01T00:00:00",
-    ])
+    ]
+
+    if source_type:
+        cmd.append(f"--var-type={source_type}")
```
pl-gideon, line 415: Half the checks in the MR check `if X` and the other check `is / is not None` - I am not sure that it really matters which is used but we should be mindful about what the expectation here is and be consistent if possible.

pl-gideon (approved):

tests/integration/test_subscriptions_cli.py
asonnenschein: Good eye @pl-gideon!  I updated those `None` checks 👍.

PR #1119, opened 2025-04-09
Support bulk subs creation in client and add --bulk to CLI to invoke it

goldenberg:
**Related Issue(s):**

Closes #

**Proposed Changes:**

1. Adds `client.bulk_create_subscriptions()` which allows creating multiple subscriptions at once by referencing a feature collection
2. Adds support for `planet subscriptions bulk-create` to bulk create using the above mechanism

Not intended for changelog:

1. 

**Diff of User Interface**

Old behavior:

New behavior:
❯ planet subscriptions request-catalog --item-types PSScene --asset-types ortho_visual --geometry pl:features/my/adrian-test-new-guinea-10geojson-xqRXaaZ --start-time 2021-01-01 --end-time 2021-01-05 > catalog_fc_sub_source.json
❯ planet subscriptions request --name "bg test bulk sub" --source catalog_fc_sub_source.json > catalog_fc_sub.json
❯ planet subscriptions create --bulk --hosting sentinel_hub catalog_fc_sub.json

**PR Checklist:**

### DIFF
--- a/docs/cli/cli-subscriptions.md
+++ b/docs/cli/cli-subscriptions.md
@@ -247,6 +247,49 @@ planet subscriptions patch cb817760-1f07-4ee7-bba6-bcac5346343f \
     patched-attributes.json
 ```
 
+### Bulk Create Subscriptions
+
+To create many subscriptions for different geometries at once, use the `bulk-create` subcommand.
+
+This command allows submitting a bulk create request that references a feature collection defined in
+the Features API, which will create a subscription for every feature in the collection.
+
+Define a subscription that references a feature collection:
+
+```json
+{
+  "name": "new guinea psscene bulk subscription",
+  "source": {
+    "parameters": {
+      "item_types": [
+        "PSScene"
+      ],
+      "asset_types": [
+        "ortho_visual"
+      ],
+      "geometry": {
+        "type": "ref",
+        "content": "pl:features/my/test-new-guinea-10geojson-xqRXaaZ"
+      },
+      "start_time": "2021-01-01T00:00:00Z",
+      "end_time": "2021-01-05T00:00:00Z"
+    }
+  }
+}
+```
+
+And issue the `bulk-create` command with the appropriate hosting or delivery options. A link to list
+the resulting subscriptions will be returned:
+
+```sh
+planet subscriptions bulk-create --hosting sentinel_hub catalog_fc_sub.json
+{
+  "_links": {
+    "list": "https://api.planet.com/subscriptions/v1?created=2025-04-16T23%3A44%3A35Z%2F..&geom_ref=pl%3Afeatures%2Fmy%2Ftest-new-guinea-10geojson-xqRXaaZ&name=new+guinea+psscene+bulk subscription"
+  }
+}
+```
+
 ### Cancel Subscription
 
 Cancelling a subscription is simple with the CLI:
--- a/planet/cli/subscriptions.py
+++ b/planet/cli/subscriptions.py
@@ -152,11 +152,6 @@ async def list_subscriptions_cmd(ctx,
 @subscriptions.command(name="create")  # type: ignore
 @click.argument("request", type=types.JSON())
 @click.option(
-    "--bulk",
-    is_flag=True,
-    help="Bulk create many subscriptions using a feature collection reference.",
-)
-@click.option(
     "--hosting",
     type=click.Choice([
         "sentinel_hub",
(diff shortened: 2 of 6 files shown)


goldenberg:
I've now tested this via the CLI by running:

```
❯ planet subscriptions request-catalog --item-types PSScene --asset-types ortho_visual --geometry pl:features/my/adrian-test-new-guinea-10geojson-xqRXaaZ --start-time 2021-01-01 --end-time 2021-01-05 > catalog_fc_sub_source.json
❯ planet subscriptions request --name "bg test bulk sub" --source catalog_fc_sub_source.json > catalog_fc_sub.json
❯ planet subscriptions create --bulk --hosting sentinel_hub catalog_fc_sub.json
```

And I believe it works. Needs more testing though

planet/sync/subscriptions.py
```diff
@@ -136,6 +136,22 @@ def create_subscription(self, request: Dict) -> Dict:
         return self._client._call_sync(
             self._client.create_subscription(request))
 
+    def bulk_create_subscriptions(self, requests: List[Dict]) -> Dict:
```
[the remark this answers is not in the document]
goldenberg: Good eye, Copilot.

planet/cli/subscriptions.py
```diff
@@ -151,6 +151,11 @@ async def list_subscriptions_cmd(ctx,
 
 @subscriptions.command(name="create")  # type: ignore
 @click.argument("request", type=types.JSON())
+@click.option(
+    "--bulk",
```
asonnenschein, line 155: I'm split on whether or not the CLI bulk create path should be an optional flag on the standard `create` argument as you have implemented here (e.g. `create --bulk`), or if it should be it's own argument (e.g. `bulk`, 
`bulk create`, `bulk_create`, `create_bulk`, etc). 

Including an optional `--bulk` flag on the standard `create` command implies that the `--bulk` flag modifies the behavior of the `create` command.  This isn't really the case though because bulk create has it's own API endpoint `POST /subscriptions/v1/bulk`.  Thinking forward, it might be difficult to enhance and/or modify the bulk create path in the CLI if tightly couple it with the standard create path.

Defining the bulk create path as a distinct argument implies that this is a distinct path, separate from the standard create path.  Keeping the paths separate should make it easier to enhance and/or modify the bulk create CLI UX without getting wires crossed with the standard create path.  

One more thing to think about is that the payloads for the two paths are slightly different.  The standard create path expects a single object of subscription criteria, and the bulk create path expects an array of objects that define subscription criteria.  I can see supporting two different request data structures for the same `create` command, and driven by the `--bulk` flag being a confusing UX.

Having typed this out and thought about it some more, I am leaning more towards the bulk create path having it's own distinct argument on account of the different endpoints and request data structures.

tests/integration/test_subscriptions_api.py
```diff
@@ -334,6 +345,18 @@ async def test_create_subscription_success():
         assert sub['name'] == 'test'
 
 
+@pytest.mark.anyio
+@bulk_create_mock
+async def test_bulk_create_subscription_success():
```
asonnenschein, line 350: This test only tests the `async` client.  I see in this MR that you've also implemented the bulk create method on the `sync` client.  Would you be able to add an integration test for the `sync` client as well?  Here's an example of what the `async` vs `sync` integration tests for the standard create success path looks like:

- [test_create_subscription_success()](https://github.com/planetlabs/planet-client-python/blob/607e62ecbe7e9580cce95f45e707996637730b59/tests/integration/test_subscriptions_api.py#L338C11-L338C43)
- [test_create_subscription_success_sync()](https://github.com/planetlabs/planet-client-python/blob/607e62ecbe7e9580cce95f45e707996637730b59/tests/integration/test_subscriptions_api.py#L361C5-L361C44) 

FWIW this nuance in testing isn't always apparent, and kind of a side effect of the sync client implementation that we decided to live until we have room for a more uniform testing pattern.  Some other notes - this was mentioned offline but I agree that we don't need unit tests for this enhancement, and I think it's safe to ship without a failure test because these are all mocked to begin with and any mocked errors that you'd catch would be the same as the errors we get back from the mocked standard create path (which already has failure coverage).

goldenberg: Ah, thanks, missed this. Added the sync companion test.

I had the same thought about failure tests. Seems like it would just be testing the mock again.

planet/cli/subscriptions.py
goldenberg: That all makes sense to me! Update to break into a bulk-create subcommand. Let me know how it looks.

asonnenschein:
LGTM!  I left one more comment about removing the `--bulk` flag option from the `create` command.  Also the SDK docs will autogenerate from the docstrings, but the CLI docs will need to be updated manually (https://github.com/planetlabs/planet-client-python/blob/main/docs/cli/cli-subscriptions.md) .  Otherwise this PR looks great!  I'll hit approve now to unblock you so you can merge after these couple of changes are made.

asonnenschein (approved):

planet/cli/subscriptions.py
goldenberg: Oops! Thought I grabbed that but apparently not. Fixed.

tests/integration/test_subscriptions_api.py
```diff
@@ -346,6 +369,18 @@ def test_create_subscription_success_sync():
...
+def test_bulk_create_subscription_success_sync():
+    """Subscription is created, description has the expected items."""
+
+    pl = Planet()
+    pl.subscriptions._client._base_url = TEST_URL
+    resp = pl.subscriptions.bulk_create_subscriptions({
+        'name': 'test', 'delivery': 'yes, please', 'source': 'test'
+    })
+    assert '/subscriptions/v1?' in resp['_links']['list']
+
```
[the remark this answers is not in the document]
goldenberg: Good catch, why didn't the test fail then? Hmm.

docs/cli/cli-subscriptions.md
```diff
@@ -247,6 +247,49 @@ planet subscriptions patch cb817760-1f07-4ee7-bba6-bcac5346343f \
     patched-attributes.json
 ```
 
+### Bulk Create Subscriptions
```
asonnenschein, line 250: Nice!  Thanks for adding the CLI documentation for `bulk-create`!

PR #1131, opened 2025-05-12
initial mosaics cli + async client

### LINKED ISSUE #446: Implement support for Basemaps API
Implement a CLI and Python API for the [basemaps API](https://developers.planet.com/docs/basemaps/)

ischneider:
WIP - mosaics API support

I based this on the V1 CLI/SDK but added some additional functionality, mostly in filters that have been added since the original client was created.

Rather than implement the sync SDK and fully document every parameter and write every test, I decided to get this out for some tire kicking first... so there might be a broken thing or two but pretty sure everything is working

The `examples/mosaics-cli.sh` script demonstrates some basic functionality and will download 2 quads to the directory it's run from.

General questions that popped up while developing:

1. Some operations do too much (without filters), e.g. list mosaics (for Planet users is 1000s of results), list/download quads for large (or default) extents

2. Download operations have an overwrite flag but this requires checking the content-disposition header to get the destination file name (at least with the current core implementation) though this will result in the request counting as a "download" (for quota) even if the stream is never read...

3. Some operations accept a name or ID (as a convenience, since ID is opaque) and others require a dict of the the resource representation - it would be nice to allow any of the above, i.e. `NameIDObject = Union[dict, str]` (this name is terrible but what should it be?)

Proposals (where I have an idea):

1. For requests that do too much: limit can be confusing when there's a default and users may not be aware of it. Instead, the CLI command can fail once too many results are processed. For quads, could use a heuristic based on quads/square-degree (e.g. the 10 degree bounding box you are supplying will download approximately 4733 quads) and prompt the user?

2. The v0 basemaps API used a primary quad identifier that contained the zoom and tile coordinates in northing and easting, left padded with zeros, e.g. `L15-0450E-1275N.tif`. Since the v1 API only allows access to native resolution (you cannot get L14), and the padding is terribly useful, propose changing the default name to simply the quad-id (450-1275 in this case) with the option of auto-creating a directory with the name of the mosaic. Or, more forward thinking, allow choosing one of several built-in output formats. This would require modification of `planet.models.StreamingBody` (https://github.com/planetlabs/planet-client-python/blob/main/planet/models.py#L82) to allow specifying the output name.

**Related Issue(s):**

Closes #446

**Proposed Changes:**

1. Initial support for mosaics API

**PR Checklist:**

- [] This PR is as small and focused as possible
- [] If this PR includes proposed changes for inclusion in the changelog, the title of this PR summarizes those changes and is ready for inclusion in the Changelog.
- [] This PR does not break any examples or I have updated them

### DIFF
--- a/examples/mosaics-cli.sh
+++ b/examples/mosaics-cli.sh
@@ -1,0 +1,16 @@
+#!/bin/bash
+
+echo -e "Global Basemap Series"
+planet mosaics series list --name-contains=Global | jq .[].name
+
+echo -e "\nLatest Global Monthly"
+planet mosaics series list-mosaics "Global Monthly" --latest --pretty
+
+echo -e "\nHow Many Quads?"
+planet mosaics search 09462e5a-2af0-4de3-a710-e9010d8d4e58 --bbox=-100,40,-100,40.1
+
+echo -e "\nWhat Scenes Contributed to Quad?"
+planet mosaics contributions 09462e5a-2af0-4de3-a710-e9010d8d4e58 455-1273
+
+echo -e "\nDownload Them!"
+planet mosaics download 09462e5a-2af0-4de3-a710-e9010d8d4e58 --bbox=-100,40,-100,40.1 --output-dir=quads
--- a/examples/mosaics-cli.sh
+++ b/examples/mosaics-cli.sh
@@ -1,16 +1,16 @@
 #!/bin/bash
 
-echo -e "Global Basemap Series"
+echo -e "List the mosaic series that have the word Global in their name"
 planet mosaics series list --name-contains=Global | jq .[].name
 
-echo -e "\nLatest Global Monthly"
+echo -e "\nWhat is the latest mosaic in the series named Global Monthly, with output indented"
 planet mosaics series list-mosaics "Global Monthly" --latest --pretty
 
-echo -e "\nHow Many Quads?"
-planet mosaics search 09462e5a-2af0-4de3-a710-e9010d8d4e58 --bbox=-100,40,-100,40.1
+echo -e "\nHow many quads are in the mosaic with this ID (name also accepted!)?"
+planet mosaics search 09462e5a-2af0-4de3-a710-e9010d8d4e58 --bbox=-100,40,-100,40.1 | jq .[].id
 
-echo -e "\nWhat Scenes Contributed to Quad?"
+echo -e "\nWhat scenes contributed to this quad in the mosaic with this ID (name also accepted)?"
 planet mosaics contributions 09462e5a-2af0-4de3-a710-e9010d8d4e58 455-1273
 
-echo -e "\nDownload Them!"
+echo -e "\nDownload them to a directory named quads!"
 planet mosaics download 09462e5a-2af0-4de3-a710-e9010d8d4e58 --bbox=-100,40,-100,40.1 --output-dir=quads
--- a/planet/clients/mosaics.py
+++ b/planet/clients/mosaics.py
@@ -14,10 +14,10 @@
 
 import asyncio
 from pathlib import Path
-from typing import AsyncIterator, Optional, Tuple, Type, TypeVar, Union, cast
+from typing import AsyncIterator, Optional, Sequence, Type, TypeVar, Union, cast
 from planet.clients.base import _BaseClient
 from planet.constants import PLANET_BASE_URL
-from planet.exceptions import MissingResource
+from planet.exceptions import ClientError, MissingResource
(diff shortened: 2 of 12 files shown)


joferkington:
Thanks a ton for this!!

Minor doc comment that could be addressed later:  It's a bit unclear from the CLI --help docs that `name` can be either an ID or a name.  I think allowing either ID or name in the CLI as the same argument is a great idea! It's just not immediately clear to someone who isn't reading the code that that's an option.  The examples also could be interpreted to show that series operations use name and mosaic operations use ID.

With that said, that's a very minor nitpick that could be handled via more examples or other documentation.

planet/clients/mosaics.py
```diff
@@ -0,0 +1,397 @@
...
+        """
+        resp = await self._session.request(
+            "GET",
+            quad["_links"]["items"],
+        )
+        return resp.json()["items"]
+
+    async def download_quad(self,
+                            quad: dict,
+                            directory,
```
stephenhillier, line 348: In the Data API client, the `directory` arg defaults to current dir. Is it worth making that consistent? I also think your idea of defaulting to the mosaic name would work

examples/mosaics-cli.sh
```diff
@@ -0,0 +1,16 @@
...
+planet mosaics series list-mosaics "Global Monthly" --latest --pretty
+
+echo -e "\nHow Many Quads?"
+planet mosaics search 09462e5a-2af0-4de3-a710-e9010d8d4e58 --bbox=-100,40,-100,40.1
+
+echo -e "\nWhat Scenes Contributed to Quad?"
+planet mosaics contributions 09462e5a-2af0-4de3-a710-e9010d8d4e58 455-1273
+
+echo -e "\nDownload Them!"
+planet mosaics download 09462e5a-2af0-4de3-a710-e9010d8d4e58 --bbox=-100,40,-100,40.1 --output-dir=quads
```
stephenhillier, line 16: Just curious if we think users might need guard rails for this. Could this be a significant enough amount of data to warrant a confirmation step?

planet/clients/mosaics.py
```diff
@@ -0,0 +1,397 @@
...
+        acquired_lt: Optional[str] = None,
+        latest: bool = False,
+    ) -> AsyncIterator[dict]:
+        """
+        List the mosaics you have access to.
+
+        Example:
+
+        ```
+        mosaics = await client.list_mosacis()
```
stephenhillier, line 167:
```suggestion
        mosaics = await client.list_mosaics()
```

planet/clients/mosaics.py
```diff
@@ -0,0 +1,397 @@
...
+
+        ```
+        quad = await client.get_quad("d5098531-aa4f-4ff9-a9d5-74ad4a6301e5", "1234-5678")
+        print(quad)
+        ```
+        """
+        if not _is_uuid(name_or_id):
+            mosaic = await self.get_mosaic(name_or_id)
+            if mosaic is None:
+                return {}
```
stephenhillier, line 320: Along the same lines as the comment about the `Optional[dict]` return value, raising some kind of MosaicNotFoundException here would be more clear.

stephenhillier: also I notice some methods require the metadata dict up front, and others (like this one) fetch it under the hood by name/ID. It would be great to standardize on accepting either, fetching metadata if needed.

ischneider:
Thanks @joferkington 

> Minor doc comment that could be addressed later: It's a bit unclear from the CLI --help docs that `name` can be either an ID or a name. I think allowing either ID or name in the CLI as the same argument is a great idea! It's just not immediately clear to someone who isn't reading the code that that's an option. The examples also could be interpreted to show that series operations use name and mosaic operations use ID.

Yes, good call to make this clearer. Initially I tried to add `help` to the click argument but it doesn't allow that - seems the appropriate way is to add details to the doc string for the command, so I'll do that.

As for the examples, I did not consider that interpretation - will make that clearer!

examples/mosaics-cli.sh
ischneider: Yes, this was one of the issues (1) I raised in the MR description.

carl-adams-planet:
Looks like with the branch re-targeting there are some conflicts to resolve.

I am planning to push beta-2 when this is in.

ischneider:
> Looks like with the branch re-targeting there are some conflicts to resolve.
> 
> I am planning to push beta-2 when this is in.

Done! For whatever reason, getting a test fail locally (hangs) but can figure that out later... Going to follow up w/ narrative docs next

PR #1146, opened 2025-05-21
session stops background loop on exit/close

ischneider:
I was unable to cleanly close the event loop and, therefore, the background thread.

One issue is that the sync API doesn't have any explicit resource cleanup hook.

I attempted to override `__del__` for the `Session` as a means to cleanup and while this "worked" (tests all pass without -Werror)... When run w/ -Werror, it seems like a socket gets garbage collected before close is called and this triggers a warning that fails the test due to -Werror (convert warnings to errors)

### DIFF
--- a/planet/http.py
+++ b/planet/http.py
@@ -282,6 +282,12 @@ class Session(BaseSession):
         self._limiter = _Limiter(rate_limit=RATE_LIMIT, max_workers=MAX_ACTIVE)
         self.outcomes: Counter[str] = Counter()
 
+        self._loop: asyncio.AbstractEventLoop = None  # type: ignore
+
+    def _init_loop(self):
+        if self._loop:
+            return
+
         # create a dedicated event loop for this httpx session.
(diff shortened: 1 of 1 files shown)


asonnenschein (approved):
LGTM!

tests/integration/test_orders_api.py
```diff
@@ -337,7 +337,7 @@ async def test_get_order(oid, order_description, session):
 
 
 @respx.mock
-def test_get_order_sync(oid, order_description, session):
+def test_get_order_sync(oid, order_description, ):
```
asonnenschein, line 340:
```suggestion
def test_get_order_sync(oid, order_description):
```

nit: remove trailing comma and whitespace character

ischneider: oops, didn't mean to add this - reverted

fwiw, was trying to figure out the cause of "leaks" and while session isn't used here (or in other tests), didn't want to make too much noise in this commit

PR #1156, opened 2025-06-06
Features API item delete support

stephenhillier:
Support for deleting items in Features API sync/async/cli clients

### DIFF
--- a/planet/cli/features.py
+++ b/planet/cli/features.py
@@ -97,6 +97,7 @@ async def collection_get(ctx, collection_id, pretty):
         result = await cl.get_collection(collection_id)
         echo_json(result, pretty)
 
+
 @command(collections, name="delete")
 @click.argument("collection_id", required=True)
 async def collection_delete(ctx, collection_id, *args, **kwargs):
@@ -108,7 +109,6 @@ async def collection_delete(ctx, collection_id, *args, **kwargs):
     """
     async with features_client(ctx) as cl:
         await cl.delete_collection(collection_id)
-
 
 
 @features.group()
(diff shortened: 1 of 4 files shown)


planet/cli/features.py
```diff
@@ -150,6 +150,36 @@ async def item_get(ctx, collection_id, feature_id, pretty):
         echo_json(feature, pretty)
 
 
+@command(items, name="delete")
+@click.argument("collection_id")
+@click.argument("feature_id", required=False)
+async def item_delete(ctx, collection_id, feature_id, pretty):
```
wiki0831, line 156:
```suggestion
async def item_delete(ctx, collection_id, feature_id):
```

#nit, `pretty` param is not needed for delete

planet/cli/features.py
```diff
@@ -150,6 +150,36 @@ async def item_get(ctx, collection_id, feature_id, pretty):
...
+    """Delete a feature in a collection.
+
+    You may supply either a collection ID and a feature ID, or
+    a feature reference.
+
+    Example:
+
+    planet features items delete my-collection-123 item123
+    planet features items delete "pl:features/my/my-collection-123/item123"
```
wiki0831, line 165: tested with these examples, works great!

wiki0831 (approved):

planet/cli/features.py
stephenhillier: That's true, but currently `pretty` is an option that all commands get, so each function has to accept it. I changed this to `item_delete(ctx, collection_id, feature_id, *args, **kwargs)` to accept additional args without having to list out args we don't intend to use. Let me know what you think.

wiki0831: nice, this looks good!

asonnenschein (approved):

PR #1159, opened 2025-06-25
fix incorrect search_filter arg in Data API documentation example

stephenhillier:
**Proposed Changes:**

Fix incorrect arg in doc example.

`filter` -> `search_filter`


stephenhillier:
also note that use of `search_filter` here is inconsistent with other methods as well as our APIs themselves. We generally use `filter`. However, this would be a breaking change so some discussion is needed before changing that.

asonnenschein (approved):

PR #1161, opened 2025-06-27
Update subscriptions clip tool docstrings

asonnenschein:
This MR removes confusing verbiage from the `planet.subscription_request.build_request()` docstring about clipping, and replaces it with more explicit instructions to `build_request()` and `clip_tool()`.

### DIFF
--- a/docs/python/sdk-guide.md
+++ b/docs/python/sdk-guide.md
@@ -231,7 +231,7 @@ You will need your ACCESS_KEY_ID, SECRET_ACCESS_KEY, bucket and region name.
 To subscribe to scenes that match a filter, use the `subscription_request` module to build a request, and
 pass it to the `subscriptions.create_subscription()` method of the client.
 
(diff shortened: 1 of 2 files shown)


adamweiner (approved):

docs/python/sdk-guide.md
```diff
@@ -256,7 +258,7 @@ source = catalog_source(
     time_range_type="acquired",
 )
 
-request = build_request("Standard PSScene Ortho Analytic", source=source, delivery={})
+request = build_request("Standard PSScene Ortho Analytic", source=source, delivery={}, clip_to_source=True)
 
 # define a delivery method. In this example, we're using AWS S3.
 delivery = amazon_s3(ACCESS_KEY_ID, SECRET_ACCESS_KEY, "test", "us-east-1")
```
adamweiner, line 264: shall we move this above `request = ...`, and then set `delivery=delivery`?

if copied as-is, delivery will not be configured on the subscription request.

asonnenschein: Ahh good catch!!  Updated.

PR #1170, opened 2025-07-21
Refactor planetary_variable_source() signature

asonnenschein:
This MR refactors the `subscription_request.planetary_variable_source()` to better reflect API functionality by removing the `source_type` positional argument from the signature.  This is a breaking change and will need to ship with v3.x.x.  Source type is an optional property on the Subscriptions API service contract that we are moving away from in favor of more precise parameters like source ID, item types, asset types, etc.  https://github.com/planetlabs/planet-client-python/pull/1107 started the effort to reduce reliance of source type in the SDK, but had to stop short of removing the source type positional argument from `subscription_request.planetary_variable_source()` because it would've been a breaking change in v2.x.x.

### DIFF
--- a/planet/cli/subscriptions.py
+++ b/planet/cli/subscriptions.py
@@ -535,10 +535,9 @@ def request_catalog(item_types,
 
 @subscriptions.command()  # type: ignore
 @translate_exceptions
-@click.option(
-    '--var-id',
-    required=True,
-    help='A Planetary Variable ID. See documentation for all available IDs.')
+@click.option('--source-id',
+              required=True,
+              help='A source ID. See documentation for all available IDs.')
 @click.option(
     '--geometry',
     required=True,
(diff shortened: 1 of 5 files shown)


planet/cli/subscriptions.py
```diff
@@ -559,16 +554,13 @@ def request_catalog(item_types,
               type=types.DateTime(),
               help='Date and time to end subscription.')
 @pretty
-def request_pv(var_type, var_id, geometry, start_time, end_time, pretty):
+def request_pv(var_id, geometry, start_time, end_time, pretty):
```
adamweiner, line 557: full support for removing source_type as required input here. 

however! 3.x is an opportunity to better align the SDK with how these products are currently described in the API.

this method is also how one can create an ARPS subscription, which is not a PV.

in the docs [1] [2] we refer to these as "subscription sources" - not the most descriptive name ever but at least it is not coupled to a subset of the products available here.

what do you think about making that change as well? not necessarily as part of this MR, of course.

[1] https://docs.planet.com/develop/apis/subscriptions/sources/#planetary-variable-and-analysis-ready-source-types
[2] https://docs.planet.com/develop/apis/subscriptions/reference/#tag/subscriptions/operation/createSubscription

asonnenschein:
Added breaking change note to v3 migration doc https://github.com/planetlabs/planet-client-python/pull/1173

PR #1188, opened 2025-08-13
orders + subscriptions: support user_id query param

adamweiner:
**Proposed Changes:**

1. Add support for Orders API & Subscriptions API `user_id` query parameter 

**PR Checklist:**

### DIFF
--- a/tests/integration/test_orders_api.py
+++ b/tests/integration/test_orders_api.py
@@ -191,9 +191,7 @@ async def test_list_orders_user_id_filtering(order_descriptions, session):
 
     # if the value of user_id doesn't get sent as a url parameter,
     # the mock will fail and this test will fail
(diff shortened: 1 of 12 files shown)


tests/integration/test_orders_api.py
```diff
@@ -171,6 +171,52 @@ async def test_list_orders_filtering_and_sorting(order_descriptions, session):
     ]
 
 
+@respx.mock
+@pytest.mark.anyio
+async def test_list_orders_user_id_filtering(order_descriptions, session):
```
asonnenschein, line 176: nit: add a test case with a user_id

asonnenschein (approved):
LGTM!  I think the issue that co-pilot flagged should be addressed (the orders client casts user_id to string, and the subscriptions client does not).

PR #1205, opened 2025-12-18
geometry relation support for data search

### LINKED ISSUE #1204: Add geometry relation optional arg to data_filter.geometry_filter
Recently, `geometry_relation` support was added to [catalog subscriptions](https://github.com/planetlabs/planet-client-python/pull/1203). This would be a really nice feature to add to [data_filter.geometry_filter](https://github.com/planetlabs/planet-client-python/blob/84944f616603018150850d8e1f7e2751b800dbbc/planet/data_filter.py#L221), as this isn't exposed anywhere. You can add `"relation": "intersects"` to the geometry_filter output dict pretty easily, but it's not very obvious by the doc string in the SDK that you can do this yourself.

ischneider:
- optional arg in filter creation
- add support in CLI filter command
- remove some superfluous respx.mock annotations

I considered adding this for the other CLI commands but this would have required significant reworking how geometry is handled, as we currently pass it along as a top-level item in the request and relation is only supported in a filter.

While it would be conceivable to unpack a user's filters and merge in a geometry filter, this felt beyond the scope of the issue.

In other words, with this MR, the only way to use relation via the CLI is to use `planet data filter --geom=... --geom-relation=within | planet data search psscene --filter=-`

**Related Issue(s):**

Closes #1204

**Proposed Changes:**

1. Add relation support for geometry filter in data

**PR Checklist:**

- [X] This PR is as small and focused as possible
- [X] If this PR includes proposed changes for inclusion in the changelog, the title of this PR summarizes those changes and is ready for inclusion in the Changelog.
- [X] I have updated docstrings for function changes and docs in the 'docs' folder for user interface / behavior changes
- [X] This PR does not break any examples or I have updated them

### DIFF
--- a/planet/cli/data.py
+++ b/planet/cli/data.py
@@ -172,8 +172,7 @@ def string_in_to_filter(ctx, param, values) -> Optional[List[dict]]:
 @click.option(
     "--geom-relation",
     type=click.Choice(["intersects", "contains", "within", "disjoint"]),
-    help="""Optional geometry search refinement, defaults to intersects.
-            May also be contains, within, or disjoint.""",
+    help="""Geometry search relation. Options are intersects (default), contains, within, or disjoint.""",
 )
 @click.option('--number-in',
               type=click.Tuple([types.Field(), types.CommaSeparatedFloat()]),


asonnenschein (approved):
nit: I agree with co-pilot's feedback about this docstring https://github.com/planetlabs/planet-client-python/pull/1205/changes#r2662508698.  Otherwise LGTM!

PR #1217, opened 2026-06-09
Quota API SDK/CLI

asonnenschein:
### Description

  Adds support for the Planet [Quota Reservations API](https://docs.planet.com/develop/apis/quota/) to the SDK. Includes an async client, sync client, and CLI built on the
  async client, following the same patterns used by the existing Planet APIs that are supported in the SDK.

  ### Changes

  - Add `QuotaClient` (`planet/clients/quota.py`) covering all Quota Reservations API endpoints: `list_reservations`, `get_reservation`, `create_reservation`,
  `bulk_create_reservations`, `estimate_reservation`, `list_jobs`, `get_job`, plus `list_products` (with optional `supports_reservation` filter) for looking up the
  `product_id` required by reservation requests.
  - Add a `_QuotaPaged` subclass of `models.Paged` to handle the API's `{meta: {next: ...}, results: [...]}` pagination shape.
  - Add `QuotaAPI` (`planet/sync/quota.py`) synchronous wrapper, exposed on `Planet` as `pl.quota`.
  - Add `planet quota` CLI group (`planet/cli/quota.py`) with `products`, `reservations`, and `jobs` subcommands. List commands support `--limit`, `--page-size`, `--sort`,
  `--fields`, and repeated `--filter KEY=VALUE`. Mutation commands accept repeated `--aoi-ref` or `--aoi-refs <json>` (string, file, or stdin) for large AOI lists.
  - Register `QuotaClient` in `planet.clients._client_directory` so `sess.client('quota')` resolves it, and export `QuotaClient` from the top-level `planet` package.

### DIFF
--- a/planet/cli/quota.py
+++ b/planet/cli/quota.py
@@ -1,0 +1,330 @@
+# Copyright 2026 Planet Labs PBC.
+#
+# Licensed under the Apache License, Version 2.0 (the "License"); you may not
+# use this file except in compliance with the License. You may obtain a copy of
+# the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+# WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+# License for the specific language governing permissions and limitations under
+# the License.
+"""Quota Reservations CLI"""
+from contextlib import asynccontextmanager
+from typing import Dict, List, Optional, Tuple
+
+import click
+
+from planet.cli.io import echo_json
+from planet.clients.quota import QuotaClient
+
+from .cmds import command
+from .options import compact, limit
+from .session import CliSession
+from . import types
+
+
+@asynccontextmanager
+async def quota_client(ctx):
+    async with CliSession(ctx) as sess:
+        cl = QuotaClient(sess, base_url=ctx.obj['BASE_URL'])
+        yield cl
+
+
+def _parse_filters(raw: Tuple[str, ...]) -> Optional[Dict[str, str]]:
+    """Parse repeated `--filter KEY=VALUE` options into a dict."""
+    if not raw:
+        return None
+    out: Dict[str, str] = {}
+    for entry in raw:
+        if '=' not in entry:
+            raise click.BadParameter(
+                f"--filter expects KEY=VALUE; got {entry!r}")
+        key, value = entry.split('=', 1)
+        key = key.strip()
+        if not key:
+            raise click.BadParameter(f"--filter has empty key: {entry!r}")
+        out[key] = value
+    return out
+
+
+@click.group()  # type: ignore
+@click.pass_context
+@click.option('-u',
+              '--base-url',
+              default=None,
+              help='Assign custom base Account API URL (e.g. '
+              'https://api.planet.com/account/v1).')
+def quota(ctx, base_url):
+    """Commands for interacting with the Quota Reservations API."""
+    ctx.obj['BASE_URL'] = base_url
+
+
+# ---------------------------------------------------------------------------
+# Products
+# ---------------------------------------------------------------------------
+
+
+@quota.group()
+def products():
+    """Commands for inspecting products that support quota reservations."""
+    pass
+
+
+@command(products, name='list', extra_args=[compact])
+@click.option(
+    '--supports-reservation/--no-supports-reservation',
+    default=None,
+    help='Filter products by whether they support quota reservations.')
+async def products_list(ctx, supports_reservation, pretty, compact):
+    """List products available to your organization.
+
+    Each product's `id` is the value to pass as `--product-id` when creating
+    or estimating a reservation.
+
+    Example:
+
+    planet quota products list --supports-reservation
+    """
+    async with quota_client(ctx) as cl:
+        results = await cl.list_products(
+            supports_reservation=supports_reservation)
+        if compact:
+            keys = ('id',
+                    'name',
+                    'title',
+                    'supports_reservation',
+                    'quota_total',
+                    'quota_used',
+                    'unlimited_quota')
+            results = [{
+                k: v
+                for k, v in p.items() if k in keys
+            } for p in results]
+        echo_json(results, pretty)
+
+
+# ---------------------------------------------------------------------------
+# Reservations
+# ---------------------------------------------------------------------------
+
+
+@quota.group()
+def reservations():
+    """Commands for managing quota reservations."""
+    pass
+
+
+@command(reservations, name='list', extra_args=[limit])
+@click.option('--fields',
+              default=None,
+              help='Comma-separated list of fields to include in results.')
+@click.option('--sort',
+              default=None,
+              help='Sort spec, e.g. `created_at` or `-created_at`.')
+@click.option('--filter',
+              'filters',
+              multiple=True,
+              metavar='KEY=VALUE',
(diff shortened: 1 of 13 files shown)


asonnenschein:
@asonnenschein to do - add tests.

planet/clients/quota.py
```diff
@@ -0,0 +1,422 @@
...
+            params.update(filters)
+        return params
+
+    async def list_reservations(
+        self,
+        limit: int = 100,
+        fields: Optional[str] = None,
+        sort: Optional[str] = None,
+        filters: Optional[Dict[str, Any]] = None,
+        page_size: Optional[int] = None,
```
tbarsballe: Do we want to set a default here? Especially since this is only really relevant to the internal performance, as end-users just get an iterator. Subscriptions [uses 500](https://github.com/planetlabs/planet-client-python/blob/main/planet/clients/subscriptions.py#L76), for example.

planet/clients/quota.py
```diff
@@ -0,0 +1,422 @@
...
+import logging
+from typing import Any, AsyncIterator, Dict, List, Optional, TypeVar
+
+from planet.clients.base import _BaseClient
+from planet.exceptions import APIError, ClientError
+from planet.http import Session
+from planet.models import Paged, Response
+from ..constants import PLANET_BASE_URL
+
+BASE_URL = f'{PLANET_BASE_URL}/account/v1'
```
tbarsballe, line 25: I don't think the way this was done is necessarily wrong or needs to be changed, but strictly speaking the base url for the Quota API (following what's documented in the [API reference](https://docs.planet.com/develop/apis/quota/reference/)) is `f'{PLANET_BASE_URL}/account/v1/quota-reservations'`

It looks like this is using the parent URL so that it can also accommodate the `/my/products` endpoint, which is technically a different API. It is possible in the future we may have need of other features under the `/my/` endpoint elsewhere. We can probably just refactor if it ever comes up, but I wanted to at least point it out.

planet/sync/quota.py
```diff
@@ -0,0 +1,156 @@
...
+        self,
+        limit: int = 100,
+        fields: Optional[str] = None,
+        sort: Optional[str] = None,
+        filters: Optional[Dict[str, Any]] = None,
+        page_size: Optional[int] = None,
+    ) -> Iterator[dict]:
+        """Iterate over quota reservations.
+
+        See [QuotaClient.list_reservations][planet.clients.quota.QuotaClient.list_reservations]
```
tbarsballe: Should probably verify these links work as expected after deploying the docs - the only other place I found links is in subscription_request and they're [using a slightly different syntax](https://github.com/planetlabs/planet-client-python/blob/main/planet/subscription_request.py#L64)

tests/integration/test_quota_api.py
```diff
@@ -0,0 +1,379 @@
...
+        "status": "complete",
+        "processed_items": 2,
+        "total_items": 2,
+        "percentage": 100,
+    }
+
+
+# ---------------------------------------------------------------------------
+# list_reservations
+# ---------------------------------------------------------------------------
```
tbarsballe: This does not match the structure of comments elsewhere in our test code, and seems to be a bit of an AI-code smell.

tests/integration/test_quota_api.py
```diff
@@ -0,0 +1,379 @@
...
+        httpx.Response(200, json=_reservations_page(2, 4))
+    ])
+
+    items = [r async for r in cl_async.list_reservations()]
+    assert [r["id"] for r in items] == [0, 1, 2, 3]
+
+
+@respx.mock
+async def test_list_reservations_respects_limit():
+    # Two pages of two items each — limit=3 cuts iteration short.
```
tbarsballe: Nitpick: This should maybe be a pydoc (`"""`) rather than an inline comment, especially since the line that actually does this is 122, not 120

planet/cli/quota.py
```diff
@@ -0,0 +1,330 @@
...
+              help='Assign custom base Account API URL (e.g. '
+              'https://api.planet.com/account/v1).')
+def quota(ctx, base_url):
+    """Commands for interacting with the Quota Reservations API."""
+    ctx.obj['BASE_URL'] = base_url
+
+
+# ---------------------------------------------------------------------------
+# Products
+# ---------------------------------------------------------------------------
```
tbarsballe: This comment structure is not found elsewhere in our codebase, and doesn't add much of value. Remove?

tbarsballe (commented):
Overall, this looks good. I've run some test commands locally, and everything works how I'd expect (though I didn't actually try reserving anything).

There's a little bit of an issue with verbose structure + vague or redundant comments throughout the MR, that is common with AI code but it's not too bad there. I've called out a couple places where it's noticeably inconsistent.

Following on from the discussion points during last weeks office hours: I ran `quota reservations list` on a customer account, and can confirm non-null organization_ids are returned in the results - it looks like the weirdness you were seeing when testing is a quirk of org 1.

planet/clients/quota.py
asonnenschein: Yes!  Good idea.  Updated with default page_size of 500.

tests/integration/test_quota_api.py
asonnenschein: You're right, this is an AI code smell.  Removed all of the #--- section ---# dividers.

planet/cli/quota.py
asonnenschein: Removed, this is more AI code smell.

tests/integration/test_quota_api.py
asonnenschein: Converted to proper test docstring.

planet/sync/quota.py
asonnenschein: Aligned all quota doc links with the structure in subscription_request.

planet/clients/quota.py
```diff
@@ -0,0 +1,422 @@
...
+
+LOGGER = logging.getLogger()
+
+T = TypeVar("T")
```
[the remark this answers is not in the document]
asonnenschein: Good catch - `T`/`TypeVar` were leftovers and never used. Removed both the declaration and the import.

planet/cli/quota.py
```diff
@@ -0,0 +1,330 @@
...
+def _aoi_refs_from_options(aoi_ref: Tuple[str, ...],
+                           aoi_refs_json: Optional[List[str]]) -> List[str]:
+    refs: List[str] = list(aoi_ref)
+    if aoi_refs_json:
+        refs.extend(aoi_refs_json)
+    if not refs:
+        raise click.BadParameter(
+            'At least one AOI ref must be supplied via --aoi-ref or '
+            '--aoi-refs.')
+    return refs
```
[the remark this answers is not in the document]
asonnenschein: Confirmed - `types.JSON()` returns whatever the payload parses to, so `--aoi-refs '{"a": 1}'` would silently extend the ref list with the dict's keys and submit them as AOI refs. `_aoi_refs_from_options` now rejects anything that is not a JSON array, and rejects non-string entries within the array. Added regression tests for both cases.

planet/clients/quota.py
asonnenschein: Thanks for flagging - agreed on the reasoning. Leaving `BASE_URL` at the `/account/v1` parent for now so the client can reach both `/quota-reservations` and `/my/products`, which is the pragmatic option while those two live behind one client. Noted that `/my/` is technically a separate API, so if we grow more `/my/` surface area elsewhere this is the seam to split on.

tbarsballe (approved):
Thanks for the fixes, looks good to me.

PR #1223, opened 2026-09-02
Add Catalog API SDK and CLI

asonnenschein:
### Description

Adds support for the [Planet Catalog API](https://docs.planet.com/develop/apis/catalog/) to the SDK.  Includes an async client, sync client, and CLI built on the async client, following the same patterns used by the existing Planet APIs that are supported in the SDK.

### Changes

- Add `CatalogClient` (`planet/clients/catalog.py`) covering all Catalog API endpoints: `get_landing_page`, `get_conformance`, `list_collections`, `get_collection`, `get_collection_queryables`, `list_items`, `get_item`, `simple_search`, `search`.
- Add `CatalogAPI` (`planet/sync/catalog.py`) synchronous wrapper, exposed on `planet` as `pl.catalog`.
- Add `planet catalog` CLI group (`planet/cli/catalog.py`) with `collections`, `conformance`, `items`, `landing-page`, `search`, and `simple-search` subcommands.
- Defaults to legacy base URL and authenticates with an OAuth bearer token.  Both regional legacy URLs are added to `constants.py`.
- Paging works two ways. The GET endpoints return a STAC `links` entry with `rel: next`, while POST /search returns a `context.next` token that must be re-sent with the original query. Both pagers subclass `planet.models.Paged` so the limit semantics and page-cycle detection are shared.


ischneider:
A few quick bits of feedback:

- as copilot noted, there are numerous places where client validation is missing, both CLI flags (required=True) and in the client, e.g. `if required_thing is None: raise ClientException('required_thing ...')`
- the try/except/raise pattern in the client should go away - adds nothing as the client methods and the `request` method both note the types of exception that can be raised
- i think compliance, landing-page, queryables, etc. are not interesting to CLI or client users and can be removed - instead these allow powering a generic STAC client via discovery...

the tests seem like a funny mix of sync/async and one-of cases, e.g. `test_list_items_omits_unset_params` ... there is real fragility in our sync wrappers omitting or mixing up a parameter but I think we can test this using permutations and some python test magic (e.g. loop over client and make sure at a code level, each function has the same params?)

PR #1225, opened 2026-09-09
Generate Pydantic models from Destinations API spec, validate on release

Regan-Koopmans:
I have introduced typing for the Destination API client methods using Pydantic models. I generated these models from the live Destinations API spec using `datamodel-code-generator`. I have added a helpful `nox` command to conveniently regenerate these models going forward:
```
nox -s generate_models
```
Having this in place means that we can validate our models against the production API. I have added a `nox` command to do just this:

```
nox -s validate_models 
```

I have configured this to run in the CI for every release, which guarantees that our code does not diverge from the API (at least at the time of release). This will not detect new endpoints in the Destinations API that we have not implemented yet (unless they introduce new DTOs). This could be useful but it is out of scope for this PR. 

**PR Checklist:**


Regan-Koopmans:
Hey @asonnenschein, let me know if you want to have a sync call on this to discuss approach.

ischneider:
Hi Regan,

This would be a breaking change for the Destinations clients but it raises the issue of untyped dicts (throughout the API) as responses...

This is a long-standing critique of mine that also includes a lack of a "request" object (that could be used by both sync/async dispatch, rather than duplicating all request params and docs for each client) and a broader more/useful "response" object, that for example, allows reading raw JSON or as a typed object. This would also support non-body content, such as rate-limiting headers, etc.

I've suggested we could migrate to this approach, without breaking existing clients, by adding a new dispatcher type and defining request/response objects that could be reused under the hood by existing client implementations.

One discussion we had was "is there value in just having users generate clients from openapi specializations?". I think there's something to consider there (as it allows other language users to build on our APIs) unless we provide some value/convenience over generated code.

Finally, while there's certainly value in having a stronger typed response, I think validation of requests is more valuable than validation of responses - our services should ensure the contract and validity of their responses.

planet/api_models/destinations.py
```diff
@@ -0,0 +1,464 @@
...
+        | S3CompatiblePatchParams
+    ]
+):
+    root: Annotated[
+        GoogleCloudStoragePatchParams | AmazonS3PatchParams | AzureCloudStoragePatchParams | OracleCloudStoragePatchParams | S3CompatiblePatchParams,
+        Field(description='Patch parameters for the given Destination type.'),
+    ]
+
+
+class DestinationPatchRequest1(BaseModel):
```
asonnenschein: I am worried that names like `DestinationPatchRequest1`, `DestinationPatchRequest2`, `DestinationPatchRequest3`, etc are not descriptive enough and could present a maintenance burden in the future.  Let's give naming some extra attention to ensure that the name of the type is descriptive and ideally self-documenting.  Using `DestinationPatchRequestN` as an example, I cannot clearly understand why they all exist.

asonnenschein:
I agree with @ischneider's point about this being a breaking change, but otherwise I think this is a big step in the right direction!  Perhaps we can refactor the approach here so that the default response objects are still JSON, and typed response object are opt-in?

pyproject.toml
```diff
@@ -11,6 +11,7 @@ dependencies = [
   "geojson",
   "httpx>=0.28.0",
   "jsonschema",
+  "pydantic>=2.0",
```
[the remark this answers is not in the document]
Regan-Koopmans: Fixed!

planet/clients/destinations.py
```diff
@@ -13,19 +13,18 @@
 # the License.
 
 import logging
-from typing import Any, Dict, Optional, TypeVar
+from typing import Any, Dict, Optional
 
 from planet.clients.base import _BaseClient
 from planet.exceptions import APIError, ClientError
 from planet.http import Session
+from ..api_models.destinations import Destination, DestinationsResponse
```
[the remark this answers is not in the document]
Regan-Koopmans: Fixed!

planet/api_models/destinations.py
Regan-Koopmans: I have managed to massage the code generation to prevent these cases. Let me know what you think!

Regan-Koopmans:
Hi @ischneider, @asonnenschein. I have reverted changes to the Destinations client code. The changes are now backwards compatible for existing users, but the models are still available for users who would like to validate against them ( via`Destination.model_validate(...)`). Having this in place will allow incremental adoption, but already adds some safety for detecting changes in the underlying API (even if we do not implement the client code in terms of these new models).

.github/workflows/publish-pypi.yml
```diff
@@ -24,9 +24,13 @@ jobs:
           restore-keys: |
             ${{ runner.os }}-pip
 
-      - name: Build, verify, and upload to PyPI
+      - name: Validate models against live API specs
         run: |
           pip install --upgrade nox
+          nox -s validate_models
```
asonnenschein: This makes fetching the Destinations API specification and validating a hard release dependency.  Let's think about how we want to ship this.  As we expand stronger typing to other APIs it will add additional dependencies through API calls to other OpenAPI specifications.  Should this be enforced earlier on, perhaps in MR workflow?  This way a regression or failure can be caught and actioned earlier on.  The MR workflow (https://github.com/planetlabs/planet-client-python/blob/main/.github/workflows/test.yml)  already has steps for things like linting, analyzing types, etc.

planet/api_models/destinations.py
asonnenschein: Nice!  The single `DestinationPatchRequest()` type is clean and simple.  I still see a nuance on the refactored version, will add my comment there.

planet/api_models/destinations.py
```diff
@@ -0,0 +1,411 @@
...
+):
+    root: Annotated[
+        GoogleCloudStoragePatchParams | AmazonS3PatchParams | AzureCloudStoragePatchParams | OracleCloudStoragePatchParams | S3CompatiblePatchParams,
+        Field(description='Patch parameters for the given Destination type.'),
+    ]
+
+
+class DestinationPatchRequest(BaseModel):
+    model_config = ConfigDict(
+        extra='allow',
```
asonnenschein: [Here is the current API specification for `DestinationPatchRequest`](https://docs.planet.com/redocusaurus/destinations-api-spec.yaml):

```yaml
    DestinationPatchRequest:
      additionalProperties: false
      anyOf:
        - required:
            - parameters
        - required:
            - archive
        - required:
            - name
      description: >-
        A DestinationPatchRequest is an object describing how to update a
        Destination.
      properties:
        archive:
          description: True to archive the destination, false to unarchive.
          type: boolean
        name:
          description: A string to uniquely identify a Destination.
          maxLength: 63
          minLength: 3
          type: string
        parameters:
          $ref: '#/components/schemas/DestinationPatchParameters'
      title: Destination patch request
      type: object
```

So it requires any of `parameters`, `archive`, and/or `required` and no additional properties.  Should this be `'forbid'` instead of `'allow'`?

planet/api_models/destinations.py
```diff
@@ -0,0 +1,411 @@
...
+            max_length=63,
+            min_length=3,
+        ),
+    ] = None
+    parameters: DestinationPatchParameters | None = None
+
+
+class DestinationRequest(BaseModel):
+    model_config = ConfigDict(
+        extra='allow',
```
asonnenschein: Similar to https://github.com/planetlabs/planet-client-python/pull/1225/changes#r4042786817, `DestinationRequest` specifies `additionalProperties: false`.  Should this be `'forbid'`?  There are probably more nuances like this in other models.

One place the typed objects will be valuable is in upfront request validation.  The SDK can validate a user's request object client side and raise an error if it doesn't match the service contract (like it has unsupported properties).

pyproject.toml
```diff
@@ -41,6 +42,12 @@ test = [
     "respx>=0.22.0",
     "coverage[toml]"
 ]
+validate_models = [
+    "pytest==8.3.3",
+    # Pinned exactly: the drift check byte-compares regenerated output, so a
+    # codegen release that changes formatting would fail it and block a release.
+    "datamodel-code-generator[http]==0.79.0",
+]
```
asonnenschein, line 50: Can we make these dev dependencies?

pyproject.toml
asonnenschein: This MR adds `pydantic>=2.0` as a dependency.  `pydantic-core` is implemented in Rust and uses Rust binaries.  @Regan-Koopmans and I have reviewed the `pydantic` and `pydantic-core` source code and confirmed that they ship with pre-built wheels and no local compilation needed.

asonnenschein:
Can we change the models import path from `planet.api_models.*` to something more succinct like `planet.types.*`?

tests/drift/codegen_config.py
```diff
@@ -0,0 +1,188 @@
+# Copyright 2024 Planet Labs PBC.
```
asonnenschein, line 1: This script fetches the latest OpenAPI specification and generates models, that go into `planet.api_models.*`, but it's in the test directory.  We should move this file somewhere else outside of `tests.*` because it generates production code.

tests/drift/codegen_config.py
```diff
@@ -0,0 +1,188 @@
...
+REPO_ROOT = pathlib.Path(__file__).parent.parent.parent
+MODELS_DIR = REPO_ROOT / "planet" / "api_models"
+
+# TODO: extend to other APIs as Pydantic models are adopted:
+#   "subscriptions": "https://api.planet.com/subscriptions/v1/spec",
+#   "orders":        "https://api.planet.com/compute/ops/spec",
+#   "data":          "https://api.planet.com/data/v1/spec",
+SPECS = {
+    "destinations": "https://api.planet.com/destinations/v1/spec",
+}
```
asonnenschein, line 34: Related to pulling this file into it's own namespace (https://github.com/planetlabs/planet-client-python/pull/1225/changes#r4046919238), can this reference be pulled into a separate file?  Something like a constants file?

.github/workflows/publish-pypi.yml
Regan-Koopmans: Agreed. Moved it out of the release workflow and into `test.yml`, next to lint and analyze. Drift now fails on the PR that causes it, and a release no longer depends on a live call to api.planet.com.

One tradeoff: every PR now hits the live spec, so an API outage will redden unrelated PRs. A nightly cron would avoid that. Happy to switch if you prefer.

planet/api_models/destinations.py
Regan-Koopmans: Yes. I dropped the global `--extra-fields allow` so codegen honours the spec. Anything marked `additionalProperties: false` now generates as `extra='forbid'`, and `DestinationPatchRequest` is one of them.

Responses need the opposite, or an additive server change breaks a released SDK. So the codegen script patches `additionalProperties: true` onto every schema reachable from a response body before generation. Reachability is computed from the spec rather than hardcoded, so it carries to the other APIs as we add them.

planet/api_models/destinations.py
Regan-Koopmans: `DestinationRequest` is `forbid` now too.

There is one nuance worth knowing. `AmazonS3Params` and its siblings are marked `additionalProperties: false`, but the API echoes them back inside `Destination.parameters`. Forbidding there would break an older SDK the day we add a param, so tolerance wins and they stay `allow`. Only `*PatchParams` and the top-level request bodies are request-only.

The split:

- **forbid:** `DestinationRequest`, `DestinationPatchRequest`, `DefaultDestinationRequest`, `*PatchParams`
- **allow:** `Destination`, `DestinationsResponse`, `Links`, `Ownership`, `Permissions`, `Error`, `*Params`

Note the clients still take dicts, so `forbid` is latent until we wire up request validation. Your point about upfront validation is the reason to do that.

Added `tests/unit/test_api_models.py` to pin both halves.

pyproject.toml
Regan-Koopmans: Done. `validate_models` now folds into `dev`.

pyproject.toml
Regan-Koopmans: I have made it opt-in. `pydantic` moves out of the core dependencies into a `models` extra, so `pip install planet` no longer pulls in `pydantic-core` at all. Typed models are `pip install planet[models]`.

Nothing outside `planet/types` imports pydantic, and there is a unit test that walks the AST of every module under `planet/` to keep it that way. `planet/types/__init__.py` raises a pointed ImportError naming the extra if it is missing.

Verified in a clean venv: `pip install .` installs no pydantic, and both `import planet` and the CLI work

scripts/codegen_config.py
```diff
@@ -0,0 +1,188 @@
+# Copyright 2024 Planet Labs PBC.
```
Regan-Koopmans: Agreed. Moved to `scripts/codegen_config.py`. The drift test reaches it through a conftest that puts `scripts/` on `sys.path`.

tests/drift/codegen_config.py
Regan-Koopmans: Done! `scripts/codegen_constants.py` now holds `SPECS`, `MODELS_DIR`, `REPO_ROOT`, `HEADER`, `DROP_CONSTRAINT_ANY_OF` and `TARGET_PYTHON_VERSION`. `codegen_config.py` keeps only the logic. Adding an API is one line in `SPECS`.

.github/workflows/publish-pypi.yml
asonnenschein: > ... every PR now hits the live spec, so an API outage will redden unrelated PRs.

This is a risk I am comfortable living with.  If our live API specifications are unreachable we have much bigger problems to solve!

pyproject.toml
asonnenschein: Nice!  Will we need to make `pydantic` a core dependency after we start pulling typed models into code paths that support user access patterns?

planet/api_models/destinations.py
asonnenschein: Thanks!  I still see some types that are marked with `additionalProperties: false` and `extra='allow` though, for example look at the types for `AzureCloudStorageParams` and `GoogleCloudStorageParams`.

> Responses need the opposite, or an additive server change breaks a released SDK. So the codegen script patches additionalProperties: true onto every schema reachable from a response body before generation.

Not sure how much I agree with this statement.  If additive (e.g. non-breaking) changes are made to an API service contract and schema upstream of the SDK, wouldn't we want the model generation to read those changes from the API specification and update the types?  My understanding is that Pydantic won't validate 'extra' properties when they are allowed to pass in.  I feel like any properties defined on a type in the API specification should be explicitly defined in the models.

pyproject.toml
asonnenschein: Disregard, my last comment.  I see how this works now in `types/__init__.py`.  Nice!

planet/api_models/destinations.py
asonnenschein: Hmm I think I understand the approach here better after reading https://github.com/planetlabs/planet-client-python/pull/1225#discussion_r4062407375.  Is the idea that by configuring `extra='allow'`, we can ship additive changes the types as minor versions that don't break users who haven't bumped their version yet?  This way these users aren't broken when we add new properties to the types.

setup.cfg
```diff
@@ -1,5 +1,5 @@
 [options]
-packages = planet, planet.cli, planet.clients, planet.data, planet.sync
+packages = planet, planet.types, planet.cli, planet.clients, planet.data, planet.sync
```
asonnenschein, line 2:
```suggestion
packages = planet, planet.cli, planet.clients, planet.data, planet.sync, planet.types
```

tests/unit/test_api_models.py
```diff
@@ -0,0 +1,156 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+# WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+# License for the specific language governing permissions and limitations under
+# the License.
+"""Contract tests for the generated Destinations models.
+
+These pin the two halves of the unknown-field policy set in
+tests/drift/codegen_config.py. Request models reject unknown fields so a typo
```
asonnenschein, line 17:
```suggestion
scripts/codegen_config.py. Request models reject unknown fields so a typo
```

tests/unit/test_api_models.py
```diff
@@ -0,0 +1,156 @@
+# Copyright 2026 Planet Labs PBC.
```
asonnenschein, line 1: This test should be renamed `test_types.py` after we renamed the `api_models` namespace to `types`.

tests/drift/validate_models.py
```diff
@@ -0,0 +1,87 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+# WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+# License for the specific language governing permissions and limitations under
+# the License.
+"""Pre-release drift detection: regenerate Pydantic models and diff against committed files.
+
+How it works:
+  - datamodel-codegen fetches the live OpenAPI spec and generates models into a temp file.
```
asonnenschein, line 17: > datamodel-codegen fetches the live OpenAPI spec ...

Is this still true?  IIRC `fetch_and_patch_spec` fetches the spec and codegen reads it from a temp file.

scripts/codegen_config.py
```diff
@@ -0,0 +1,171 @@
...
```
asonnenschein, line 1:
```suggestion
# Copyright 2026 Planet Labs PBC.
```

noxfile.py
```diff
@@ -63,8 +70,13 @@ def test(session):
 def lint(session):
     session.install("-e", ".[lint]")
 
-    session.run("flake8", *source_files)
-    session.run('yapf', '--diff', '-r', *source_files)
+    session.run("flake8",
+                f"--exclude={','.join(generated_dirs)}",
```
asonnenschein, line 74:
```suggestion
                f"--extend-exclude={','.join(generated_dirs)}",
```

`--extend-exclude` will append new patterns without erasing defaults.

asonnenschein:
Note - this MR doesn't include a `py.typed`  marker file in the package.  I think that's OK for this MR because the typed models aren't actually invoked in any hot code paths.  Calling this out though because eventually we'll need to add a `py.typed` marker file in the distribution to indicate that the package supports type checking (https://typing.python.org/en/latest/spec/distributing.html#packaging-type-information).

noxfile.py
```diff
@@ -114,6 +126,66 @@ def examples(session):
     session.run('pytest', '--no-cov', 'examples/', '-s', *options)
 
 
+@nox.session
```
asonnenschein, line 129: Consider pinning this to a specific Python version, i.e. `@nox.session(python="3.12")` so that different versions of Python don't adversely impact the byte comparison that this method supports.

noxfile.py
```diff
@@ -114,6 +126,66 @@ def examples(session):
...
+                                         mode="w") as spec_tmp:
+            json.dump(spec, spec_tmp)
+            spec_path = Path(spec_tmp.name)
+        try:
+            session.run(*codegen_config.codegen_argv(spec_path, output))
+        finally:
+            spec_path.unlink(missing_ok=True)
+
+
+@nox.session
```
asonnenschein, line 162: Same comment here https://github.com/planetlabs/planet-client-python/pull/1225/changes#r4073498436

noxfile.py
```diff
@@ -114,6 +126,66 @@ def examples(session):
...
+    session.install("-e", ".[validate_models]")
+
+    import json
+    import tempfile
+
+    sys.path.insert(0, str(Path(__file__).parent / "scripts"))
+    import codegen_config
+
+    for name, url in codegen_config.SPECS.items():
+        output = Path("planet/types") / f"{name}.py"
```
asonnenschein, line 149:
```suggestion
        output = codegen_config.MODELS_DIR / f"{name}.py"
```
 We should use the constant here, not sure if we still need to wrap in `Path()`.

tests/unit/test_api_models.py
```diff
@@ -0,0 +1,156 @@
...
+    assert "field_links" not in dumped["destinations"][0]
+
+
+def test_sdk_does_not_import_pydantic_outside_planet_types():
+    """pydantic is an optional extra: `pip install planet[models]`.
+
+    Nothing outside planet/types may import it, or a plain `pip install planet`
+    breaks at import time. Checked by AST rather than by installing the package
+    two ways, so it runs in the normal suite.
+    """
```
asonnenschein, line 131:
```suggestion
    """pydantic is an optional extra: `pip install planet[models]`."""
```

noxfile.py
```diff
@@ -114,6 +126,66 @@ def examples(session):
...
+def generate_models(session):
+    """Re-generate Pydantic models for the Destinations API in planet/types/.
+
+    Uses the same pinned datamodel-code-generator as `nox -s validate_models`,
+    so the committed output is byte-identical to what the drift check
+    regenerates. Do not reformat the result.
+
+    Run after a known API spec change to refresh the models, then re-run
+    validate_models to confirm compatibility.
+    """
```
asonnenschein, line 139:
```suggestion
    """Re-generate the Pydantic models in planet/types/ from the live specs.

    Output must stay byte-identical to what `nox -s validate_models`
    regenerates. Run after a spec change, then re-run validate_models.
    """
```

scripts/codegen_constants.py
```diff
@@ -0,0 +1,63 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+# WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+# License for the specific language governing permissions and limitations under
+# the License.
+"""Constants for Pydantic model code generation.
+
+Separated from the codegen logic so the spec URLs and output paths can be
+read and changed without reading the generator. Adding an API means adding
+one line to SPECS.
+"""
```
asonnenschein, line 19:
```suggestion
```
I think it's safe to remove this docstring altogether if we decide to use this file for all script level constants.

scripts/codegen_config.py
```diff
@@ -0,0 +1,171 @@
...
+# the License.
+"""Single source of truth for the Pydantic model codegen invocation.
+
+Both `nox -s generate_models` and the drift test import this module. The drift
+test byte-compares regenerated output against the committed models, so the two
+must build an identical command line from an identical version of
+datamodel-code-generator (pinned in the `validate_models` extra).
+
+Spec URLs, output paths and other settings live in codegen_constants.
+"""
```
asonnenschein, line 22:
```suggestion
```

Consider removing this docstring altogether, I don't think it's important to call out how the module is invoked here.

scripts/codegen_config.py
```diff
@@ -0,0 +1,171 @@
...
+       before the round trip.  It is wrong for responses: a shipped SDK must
+       not raise when Planet adds a field.  So every schema reachable from a
+       response is forced to ``additionalProperties: true``, giving
+       ``extra='allow'``.
+
+    Note the overlap.  ``AmazonS3Params`` and its siblings appear in both
+    requests and responses, so tolerance wins and they are generated as
+    ``allow``.  Only ``*PatchParams`` and the top-level request bodies are
+    request-only, and those get ``forbid``.
+    """
```
asonnenschein, line 111:
```suggestion
    """Fetch an OpenAPI spec and patch it for generating typed models.

    Two patches are applied before datamodel-codegen sees the spec.

    1. Strip pure-constraint ``anyOf`` blocks.  Some schemas use ``anyOf``
       exclusively to express "at least one of these fields must be present",
       using inline objects that each carry only a ``required`` key.
       datamodel-codegen cannot name these inline schemas and falls back to
       numbered suffixes (``DestinationPatchRequest1``, etc.).  Removing the
       block yields a single, flat model.  The constraint is server-enforced;
       the client SDK does not need to replicate it.

    2. Relax ``additionalProperties`` on response schemas.  Codegen is run
       without a global ``--extra-fields`` override, so it honours the spec:
       ``additionalProperties: false`` becomes ``extra='forbid'``.  That is
       what we want for request models -- a typo'd key fails client side,
       before the round trip.  It is wrong for responses: a shipped SDK must
       not raise when Planet adds a field.  So every schema reachable from a
       response is forced to ``additionalProperties: true``, giving
       ``extra='allow'``.
    """
```

scripts/codegen_config.py
asonnenschein, line 1: Can we rename this file something more specific - something like `scripts/types.py` or `scripts/type_gen.py` (or another more specific name of your choosing)?  We're only generating code for typed data models, so I don't want to confuse folks with a filename that could be interpreted to as having a wider impact.

scripts/codegen_constants.py
asonnenschein, line 1: Let's rename this file simply to `scripts/constants.py` so that it's not limited to codegen and we can maintain other script level constants here.

noxfile.py
```diff
@@ -114,6 +126,66 @@ def examples(session):
...
+    session.install("-e", ".[validate_models]")
+
+    import json
+    import tempfile
```
asonnenschein, line 143: These imports should be at the top of the file.

noxfile.py
```diff
@@ -114,6 +126,66 @@ def examples(session):
...
+
+    sys.path.insert(0, str(Path(__file__).parent / "scripts"))
+    import codegen_config
```
asonnenschein, line 146: This import should be at the top of the file.

.github/workflows/test.yml
```diff
@@ -56,6 +56,29 @@ jobs:
...
+      with:
+        python-version: 3.12
+    - name: Pip cache
+      uses: actions/cache@v4
+      with:
+        path: ~/.cache/pip
+        key: ${{ runner.os }}-pip
+        restore-keys: |
+          ${{ runner.os }}-pip
+    # Fetches live OpenAPI specs from api.planet.com. No API key required.
```
asonnenschein, line 76:
```suggestion
```

Not a necessary comment IMO; the validate_models command documents this, and I don't see any other comments in this file.

planet/api_models/destinations.py
Regan-Koopmans: Yes, that's the idea. `AzureCloudStorageParams`, `GoogleCloudStorageParams` and the other `*Params` are `additionalProperties: false` in the spec. The API also returns them inside `Destination.parameters`. If they were `forbid`, an SDK already installed would raise the day the server adds a field. Users would have to upgrade to read a response.

Nothing is lost when the spec changes. Codegen reads the new field, the drift check fails, and we regenerate and ship it as a typed field in a minor release. `allow` only covers the gap between the server change and the user upgrading. Request-only models (`*PatchParams` and request bodies) stay `forbid`, so typos are still caught before the request is sent.

setup.cfg
Regan-Koopmans: Already applied.

tests/unit/test_types.py
```diff
@@ -0,0 +1,156 @@
...
+    assert "field_links" not in dumped["destinations"][0]
+
+
+def test_sdk_does_not_import_pydantic_outside_planet_types():
+    """pydantic is an optional extra: `pip install planet[models]`.
+
+    Nothing outside planet/types may import it, or a plain `pip install planet`
+    breaks at import time. Checked by AST rather than by installing the package
+    two ways, so it runs in the normal suite.
+    """
```
Regan-Koopmans: Already applied.

tests/unit/test_api_models.py
Regan-Koopmans: Done.

scripts/codegen_config.py
Regan-Koopmans: Done.

noxfile.py
Regan-Koopmans: Done.

noxfile.py
Regan-Koopmans: Done.

noxfile.py
Regan-Koopmans: Done.

scripts/codegen_constants.py
Regan-Koopmans: Done.

scripts/codegen_config.py
Regan-Koopmans: Done.

noxfile.py
Regan-Koopmans: Done.

noxfile.py
Regan-Koopmans: Done.

.github/workflows/test.yml
Regan-Koopmans: Done.

tests/unit/test_types.py
```diff
@@ -0,0 +1,156 @@
+# Copyright 2026 Planet Labs PBC.
```
Regan-Koopmans: Done. Renamed to `tests/unit/test_types.py`.

tests/drift/validate_models.py
Regan-Koopmans: Not anymore. Updated: the spec is fetched and patched into a temp file, and codegen reads that file.

noxfile.py
Regan-Koopmans: Done. Both `generate_models` and `validate_models` are pinned to 3.12.

noxfile.py
Regan-Koopmans: Done.

scripts/codegen_config.py
Regan-Koopmans: Applied, but I kept the closing "Note the overlap" paragraph. It explains why `AzureCloudStorageParams` and friends are `allow`. Happy to drop it if you prefer.

scripts/type_gen.py
Regan-Koopmans: Renamed to `scripts/type_gen.py`. I avoided `scripts/types.py`: with `scripts/` on `sys.path` it would shadow the stdlib `types` module.

scripts/constants.py
Regan-Koopmans: Done. Renamed to `scripts/constants.py`.

scripts/type_gen.py
asonnenschein: Thanks, and good catch!

asonnenschein (approved):
LGTM!  I think we've given this MR appropriate technical due diligence, great job @Regan-Koopmans!

PR #1227, opened 2026-09-10
Make HTTP retry behavior configurable

asonnenschein:
Adds configurable options for HTTP retry attempts, backoff seconds, and jitter seconds.  Keeps original constant values as default if not explicitly configured.

### Changes
- Session accepts optional `max_retries`, `max_retry_backoff`, and `max_retry_jitter`, each falling back to the module default.
  - `max_retries=0` disables retry
  - `max_retry_jitter=0` makes waits deterministic.
- Jitter is now applied to thresholded waits.  Base wait is now capped at `max_retry_backoff - jitter` and jitter added afterward, so every wait is jittered and the cap still holds.
- New top-level CLI options `--max-retries,` `--max-retry-backoff`, `--max-retry-jitter`, passed through to the session via `CliSession`.
- Unit tests for the new arguments, retry disabling, context passthrough, and `_calculate_wait` edge cases; an integration test asserting `--max-retries` 0 prevents a retry.


tbarsballe:
I'd suggest also including [RATE_LIMIT](https://github.com/planetlabs/planet-client-python/blob/main/planet/http.py#L54) and [MAX_ACTIVE](https://github.com/planetlabs/planet-client-python/blob/main/planet/http.py#L55) as configurable options for Session as well (passed to Limiter), as those control preemptive request limiting.

See also #580 for some background on some significant historical rate-limiting work in the SDK

